PR #41, opened 2022-06-29
postgres syntax

franTarkenton:
* swapped out SYSDATE to currrent_date to make work with postgres


ConradBoydElliottGustafson (approved):
Looks good. Richard is fixing the ER Studio model to generate this code correctly.

PR #78, opened 2022-07-20
Feat/fam user endpoints

franTarkenton:
The following bugs encountered during the sprint demo have now been addressed, ie:

* tests run as expected with output 
* api can be spun up locally as without errors

was tested using python 3.8


explorertolu (approved):

ConradBoydElliottGustafson (approved):
In the future, let's resolve conflicts locally before creating the PR so it doesn't complain about merge conflicts? Although I don't actually see any conflicts here. Let's merge this sucker.

Issue #55, opened 2022-07-07
FAM-API - Define User End Points


franTarkenton:
Define:
* POST User - adding a new user to the api
* GET User - listing the existing users in the system

Related to: https://github.com/bcgov/nr-forests-access-management/issues/20

franTarkenton:
At the moment:
* GET call will return everything
* POST the primary key is not part of the call as it is autogenerated

Question: Should the add / change dates be autogenerated?

basilv:
See https://apps.nrs.gov.bc.ca/int/confluence/display/FOM/API+Design for some design guidelines we applied in FOM. In general, from a security/performance perspective you might not want to always return all fields in a GET (especially one that returns a large number of entities - but this might not apply to FAM as much as FOM. I would generally auto-generate audit fields like create/update date and not supply them as POST parameters, partly to simplify the API and partly to mitigate the impacts of a hostile attacker/user from being able to mess with data.

PR #80, opened 2022-07-20
removing update-date and create-date form POSTS

franTarkenton:
* For FAM user creation, the update_date / create_date are now autogenerated

This ticket should closes https://github.com/bcgov/nr-forests-access-management/issues/55


ConradBoydElliottGustafson (approved):
Elegant solution!

PR #100, opened 2022-08-10
Fix/75 update datamodel

rparfig:
Merging changes on fam_role to fix the not null mandatory columns to null


ConradBoydElliottGustafson:
Closed without merging due to weird merge mystery

PR #105, opened 2022-08-10
Feat/role endpoint

ianliuwk1019:
- Include two basic simple Role endpoints (GetFamRole, CreateFamRole) implementation.
- Add Alembic migration script to adjust model for fam_role table (column nullable).
- Note: please ignore existing role tests for now. Will address that separately.


franTarkenton (dismissed):
Looks good @ianliuwk1019, thinking we should create another ticket that describes having the role GET operation return the roles and the related applications to which they belong.

backend/api/app/crud.py
```diff
@@ -219,6 +219,29 @@ def deleteUser(db: Session, user_id: int):
     return famUser
 
 
+def createFamRole(famRole: schemas.FamRole, db: Session):
+    LOGGER.debug(f"Fam role: {famRole}")
+
+    famRoleDict = famRole.dict()
+
+    # maybe there is a way to get the db to do this for us, but just as easy
```
basilv, line 227: Yes, simple way for create_date is to have the column default set to now(). In FOM DDL scripts I used the following column definition: create_timestamp timestamptz not null default now().

In theory you could have a trigger update the update_date column in the DB, but that I'm generally fine with doing programmatically (I'm not a big fan of triggers).

ianliuwk1019: Hm.. thanks a lot Basil.
Perhaps in this case it is in model.py from the SqlAlchemy that can be set as default.
I will do a bit research on it and change this if possible for this PR.

ianliuwk1019: Hi Basil,
I have provided immediate solution at model.py (This is ORM mapping in python to postgres db).
Previous **server_default=text("CURRENT_DATE")** is wrong and instead I use **default=datetime.datetime.utcnow** for now.

This model.py file was previously generated from ER Studio. And **server_default=text("CURRENT_DATE")** was already there for **create_time** and **update_time**. However, this field with **text("CURRENT_DATE")** is still wrong from what I see, I think it was intended to have **TIMESTAMP** for the column (as can be seen from ORM mapping generated). Even if I change it to **text("LOCALTIMESTAMP")** it will not work. The ER Studio for these two columns are having default value with **CURRENT_DATE**, which does not include time. So the current solution is to use **default** attribute for the ORM mapping from python model.py and it worked when I tested it.

Since I will create another PR for some very simple tests with this role endpoint. I will also try solving these two date columns using migration script to alert the column to have **server_default=text("LOCALTIMESTAMP")** so we can then eliminate the need to set **default** in ORM mapping. Hope it will work.

PR #107, opened 2022-08-11
Migration script for date columns default.

ianliuwk1019:
- Alembic migration script for date columns default.
- Adjust model.py to reflect the default date value change from script.
- Fix for failing tests at schema.py.


franTarkenton (approved):
Thanks for taking care of that Ian

PR #121, opened 2022-08-23
Api router refactoring

ianliuwk1019:
* Split routers into modules.
* Adjust existing tests to pass for router change.


franTarkenton (approved):
Thanks @ianliuwk1019 looks great, as the app grows I think this struct will create a sustainable structure, that makes it easier for multiple people to develop at the same time on different end points.

PR #135, opened 2022-08-26
Add aws secret access

explorertolu:


ConradBoydElliottGustafson (approved):
Pair programmed with myself

PR #136, opened 2022-08-26
Add get secret function

explorertolu:


ConradBoydElliottGustafson (approved):
pair programmed with myself

PR #138, opened 2022-08-26
Add return secret

explorertolu:


ConradBoydElliottGustafson (approved):
pair programmed with myself

PR #139, opened 2022-08-29
fix: get username password from secret

ConradBoydElliottGustafson:


franTarkenton (approved):
Looks good, should be really easy to implement the changes for the real fam-api here: 
https://github.com/bcgov/nr-forests-access-management/blob/dev/server/backend/api/app/config.py

thinking we could add logic to see if the env vars exist, and if they don't then tries to retrieve from AWS secrets.

PR #173, opened 2022-09-03
changed instance storage type to standard

ConradBoydElliottGustafson:


basilv:
FYI, for production we may need to use provisioned-IIOPS storage for more consistent performance, I had to do that for one app, but it was a lot more sensitive to latency / response times than I think FAM will be.

PR #181, opened 2022-09-08
Richard's July ER Studio Changes

ConradBoydElliottGustafson:

Issue #128, opened 2022-08-24
Define and test database migration flow 


franTarkenton:
**As a** *(User Type/Persona)* **I want** *(Feature/enhancement)* **So That** *(Value, why is this wanted, what is the user trying to accomplish)*

**Additional Context**
- Proposing that the definition of truth be the ORM model
- This ticket will verify that we can create migrations from the ORM model, that can then be transitioned to flyway migrations
- should provide a definition for how this works
- ticket will define how to create an alembic migration ->
- from the alembic migration generate a *flyway* migration

**Acceptance Criteria**
- [ ] Given ?, When a database change is created in the model, there is a defined process for how that migration will get implemented in the database

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

franTarkenton:
* Determined that the @ianliuwk1019 has encorporated the changes in the migrations defined in the PR: https://github.com/bcgov/nr-forests-access-management/pull/100.  Propose we delete that pr
* Figured out how to get alembic to name its migrations in a way that corresponds with the naming convention that flyway wants to use
* identified manual process for how to generate the flyway migrations from alembic migrations
* working on script to automate the conversion from alembic to flyway scripts

franTarkenton:
Noticed that the seed / fixture data identified in the flyway V2 migration is not also included in the alembic migrations.  

The following shows how data can be added using alembic migrations:
https://github.com/bcgov/nr-forests-access-management/blob/e0301b55b329a687a4d3ed62e798b04342b56446/server/backend/api/alembic/versions/V2_first_application.py#L207

ianliuwk1019:
Just a note.
**problem**:
I was running into trouble for using Alembic '--autogenerate' to create a delta migration script. Locally at this moment it did not compare my current db state and create the delta; instead, it creates whole new 'Create' script for all tables.

**Get around the problem**:
Kevin points out he has found a solution for this gotcha issue from Alembic and would include the fix for 'autogenerate'. From talking to Kevin, it seems Alembic only use default schema while our database schema is different '**app_fam**' so Alembic can't detect it to create the delta script. The Context.config will need to include this setting:
**include_schemas=True**
[...]

PR #183, opened 2022-09-09
Feat/db migration

franTarkenton:
see server/backend/api/alembic/dbMigrationsBackground.md for summary of work

closes: https://github.com/bcgov/nr-forests-access-management/issues/128

    defined the ORM model as the source of truth.
    created docs for how to manage the model.py->alembic->flyway migrations (see server/backend/api/alembic/readme.md)
    created script to sync alembic migrations with flyway.
    going to close PR 

https://github.com/bcgov/nr-forests-access-management/pull/181 without merging as all changes are implemented in this PR
modified alembic config to generate migration names using the convention that flyway uses. Makes it easy to map alembic migrations to flyway migrations
this PR establishes a 1:1 relationship between alembic migrations and flyway migrations


ianliuwk1019 (approved):
Thanks for the hard work Kevin!

PR #200, opened 2022-09-21
Added role type

franTarkenton:
* role type was added to model.py
* from the model autogenerated an alembic migration
* from alembic migration autogenerated the flyway migration
* manually reviewed flyway sql and alembic code


rparfig (changes requested):
fam_role has a column that identifies the parent role through a recursive relationship. If the column parent_role_id is NOT NULL, then the role is a "child" role, as the value in that column will allow for the identification of that such parent role. 
I do not think this data model change is necessary

ianliuwk1019:
> fam_role has a column that identifies the parent role through a recursive relationship. If the column parent_role_id is NOT NULL, then the role is a "child" role, as the value in that column will allow for the identification of that such parent role. I do not think this data model change is necessary

Hi @rparfig , perhaps the name Kevin choose isn't catching what our application needs.
What Kevin needs is a column to indicate it needs a child role to associate with this role, when assign role to a user. 

In case of FOM:
A role like "fom_forest_client_00001234" is a child role that it will be linking to "FOM Submitter" role as its parent and a user will be assigned this "fom_forest_client_00001234" role.

rparfig:
Hi @ianliuwk1019 : Thanks for clarifying. The way I understand it then is that this column is more like a flag, rather than a type. Perhaps a name such as requires_child_role_ind would be more proper. The size could be varchar(1), default value "N" and a check constraint that only allows Y/N values. Thoughts?

ConradBoydElliottGustafson:
From a business perspective, there can be a "parent" role that does not (yet) have any children, but is intended to have children. We are identifying an attribute that captures the business logic that states "This role should not be assigned directly, but only by means of assignments to its children".

PR #201, opened 2022-09-22
Bugfix/failing lambda build

franTarkenton:
* resolves the issue that was causing the previous PR builds to fail
* suspect that merging this code will still use the old action that exists in dev, and may fail
* subsequent PR's should start using the actions here instead.


basilv (approved):
I'm no Python expert but these changes look reasonable. I would expect this PR to be using the changed pull request logic but Derek would know for sure.

PR #210, opened 2022-09-28
Feat/add role type 199

franTarkenton:


rparfig (changes requested):
Please remove the restrictive check constraint on role_type_code

server/backend/api/app/models/model.py
```diff
@@ -334,32 +335,60 @@ class FamGroup(Base):
...
-            ["app_fam.fam_forest_client.client_number_id"],
-            name="reffam_forest_client24",
-        ),
-        ForeignKeyConstraint(
-            ["parent_role_id"], ["app_fam.fam_role.role_id"], name="reffam_role23"
-        ),
-        PrimaryKeyConstraint("role_id", name="fam_rle_pk"),
-        UniqueConstraint("role_name", name="fam_rle_name_uk"),
+        PrimaryKeyConstraint("role_type_code", name="fam_role_type_code_pk"),
+        CheckConstraint(role_type_code.in_(['C', 'A'])),
```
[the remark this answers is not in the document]
franTarkenton: Sounds good... should be fixed now

server/backend/api/alembic/versions/V6_add_role_type.py
```diff
@@ -0,0 +1,74 @@
...
+    # ### end Alembic commands ###
+
+    famRoleType = app.models.model.FamRoleType.__table__
+    currentDate = datetime.datetime.now()
+    dateEntry = currentDate
+
+    if isOffline:
+        currentDateStr = currentDate.strftime("%Y-%m-%d %H:%M:%S")
+        dateEntry = op.inline_literal(currentDateStr)
+
```
ianliuwk1019, line 52: Hi Kevin,
Maybe I don't understand why 'isOffline' and online needs different 'dateEntry' value.
Would you explain a little? Thanks, just want to make sure I understand it.

franTarkenton: Hey @ianliuwk1019, so when we are doing development with the containerized postgres database, we can run the alembic migrations directly against the database using a command like: `alembic upgrade head` to run all migratiions, or for a specific version `alembic upgrade V5`.  

Then before we push the code up to github, and create a PR, we need to generate a flyway migration.  The script server/backend/api/alembic/generateFlywayMigrations.py basically runs a "offline" migration, which instead of executing the migration against a database, it outputs a SQL file that it would have normally sent to the database.  That SQL file becomes the flyway migration.

For some reason when the migration is run "offline" (with the --sql switch) I was getting errors when it tried to generate the insert statements that insert dates.  To resolve that, is why i needed to detect online vs offline migration.  

Hope that helps

server/backend/api/app/models/model.py
```diff
@@ -334,32 +335,60 @@ class FamGroup(Base):
...
-            ["app_fam.fam_forest_client.client_number_id"],
-            name="reffam_forest_client24",
-        ),
-        ForeignKeyConstraint(
-            ["parent_role_id"], ["app_fam.fam_role.role_id"], name="reffam_role23"
-        ),
-        PrimaryKeyConstraint("role_id", name="fam_rle_pk"),
-        UniqueConstraint("role_name", name="fam_rle_name_uk"),
+        PrimaryKeyConstraint("role_type_code", name="fam_role_type_code_pk"),
+        #CheckConstraint(role_type_code.in_(['C', 'A'])),
```
ianliuwk1019, line 375: Is the comment out code intended?

server/backend/api/app/models/model.py
```diff
@@ -334,32 +335,60 @@ class FamGroup(Base):
...
+            "comment": "A role type is a code that is associated with roles "
+            "that will influence what can be associate with a role.  At time "
+            "of implementation an abstract role can only have other roles "
+            "related to it, while a concrete role can only have users "
+            "associated with it",
             "schema": "app_fam",
         },
     )
 
+    #fam_role_relation = relationship("FamRole", back_populates="role_type_relation")
```
ianliuwk1019, line 386: Is the comment out code intended?

ianliuwk1019 (approved):
Thanks Kevin!

ianliuwk1019 (dismissed):
Looks good.

rparfig (approved):
Looks great, thanks !

rparfig (approved):

PR #214, opened 2022-09-29
Feat/84 configure lambda

franTarkenton:


ianliuwk1019 (approved):
Approved. Thanks Kevin.
Interesting the trailing '/' can cause issue while it isn't issue at local testing or postman.
Maybe Mangum is messing with this?

ianliuwk1019 (approved):

PR #230, opened 2022-10-11
API Endpoint - Create User Role Assignment

ianliuwk1019:
The pr adds a new API endpoint to **create a user/role assignment** for FAM.
The new endpoint is only suited for first MVP version and is based on some assumption from meeting discussion in this ticket.
The new endpoint is currently implemented as: POST: "**/api/v1/user_role_assignment/**"

### Details:

- DB migrations:
New Alembic migration scripts (V8 to V10):
	- Use surrogate key for user_role_xref
	- Table fam_forest_client: user_name optional optional (This is to accomdiate current MVP before integration with forest client api)
	- Increase fam_user.user_type column length
	
	Generate new Flyway migration script based on Alembic and model.py: V8 to V10

- Code:

	- Add new router to main.py router collection: router_user_role_assignment.py
	- Add new crud_user_role.py to implement logic for verifying request for creating user/role assignment and inserting new
user/role assignment into fam_user_role_xref when needed.
	- Fix, adjust and refactor existing crud code. Also add a new crud_forest_client.py for this new endpoint need.
	- Add a constant.py to hold some temporary constants need for the logic for now.
	- Add new test_crud_role_assignment.py for testing new user role assignment code.
	- Use db.flush() instead of db.commit() for crud code and let db session handle db.commit() at the end of the session.


franTarkenton (approved):
Looks good Ian, longer term thinking we should add some tests for routes for the role endpoints.  It will be interesting to see how the code coverage tests run once we get to that ticket.

going to wait for @rpardofigueroa review of the datamodel before merging

ianliuwk1019:
> Looks good Ian, longer term thinking we should add some tests for routes for the role endpoints. It will be interesting to see how the code coverage tests run once we get to that ticket.
> 
> going to wait for @rpardofigueroa review of the datamodel before merging

Hey @franTarkenton , thank you so much for the review!!
Sorry, I was in a training and reply you late.

As for 'router' tests, that is noted and thank you. 
I didn't provide this time for the reason that I am still not quite sure how to write this and don't want to pollute with bad code. Essentially, I am not sure if there is better way for setting up test cases. For example, "router A (create)" and "router B (Get)", where when setting up "router B" condition for testing "router B" the test case uses (depends on) "router A" to do the testing. I am not used to seeing that and not sure if that's the practice for testing. Maybe it does not hurt to do that since if "router A" test fail, then "router B" test also fail. Anyway, if that's acceptable testing strategy, I can add "router" testing next time.

rparfig (changes requested):
@ianliuwk1019 : Just one small change: The name of the user_type table should be fam_user_type_code. Only the "_code" suffix is missing.

rparfig (approved):

Issue #59, opened 2022-07-07
FAM-API - Manage Role Assignment/ Remove Role Assignment


franTarkenton:
Add end point for:

DELETE - /Role/<roleid>/User/<userid>

removes the user from the defined role.

Relates to: https://github.com/bcgov/nr-forests-access-management/issues/20

PR #239, opened 2022-10-20
#59: API - delete user role assignment

ianliuwk1019:
* Router endpoint and crud code for delete user role assignment.
* Some unit tests and router unit tests added.


ConradBoydElliottGustafson (approved):
I'm not 100% up on how the API code works, so take this review with a grain of salt. Having said that, nothing looks WRONG to me, so I approve. If you think it's safe to merge to DEV then go for it!

Issue #156, opened 2022-08-30
Application selection screen


gormless87:
**As an** Application Access Manager **I want** to see the list of applications I am allowed to manage **So That** I can specify the application I want to manage

**Additional Context**
-- will need to have specified jwt payload
-- access is based on roles being passed via jwt
-- 1 role  per application
-- application id will be parsed out of the role
-- as part of this story, clicking the button leads to a stub 

**Acceptance Criteria**
- [ ] Given I have permission to mange access for one or more apps using FAM, When I login, Then I can select from the list of apps I am allowed to manage
- [ ] Given I have permission to mange access for one or more apps using FAM, When I login, Then I can only see the list of apps I am allowed to manage

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs
- [ ] Merged

PR #246, opened 2022-10-25
Feat/156 Select Application Screen part two

basilv:
Progress on #156, will likely need a 3rd PR after this one though.


basilv:
More refinements coming...

frontend/public/env.js
```diff
@@ -0,0 +1,2 @@
+window.localStorage.setItem('fam_api_base_url', 'https://341ihp76l2.execute-api.ca-central-1.amazonaws.com/test')
```
ianliuwk1019, line 1: Just quick question.
I understand at the moment this is hardcoded.
Will later the value come from Terraform?

frontend/src/App.vue
```diff
@@ -1,19 +1,27 @@
 <script setup lang="ts">
+import { inject } from 'vue'
 import { RouterLink, RouterView } from 'vue-router'
+
+const environmentDisplayName = inject('fam_environment_display_name') as string
+var environmentLabel = ''
```
ianliuwk1019, line 6: Does compiler complaint with using 'let' not 'var'?
Is 'inject' also get value from 'localstorage'?
Also is it possible for Vue to use 'environmentLable' directly for 'inject' (I mean so we don't need to declare two variables)?

ianliuwk1019 (approved):
Approved. Just few simple questions.

frontend/public/env.js
basilv: Yes, it will come from the pipeline.

PR #254, opened 2022-10-28
152 bootstrap theme

ianliuwk1019:
FAM layout.
   (Just first to be aligned for Basil as so we have a common ground)
- Bootstrap 5 installed and configured individually (without using Bootstrap-vue-3: feel it should be ok but reserved for future using it)
- Quick borrowing scss from FOM for making layout not looking too bad (not working initially when copy-paste). So this version is only for a quick tweaks (need a lot of care/clearn/and responsive later, feel might need extra ticket for focusing on this).
- Merge from Basil's work and refactor slightly for menu to be on Header component.


frontend/src/App.vue
```diff
@@ -1,88 +1,30 @@
-<script setup lang="ts">
-import { RouterLink, RouterView } from 'vue-router'
+<script lang="ts">
+
+  import { RouterView } from 'vue-router'
+  import Header from '@/components/header/Header.vue'
+  import Footer from '@/components/footer/Footer.vue'
+
+  export default {
```
basilv, line 7: Pretty sure this 'export default...' is unnecessary, the imports above are good enough, just have to reference the components as e.g. Header and not FamHeader.

basilv (changes requested):

frontend/src/App.vue
ianliuwk1019: Thanks @basilv , and that's a good observation!
According to Vue official document, registration of components locally needs to be done inside "export default"; without the 'setup' at the top of the <script> tag. I think it is because the 'setup' is too late to register a component locally. It is a convenient way to use composition API. (I could do a global registration, I did try global, but I didn't use that way)
Please see here: https://vuejs.org/guide/components/registration.html#local-registration

See if that make sense to you or if I am mistaken with it, please let me know and I can adjust it. Thanks Basil.

basilv: That's true if you are using the Object API, without <script setup>. But you are using <script setup> with the composition API, which does not require then the 'export default...' statement.

basilv: We could debate putting the Header/Footer components into their own subdirectories, since they are both single files, but since they are non-business-related components, I don't mind putting them out of the way. They could also potentially go under the Views directory

ianliuwk1019: Got it, thanks Basil! The Vue documentation API Preference 'toggling' is sure at the confusing spot of the document.
Thank you for pointing that out.
Fixed it and also had some cleanup.

basilv (approved):

PR #257, opened 2022-10-31
Feat/21 auth lookup function

ConradBoydElliottGustafson:
Integrate new auth lookup function code and make tests pass


franTarkenton (approved):
Chatted w/ Conrad, branch is in his name so he cannot provide review,  I made changes so I'm obviously happy with them!  😄

PR #268, opened 2022-11-03
Initial changes for unit testing front-end.

basilv:


franTarkenton (approved):
not an expert, but follow what's going on in all the changes, looks good, looking forward to getting more involved with the vue development

frontend/package.json
```diff
@@ -18,11 +21,16 @@
     "@types/bootstrap": "^5.2.5",
     "@types/node": "^16.11.56",
     "@vitejs/plugin-vue": "^3.0.3",
+    "@vitejs/plugin-vue-jsx": "^2.0.1",
```
ianliuwk1019, line 24: Hi Basil, would you mind sync with 'main' branch again?
I think this '-jsx' is already been removed from main so we can get rid of this one.

frontend/src/App.vue
```diff
@@ -3,13 +3,14 @@
...
   import Footer from '@/components/footer/Footer.vue'
+  import Breadcrumb from '@/components/Breadcrumb.vue'
 
 </script>
 
 <template>
   <Header></Header>
-
   <main class="container pt-0 mt-2">
+    <Breadcrumb />
```
ianliuwk1019, line 13: Right now SnorCloud showing having issue with 'Breadcrum'.
Perhaps for now, maybe do <Breadcrumb></Breadcrumb> will solve it?

frontend/src/components/Breadcrumb.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    router = createRouter({
+      history: createWebHistory(),
+      routes: routes
+    })
+    wrapper = mount(Breadcrumb, {
+      global: {
+        plugins: [router]
+      }
+    })
+    await flushPromises()
```
ianliuwk1019, line 25: Not sure if we are 'await' anything in particular during 'beforeEach' setup?

frontend/package.json
basilv: I did merge main, this change must have gotten missed when I merged.

frontend/src/components/Breadcrumb.test.ts
basilv: It is possible that mounting the component is asynchronous, and thus it is safest to wait until all promises are complete. This was required after calling router.push(), for example.

frontend/src/App.vue
basilv: This is a bug in SonarCloud dealing with Vue3, it doesn't think imports are used sometimes when they are. I discussed this with Derek. For now we are ignoring it as there is a SonarCloud fix for this issue, it just needs to be deployed to go into effect (no timeline given for this).

frontend/package.json
basilv: Pushing a change to address this now.

ianliuwk1019 (approved):
Approved. Thanks Basil.

PR #274, opened 2022-11-07
fix: 270 cloudfront s3 update etag

ConradBoydElliottGustafson:
closes 270


franTarkenton (approved):
assuming that cloudfront refers to objects in the S3 bucket that it needs to serve using the etag?

Issue #260, opened 2022-11-01
Frontend login authentication implementation with JWT token.


ianliuwk1019:
**As an administrator for FAM **I want** to have a login **So That** *I know I can securely log into the FAM system with my credential (IDIR/BCEID) *

**Additional Context**
- Welcome page display some text to 'indicate user login is required'.
- Login (button) is clicked then redirect to Cognito for credential inputs.
- When user is logged in, login button becomes logout button.
- When user is logged in, Welcome page shows 'welcome text about FAM system'
- Log user out when token is expired.

- Http interceptor/handler for frontend to provide system a way to log out upon request based on existence and expiration of the JWT token at header.
- Http interceptor/handler to extract/store user ID/Authorization token with browser to manage user state/session. 

**Acceptance Criteria**
- [ ] Given the user has an IDIR, When the user attempts to login, Then the FAM Welcome page will be displayed (including the login username)
- [ ] Given (Context), When (action carried out), Then (expected outcome)

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

ianliuwk1019:
A Sample Cognito app for reference: https://github.com/bcgov/cognito-example-apps/tree/main/examples.

ianliuwk1019:
The sample react-amplify code (after fixing Cognito domain configuration value) does show (on dev console) there is a request to get token and return this value as response:
[...]

PR #275, opened 2022-11-07
Feat/260 frontend fetch wrapper and auth service

ianliuwk1019:
Part 1 of #260:
* (This one is removed in favor of Axios). Add a Fetch http call wrapper: for adding auth header and intercept request/response if needed.
* Install and use Axios for http calls and use it first on Get Applications endpoint call with simple request interceptor added.
* Add an AuthService: Contains a readyOnly User(token) state with localstorage, and convenient login/logout methods.
* Header menu show/hide based on login/logout (temporarily with fake token).
* A Cognito env is added (probably removed later when impl is clear in future)
* This does not include Auth with Cognito yet (second part).


frontend/src/services/ApiService.ts
```diff
@@ -1,18 +1,21 @@
...
-    private apiUrl:string
+    private _apiUrl:string
 
     constructor() {
         const environmentSettings = new EnvironmentSettings()
-        this.apiUrl = environmentSettings.getApiBaseUrl() + '/api/v1'
+        this._apiUrl = environmentSettings.getApiBaseUrl() + '/api/v1'
     }
 
+    public get apiUrl(): string {
```
basilv, line 13: Not sure why need a public getter for the Url

frontend/src/services/ApiService.ts
```diff
@@ -1,18 +1,21 @@
+import type { Application } from '@/services/ApplicationState';
 import { EnvironmentSettings } from '@/services/EnvironmentSettings';
-import type { Application } from '@/services/ApplicationState'
 
 export class ApiService {
 
-    private apiUrl:string
+    private _apiUrl:string
```
basilv, line 6: Is this a typescript convention? (Prefix private with underscore?)

frontend/src/services/AuthService.ts
```diff
@@ -0,0 +1,56 @@
+import router from '@/router';
+import { reactive, readonly } from 'vue';
+
+// Auth state from localStorage
+const state = reactive({
```
basilv, line 5: You don't necessarily need to use local storage for the logged in user versus just using Vue's global state. 
(I know we used local storage in FOM).

frontend/src/services/AuthService.ts
```diff
@@ -0,0 +1,56 @@
...
+        name: 'fake_user',
+        token: 'fake_token'
+    };
+
+    // update famUser state
+    state.famUser = famUser;
+
+    // store user details and jwt in local storage to keep user logged in between page refreshes
+    localStorage.setItem('famUser', JSON.stringify(famUser));
+    router.push('/');
```
basilv, line 31: Not sure why redirect to home page? What if user is in the middle of a page and login times out? (We probably haven't worried about that flow yet, might be post MVP.)

frontend/src/services/AuthService.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+async function logout() {
+    state.famUser = null;
+    localStorage.removeItem('famUser');
```
basilv, line 36: For vue ref(), should use state.value.famUser... Not sure about the reactive() variant.

frontend/src/services/AuthService.ts
```diff
@@ -0,0 +1,56 @@
...
+    // TODO: Probably need to call Amplify library for sign out from Cognito?
+    router.push('/');
+}
+
+// -----
```
basilv, line 41: This section below confuses me, not sure what you are trying to do with this.

ConradBoydElliottGustafson (commented):
I don't really know the frontend code well enough to review. Leaving it for @basilv

basilv (approved):

PR #289, opened 2022-11-16
Feat/126 app role get

franTarkenton:


ConradBoydElliottGustafson (dismissed):
Running Venv inside a docker container is redundant, but it doesn't break anything.

I didn't do a detailed code review, but it looks like the tests are still running.

ConradBoydElliottGustafson (dismissed):
It's really hard to review a PR when it batches docker stuff, alembic stuff, VS Code stuff, linting and various changes. Small PRs that do a single thing are much easier to review and manage in git.

Having said that, it looks like this PR adds the functionality to the API that we are looking for in the ticket, so I'll approve to get the feature into main.

PR #291, opened 2022-11-18
Feat/260 frontend cognito auth

ianliuwk1019:
The PR is part of the issue #260. This PR includes:
- Implementation for redirection to Cognito/SSO login using AWS-Amplify library.
- Add default env.json (in /public folder)
- Load env.json into env.js for aws server/tf environment settings injection to frontend
	and use these values for Cognito authentication flow by aws-amplify lib.
- Add new Vue route and AuthCallbackHandler Vue component to deal with Cognito callback and deligate 
    it to AuthService for handling postLogin.
- Add NotFound routing, temporarily redirect to home page(can later discuss with in some cases not suitable).
- Parse retrieved token (from Amplify) to frontend object for use later (FamLoginUser).
- Add Vue router beforeEach guard to refreshToken when navigating between pages.
- Add condition to header component for displaying user IDIR name after logged in.
- Add logic to into Axios response interceptor to handle request with 403 (Forbidden) response to 
    refreshToken and retry.


basilv (commented):
Quick glance looks fine, I will review in more detail later. Main concern is figuring out the vite test issue

basilv (dismissed):

ConradBoydElliottGustafson (approved):
Did you mean to check in env.json? If this is a sample with hard-coded values for development purposes, maybe it would be clearer to call it env.json.sample and let the developer rename it locally? env.json is the file that is built dynamically by github actions.

Your call . . . approved anyway.

PR #295, opened 2022-11-21
fix: add environment value to env json file

ConradBoydElliottGustafson:
Easier to export this value from Terraform than to try to write it in with a bash script?


ianliuwk1019 (approved):
Man... I was trying to find this very same solution but didn't know how to!!!
Thanks for simplify that with target_env output!!!

ianliuwk1019 (dismissed):

basilv (approved):

PR #297, opened 2022-11-22
Initial work on error handling 

basilv:
Added support for various message types, not just errors, using 'toast' approach.
Grant access screen has demos.


frontend/src/main.ts
```diff
@@ -11,6 +16,21 @@ Amplify.configure(awsExports); // Config Amplify for Cognito resource.
...
+
+app.config.errorHandler = (err, instance, info) => {
+    // If triggered on component setup/initial render, this can display many times. 
+    // This can trigger on an uncaught error in a function.
+    // This will NOT trigger on an uncaught error in an async setup operation.
+    const toast = useToast();
+    toast.error("An application error has occurred. Please try again. If the error persists contact support.")
+    console.log(`app.config.errorHandler error ${err} with info: ${info}`)
+}
+
```
ianliuwk1019, line 33: This looks interesting to me.
Is the function still under construction?

Is '.config' a global variable for Vue just to hold the "function errorHandler" or is it actually the way to setup some errorHandler for Vue and Vue internally will use it (and for what kind of error)?

What is this "instance" in function parameter? Is there a place that calls this errorHandler?

basilv: No, this is finished. The config.errorHandler is used by Vue internals. So Vue calls it automatically when there are uncaught errors - e.g. see the uncaught error button on the Grant access screen. The 'instance' in the function parameter actually identifies the Vue application, since in theory you can have more than one on a single browser page. Obviously we only have one so we ignore that.

ianliuwk1019 (approved):
Nice lib!

basilv:
Approach documented in wiki at https://github.com/bcgov/nr-forests-access-management/wiki/FAM-Frontend-Design

ianliuwk1019 (approved):
Thanks for also fixing breadcrumb!

PR #302, opened 2022-11-22
Fix/301 add local callback aws dev

ianliuwk1019:
Part of issue#301.
- Add local redirect url for Cognito, dev environment only.
- Minor tweak for import to use alias.


basilv (approved):
Looks good.

PR #308, opened 2022-11-25
Feat/126 app role get

franTarkenton:
Hopefully this is better

### DIFF
--- a/server/backend/api/app/crud/crud_application.py
+++ b/server/backend/api/app/crud/crud_application.py
@@ -1,5 +1,6 @@
 import datetime
 import logging
+from typing import List, Union
 
 from api.app.models import model as models
 from sqlalchemy.orm import Session, load_only
@@ -42,7 +43,10 @@ def getApplicationByName(db: Session, application_name: str):
     )
     return application
 
-def createFamApplication(famApplication: schemas.FamApplicationCreate, db: Session):
+
+def createFamApplication(
+        famApplication: schemas.FamApplicationCreate,
+        db: Session) -> models.FamApplication:
     """used to add a new application record to the database
 
     :param famApplication: _description_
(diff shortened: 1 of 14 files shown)


server/backend/api/app/crud/crud_application.py
```diff
@@ -92,6 +96,31 @@ def deleteFamApplication(db: Session, application_id: int):
...
+
+    application = (
+        db.query(models.FamRole)
+        # .join(models.FamRole)
+        # .options(load_only("application_id"))
+        .filter(models.FamRole.application_id == application_id,
+                models.FamRole.parent_role_id == None)
+        .all()
+    )
+    return application
```
ianliuwk1019, line 121: Hi Kevin, already noted in the meeting.
This probably should be a return of List[models.FamRole] type, and verify if the unit test also pass on this.

server/backend/api/app/schemas.py
```diff
@@ -88,6 +88,13 @@ class FamRoleCreate(BaseModel):
     class Config:
         orm_mode = True
 
+class FamApplicationRoles(FamRoleCreate):
```
ianliuwk1019, line 91: Wondering... you probably meant to use the name: "FamApplicationRole" without "s"?

server/backend/tests/crud/test_crud_application.py
```diff
@@ -1,14 +1,20 @@
...
+import sqlalchemy
 from api.app.crud import crud_application as crud_application
 
 LOGGER = logging.getLogger(__name__)
 
-def test_getFamApplications(dbSession_famApplication_withdata, applicationData1):
-    db = dbSession_famApplication_withdata
+
+def test_getFamApplications(
+    dbSession_famApplication: sqlalchemy.orm.session.Session,
```
ianliuwk1019, line 14: Probably can just use "import" as: ( from sqlalchemy.orm import session )

franTarkenton: > Probably can just use "import" as: ( from sqlalchemy.orm import session )

True... I was thinking in the context of defining the type returned by the fixture that being explicit would provide more clarity around what is being returned.

server/backend/api/app/schemas.py
franTarkenton: Yep... That's now resolved in latest push

franTarkenton: Resolved this in latest push:

ianliuwk1019 (approved):

Issue #178, opened 2022-09-07
Front-end to add users by BCeID /IDIR


ArogeG:
As a application access manager I want to assign users to an application role by BCeID/ IDIR So That users can access an authorized application such as FOM.

**Additional Context**
-- Private page to add the user ID, select the login domain and select the application role, or add/select the forest client role.
--Ability to add the forest client role. If the forest client role does not exists, the backend would create the role. (All the business logic is handled in the API)
--If you create a role assignment that already exits, return a successful result. 
--Upon successfully assigning roles, display a message "Access granted"
--Upon clicking "Grant Access", submit an object the username, roles, and forest client
-Upon clicking "Grant Access" return a response " Access granted" or 'Error Saving' and if successful, clear the user id field and remain on page. 
-Note navigation buttons on the mockup.
--if the forest client already exists, update the table and if not add the forest client.

Note: A separate ticket, the forest client API will be integrated the front -end and the back-end for validation. 
- 

![image.png](https://images.zenhubusercontent.com/62a1029ce9497c39c087ab8b/eba4c54f-1442-4ae5-acf3-ec725331a17c)

**Acceptance Criteria**

 Given I have click on the Grant Access, create an object with user ID, application role and/or forest client as a  database record.
 Given I have logged in successfully with an IDIR, When thee are no applications returned, Then I am advised to check with the FAM Team
**Definition of Done**

 **Ready to Demo in Sprint Review**
 Does what I have made have appropriate test coverage?
 Documentation and/or scientific documentation exists and can be found
 Peer Reviewed by 2 people on the team
 Manual testing of all PRs in Dev and Prod
 Merged

ianliuwk1019:
A side note for AWS dev environment.
* To test with example like "FOM" application, roles needs to exist first (probably need another version of alembic/flyway migration).
I use this script locally for database inserts: 
```
insert into app_fam.fam_role(role_name, role_purpose, application_id, create_user, role_type_code)
values('FOM_SUBMITTER', 'FOM Submitter', 2, 'fam_proxy_api', 'A');

insert into app_fam.fam_role(role_name, role_purpose, application_id, create_user, role_type_code)
values('FOM_REVIEWER', 'FOM Reviewer', 2, 'fam_proxy_api', 'C');
```

* Also, due to some fixes in backend for this ticket, backend needs to be re-dployed to AWS.

ArogeG:
@basilv, is this what I should be looking at in Dev for the Grant Access screen? 

[...]

PR #311, opened 2022-11-28
Feat/178 frontend grant access  page title

ianliuwk1019:
Part of the change for issue #178:
* Use vue-router "meta" to specify "title" for the route.
* Add a "PageTitle" common vue component for different pages to use.

### DIFF
--- a/frontend/src/components/PageTitle.vue
+++ b/frontend/src/components/PageTitle.vue
@@ -1,0 +1,68 @@
+<script setup lang="ts">
+import { useRoute } from 'vue-router'
+import { selectedApplicationDisplayText } from '@/services/ApplicationState'
+
+const props = defineProps<{
+  withoutSelectedApplication?: boolean // TODO: use default value? Vue seems need to use special way.
+}>()
+const displaySelectedApplicationText = !props.withoutSelectedApplication
+
+</script>
+
+<template>
+  <!-- <div class="title" v-if="withoutSelectedApplication">{{useRoute().meta.title}}</div> -->
+  <!-- <div class="title" v-else>{{useRoute().meta.title}} - {{selectedApplicationDisplayText}}</div> -->
+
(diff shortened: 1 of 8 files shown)


frontend/src/router/index.ts
```diff
@@ -71,11 +79,20 @@ router.beforeEach(async (to, from) => {
   useToast().clear()
 
   // Refresh token first before navigation.
-  if (AuthService.state.value.famLoginUser) { // condition needed to prevent infinite redirect
+  if (to.meta.requiresAuth && AuthService.state.value.famLoginUser) { // condition needed to prevent infinite redirect
     await AuthService.methods.refreshToken()
   }
 })
 
+function withMeta(args: any) {
```
basilv, line 87: Two points of confusion - this seems to add on the requiresAuth: true setting, but the name of the function (withMeta) doesn't reflect this, and this method is even used with the additional setting to set requiresAuth to false, which is weird.

frontend/src/router/index.ts
```diff
@@ -57,6 +64,7 @@ const routes = [
   },
   {
     path: "/:catchAll(.*)",
+    meta: withMeta({requiresAuth: false}),
```
basilv, line 67: As mentioned below, this seems wrong to have the withMeta function default apply auth but it is unset here.

frontend/src/components/PageTitle.vue
```diff
@@ -0,0 +1,65 @@
...
+const props = defineProps<{
+  displaySelectedApplication?: boolean
+}>()
+const displaySelectedApplicationText = props.displaySelectedApplication || true // default to true.
+
+</script>
+
+<template>
+  
+  <div class="title" v-if="displaySelectedApplicationText">{{selectedApplicationDisplayText}}</div>
```
basilv, line 14: I'm personally not a fan of the format of the title, but I like the design of making it a separate component and using the Router meta - that's a neat trick!

basilv (changes requested):

basilv (approved):

PR #312, opened 2022-11-29
feat/127 - initial implementation

franTarkenton:
more frequent pr's


ianliuwk1019:
Hi Kevin,
Seems the crud is still in progress.
Is it the tests that need to be reviewed or the schema or, is this pr about refactoring?
Just want to get an idea.

franTarkenton:
Hey Ian, this is my initial implementation of a get role assignments end point.  In the interests of smaller pr's I got it to the point where its working, and tests are passing so I have pushed and created this pr... does that help?

ianliuwk1019:
Sure, I was confused, I will look again.

server/backend/api/app/crud/crud_application.py
```diff
@@ -119,6 +119,35 @@ def getFamApplicationRoles(
     )
     return application
```
ianliuwk1019, line 120: Hi Kevin, just notice this, probably is a variable name change missed, I guess it should be 'roles'?

server/backend/api/app/crud/crud_application.py
```diff
@@ -119,6 +119,35 @@ def getFamApplicationRoles(
...
+    # asking for one() as there should only be one application record where
+    # application_id == {app_id}
+    application = (
+        db.query(models.FamApplication)
+        .filter(models.FamApplication.application_id == application_id)
+        .one()
+    )
+
+    LOGGER.debug(f"application: {application}")
+    return application
```
ianliuwk1019, line 151: I assume it is still in progress but why returning 'application' as what the function needs is "role_assignment"?
It confuses me for the return object. Maybe for the meantime if it is under construction, you can throw error since we don't want other developers to think it is done?

server/backend/api/app/schemas.py
```diff
@@ -172,3 +172,28 @@ class FamForestClientGet(FamForestClientCreate):
 
     class Config:
         orm_mode = True
+
+
+class FamUserRoleXref(FamUserRoleAssignmentGet):
```
ianliuwk1019, line 177: It is confusing from the two names: "FamUserRoleXref" and "FamUserRoleAssignmentGet" , that they seem to be the same purpose....
Sorry Kevin, maybe I am confused, for the rest of the schema objects, not sure if you have time, we can discuss for my understanding how the response object would look like.

basilv:
I've worked on the front-end needing this request, and here's the sample data I'm hoping to get. If the format is different that's fine, but I need e.g. role names, user login id, instead of database primary keys. I haven't checked the PR to see what is being returned currently.

userRoleAssignments.value = [
  {
    assignment_id: 1,
    user_id: 'foo-test',
    user_domain: 'IDIR',
    role: 'Reviewer',
  },
  {
    assignment_id: 2,
    user_id: 'bar-test',
    user_domain: 'BCeID',
    role: 'Submitter',
    forest_client_number: '01234567'
  },
  {
    assignment_id: 3,
    user_id: 'do-test',
    user_domain: 'BCeID',
    role: 'Submitter',
    forest_client_number: '00008765'
  },
]

franTarkenton:
Tried hard to get something that might work for you @basilv.  My latest push will return the following struct on **GET application/<app_id>/user_role_assignements**  (below).  I'm confident I can get to a flatter structure that you identified above, just ran out of time today.  Still got a lot of cleanup to do, but its passing the tests, and its closer to what you were looking for.

``` json

[
    {
        "user_role_xref_id": 1,
        "user": {
            "user_type_code": "B",
            "cognito_user_id": "22ftw",
            "user_name": "Mike Bossy"
        },
        "role": {
            "role_name": "FAM_ADMIN",
            "create_user": "John Doe",
            "role_type_code": "C",
            "client_number": null
        }
    },
    {
        "user_role_xref_id": 2,
        "user": {
            "user_type_code": "B",
            "cognito_user_id": "22ftw",
            "user_name": "Mike Bossy"
        },
        "role": {
            "role_name": "FAM_TEST",
            "create_user": "Patrick Roy",
            "role_type_code": "C",
            "client_number": {
                "client_name": "acme forestry",
                "forest_client_number": 1101
            }
        }
    }
]
```

basilv:
This structure is fine for me, I'd say just stick with this. I'll eventually want forest client name once we populate it properly. I'd like to use user_name as well, but I'm curious when this field is populated, I didn't think we'd have it available. I assume user_type_code B = BCEID, I = IDIR. (Not sure why we don't just spell those out rather than use a one-letter code).

ianliuwk1019:
> This structure is fine for me, I'd say just stick with this. I'll eventually want forest client name once we populate it properly. I'd like to use user_name as well, but I'm curious when this field is populated, I didn't think we'd have it available. I assume user_type_code B = BCEID, I = IDIR. (Not sure why we don't just spell those out rather than use a one-letter code).

@basilv , we actually did have that (BCEID/IDIR, not B or I) as what you said in initial version. But DA consider it should be B or I as code with comment on what B/I is so that's why. I would prefer BCEID/IDIR though.... but maybe that's the standard from DA point of view.

server/backend/api/app/crud/crud_application.py
franTarkenton: Thanks for that @ianliuwk1019, resolved now

basilv (dismissed):

ianliuwk1019 (approved):

basilv (approved):

ianliuwk1019 (approved):

PR #319, opened 2022-11-30
Feat/149 view role assignments

basilv:
Note that bootstrap isn't fully working, I suspect the Javascript portion of bootstrap hasn't been loaded.


frontend/src/components/ManageAccess.vue
```diff
@@ -1,18 +1,146 @@
...
+
+  return "Showing " + visible + " of " + userRoleAssignments.value.length + " records"
+}
+
+function filterIncludes(userRoleAssignment: UserRoleAssignment):boolean {
+
+  // TODO: Review this logic.
+
+  if (userFilter.value != null) {
+    if (!userRoleAssignment.user.cognito_user_id.toLocaleUpperCase().includes(userFilter.value.toLocaleUpperCase())) {
```
ianliuwk1019, line 116: Hm... might not be correct using "cognito_user_id" for filtering.

basilv: I agree, I've raised the question on Kevin's PR - but right now that's the only user id I'm getting back.

basilv: Or maybe you mean filter by user name instead. Yes, that's one of the adjustments I should have considered after adding user name to the display.

frontend/src/components/ManageAccess.vue
```diff
@@ -1,18 +1,146 @@
...
+  user: {
+    user_type_code: string
+    user_type_desc?:string
+    cognito_user_id: string
+    user_name: string
+  },
+  role: {
+    role_name: string
+    role_type_code: string
+    client_number?: {
```
ianliuwk1019, line 21: **client_number** is an object seems to be odd.
I am assuming it will be changed from Kevin's return object.

frontend/src/components/ManageAccess.vue
ianliuwk1019: I am thinking of using "domain"+"user_name" (just like what you display), would that work better for now?

frontend/src/components/ManageAccess.vue
basilv: Yes, I provided that as feedback as well :)

frontend/src/components/ManageAccess.vue
basilv: I think we have differing defintiions of user name. My definition - using IDIR as an example:
user ID : bvandegr
user name: Basil Vandegriend

So we could combine domain + user ID (not user name). But if you want to sort the User ID column to find a particular ID, putting the domain first interferes with that.

ianliuwk1019: I see, can we clarify "User Name" "User Id" label for where the source columns are tomorrow from database point of view?
I don't see you are able to obtain "User Name" as what you indicated at the moment for MVP. The "User Name" is populated as "User ID" in database.

frontend/src/components/ManageAccess.vue
ianliuwk1019: I did suggest Kevin yesterday that if you don't mind, while he is working on real implementation, he can code some "canned data" to return for you right now so you can integrate with the right structure of data from backend and leave the frontend behind when you are done. He will eventually hook up his query and transformation to the json structure you need.
Not sure if that's a good idea you would like, but need to discuss with Kevin.

NickSaglioni (approved):

PR #334, opened 2022-12-12
fix/333 - server/backend dorker-compose runs as api user

franTarkenton:
- next up / the root docker-compose, then new migrations


basilv (dismissed):

basilv (approved):

webgismd:
it should totally be renamed Dorker

PR #352, opened 2023-01-03
Fix/local backend dev troubleshoot

ianliuwk1019:
Fix few backend broken problems so we can continue working on backend locally:
- Add try/except for jwt_validation.py so it does not fail local api start up.
- Fix function call broken probably due to commits being overridden.
- Correct backend readme so alembic migration still can run locally.


franTarkenton (approved):
LGTM, tests pass, alembic migrations can be generated, ./server/backend/docker-compose works again.

Issue #300, opened 2022-11-22
Generate client-side API code


gormless87:
**As a** *Developer* **I want** *to generate the javascript for the front-end* **So That** *I don't have to do it manually*

**Additional Context**
- Currently front-end used Axios library for HTTP request.
- Official openapi-generator reference: https://openapi-generator.tech/docs/generators; and we will use 'typescript-axios' for generator for frontend.
-- OOS - integration with front-end

**Acceptance Criteria**
- [ ] run API locally, backend exposes API spec (json), run generator to create http client for front-end

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

ianliuwk1019:
Found this npm library, [openapi-client-axios](https://www.npmjs.com/package/openapi-client-axios), that may be of more interest to us.

After initial npm installation and do:
```
const api = new OpenAPIClientAxios({ definition: './openapi.json' });
api.init();
```
without any other configuration to quickly see if app run fine, but when frontend server starts, the browser can't open app with error:
```
url.js:3 Uncaught ReferenceError: process is not defined
    at node_modules/@apidevtools/json-schema-ref-parser/lib/util/url.js (url.js:3:29)
    at __require2 (chunk-IKJM6GHQ.js?v=c7842f0a:18:50)
    at node_modules/@apidevtools/json-schema-ref-parser/lib/pointer.js (pointer.js:6:13)
    at __require2 (chunk-IKJM6GHQ.js?v=c
```
Not sure what it is about, need to investigate it.

ianliuwk1019:
I comment out the `new OpenAPIClientAxios() ` and also its `api.init()` and problem persists.

The problem seems to be very likely that the openapi-client-axios uses some lib that uses other lib that is with node environment (function using node 'process') but not compatible with browser; and most likely requires some extra configuration in vite.config.ts to solve the problem.

I can resolve issue as far as in local "npm run dev" to work and being up the frontend page but the production build (although it builds) but open up the frontend page in browser will have other problem:
`
Uncaught TypeError: Cannot convert undefined or null to object
`
To test if production build work, just do `npm run build` and in the 'dist' directory, run `npx http-server -p 5173` to start server.

[...]

PR #353, opened 2023-01-04
Feat/300 openapi client generation

ianliuwk1019:
For ticket #300. 
The pr contains:
- A folder "client-code-gen" under project root directory to hold the area for OpenAPI generated client ( under ./gen folder).
- A script in package.json to generate client code.
- A openapi.json which contains the most current (based on main) FAM Api Swagger spec.
- A api.json which contains options for OpenAPI generator to run.
- Adjust server/backend main.py for FastAPI to use custom_generate_unique_id for better generated function names.
- A readme.


client-code-gen/README.md
```diff
@@ -0,0 +1,52 @@
+# FAM API Client Code Generation for Frontend with "typescript-axios"
+
+This folder contains a mini npm project for generating FAM API client from OpenAPI(Swagger) spec. The genereated client (library) can be utilized for frontend with typescript/axio framework to make api call to FAM backend endpoints.
+
+Client library regeneration is needed when endpints contract has changed.
+
+## Package/Tool choice
+This is using the popular [OpenAPI Generator/openapi-generator-cli](https://openapi-generator.tech/docs/installation) for generation.
+However, it requires java installed in the environment.
```
basilv, line 9: We should have a Dockerfile to do this generation that contains the java environment so we don't need to install java locally and can guarantee the java version used.

client-code-gen/README.md
```diff
@@ -0,0 +1,52 @@
...
+* Javascript (ES6 supported)
+* Typedscript
+
+## Client Generation
+
+The generation utilizes [OpenAPI Generator Cli with "typescript-axios" as the client](https://openapi-generator.tech/docs/generators/typescript-axios).
+
+The *package.json* contains a simple script "gen-api" to run.
+Two things are required to generate the client code:
+* The fam backend OpenAPI spec (.json or .yaml): Default to 'local' file <mark>openapi.json</mark>.
```
basilv, line 28: Is this openapi.json checked in? I don't see it. If it isn't checked in, where does it come from. (And if it were checked in, we'd have to refresh it anyways)

basilv: Oops, I see this is documented a few lines below, so this can be marked as resolved.

client-code-gen/README.md
```diff
@@ -0,0 +1,52 @@
...
+* Generator options: This is currently using a file <mark>api.json</mark> to include options.
+
+Update OpenAPI Spec:
+```
+* Start backend server and navigate to OpenAPI spec at your browser, for example (locally) "http://localhost:8000/openapi.json".
+
+* Copy the json content and update it to the "openapi.json" file at root directory.
+```
+
+To generate, build and compile the typescript sources to javascript use:
```
basilv, line 38: again, this can all be put into the dockerfile and it can do all the work.

client-code-gen/README.md
```diff
@@ -0,0 +1,52 @@
...
+```
+npm install
+npm run gen-api
+
+The generated client is located at "./gen" directory specified by the "-o" option at the script.
+```
+
+## <mark>TODO</mark>: How to Use/Integrate with Generated Client/Lib
```
basilv, line 46: So it looks like we are not yet using this generated API code - I'm fine with this being a separate PR if necessary, but is that being worked on right now?

client-code-gen/README.md
basilv: OK, now I see it checked in. Missed it before I guess.

client-code-gen/openapitools.json
```diff
@@ -0,0 +1,7 @@
+{
```
basilv, line 1: What is this file used for?

basilv (commented):
Additional comment as discussed - need to exclude gen directory from sonarcloud scanning.

client-code-gen/README.md
ianliuwk1019: That's in a separate ticket for integration (I am working on it based on this folder)

client-code-gen/README.md
basilv: As per discussion will do separate PR for the dockerfile.

basilv (approved):
Dockerfile change will be separate PR. SonarCloud issues fixed via exclusion.

PR #356, opened 2023-01-07
Fix for 338

basilv:


server/backend/api/app/crud/crud_application.py
```diff
@@ -145,7 +145,7 @@ def get_application_role_assignments(
 
     crossref = (
         db.query(models.FamUserRoleXref)
-        .join(models.FamUser, models.FamRole, models.FamRoleType, models.FamUserType)
+        .join(models.FamUser, models.FamRole, models.FamRoleType, models.FamUserType, models.FamForestClient)
```
franTarkenton: Hey @basilv, I pulled the changes and ran the tests and am getting a couple failures.   I think the changes to the query here will result in only roles with related forest clients being returned... 

The test sets up:
* a concrete role (no forest client relation)
* an abstract role
* a concrete role, relating to the abstract role, with a forest client

Not sure if i'm correct, but I THINK we want both the concrete roles in the scenario above to be returned.  I just pushed a change that modifies what you did here to an outer join.  All tests continue to pass after that change.

let me know if as per usual I'm confused and missing something.

basilv: Yes, I'm pretty sure you are 100% correct. This particular change may not actually be necessary to fix the defect (versus the schema change), so we could revert entirely, but I never understood how the code was able to return forest client numbers without actually doing the join to the forest client table in the query in the first place. So doing an outer join sounds good to me.

basilv: Nevermind, as I catch up on emails I see that you already pushed the outer join change.

franTarkenton (approved):
LGTM

Issue #351, opened 2023-01-03
Frontend Vue Code Integration with Generated OpenAPI Client (from Backend Api Spec)


ianliuwk1019:
Integrate with Generated client-side API code.
As a Developer I want to use generated api client So That I can quickly make api calls to backend and don't need to worry about writing type/schema and http call implementation.

Additional Context:

Use generated client code based on Typescript and Axios http client.
- Integrate generated code from task: #300 (option: copy code over or can be imported?).
- Make sure we have most up-to-date backend api spec (update it if necessary).
- Supply current custom Axios instance (which has base url path setup for the server, interceptor, auth token etc...) for http setup for the client calls.
- Refactor current frontend service calls (and other code if necessary) to use generated client.
- Make sure http call to backend is successful.

Acceptance Criteria:

Run frontend, visit a page that uses service to call backend and can get back backend api result successfully.

PR #358, opened 2023-01-09
Feat/351 frontend openapiclient integration

ianliuwk1019:
For ticket #351.
This PR includes attempt to use generated OpenApi client lib:
* Downgrade axios version to avoid existing axios bug that prevents initializing generated client using frontend axios instance.
* Rename ApiService to ApiServiceFactory.
* Initialize generated apiServices from 'fam-api' library in ApiServiceFactory with frontend axios instance.
* Import 'fam-api' library to Vue components(Breadcrum, GrantAccess, ManageAccess, SelectApplication, ApplicationState) and refactor for model used.
* Add new npm script 'install-frontend' to include fam-api dependency (client-code-gen) for frontend install.
* Try out using npm run 'install-frontend' at deployment environment.
* Add docker version for OpenAPI client generation and readme update (both bash and Windows options).


client-code-gen/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "OpenAPI Generation for API Client Code",
   "scripts": {
+    "gen-api-docker": "docker run --rm -v ${PWD}:/local openapitools/openapi-generator-cli:v6.2.1 generate -i local/openapi.json -g typescript-axios --config local/api.json -o /local/gen",
```
basilv, line 6: This might not work on Windows, I'd probably just map the current directory as this would normally be run from the client-code-gen directory anyway. Or is this not true for CI build?

frontend/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "fam-frontend",
   "version": "0.0.0",
   "scripts": {
+    "install-frontend": "cd ../client-code-gen/gen && npm install && cd ../../frontend && npm ci",
```
basilv, line 5: Why don't you use 'npm ci' in client-code-gen directory as well?

frontend/src/components/GrantAccess.vue
```diff
@@ -35,25 +38,25 @@ async function grantAccess() {
   const toast = useToast();
   const grantAccessRequest = toRequest(formData.value)
```
basilv, line 39: I'd rename this variable as we are now using a userRoleAssignment(Create) object, not a grant access request.

frontend/src/services/ApiServiceFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+
+   constructor() {
+       // Instanciation for generated 'fam-api' client.
+       this.applicationsApi = new FAMApplicationsApi(undefined, '', httpInstance) // Note, Axios is strange, second parameter needs empty string, not null.
+       this.rolesApi = new FAMRolesApi(undefined, '', httpInstance)
+       this.usersApi = new FAMUsersApi(undefined, '', httpInstance)
+       this.userRoleAssignmentApi = new FAMUserRoleAssignmentApi(undefined, '', httpInstance)
+   }
+
+   getApiService(serviceName: ApiServiceName): FAMApplicationsApi | FAMRolesApi | FAMUsersApi | FAMUserRoleAssignmentApi {
```
basilv, line 30: We could just have a separate getter for each service. I'm not sure what the advantage is of a consolidated method?

basilv (commented):
Looks great, just a few comments.

client-code-gen/package.json
ianliuwk1019: Good point, sorry, I am using Windows WSL system and did not think about cross-platform.
I don't think it would work for Windows. I will need to do new checkout for Windows and try.
However, this might be a hard question (as it always is for cross-platform) as for now I feel using "./" to map current directory would work, may need to find out if there is a simple trick.
Will do some experiment and let you know. Worst case I think, is create another npm script for Windows (on Powershell) to work for code generation on Windows.

frontend/src/services/ApiServiceFactory.ts
ianliuwk1019: Perhaps in this case I am used to Factory Pattern as it feels like it is just a Factory that produces an instance with different flavors (like Shape factory: produce Circle, Square...). There is no advantage or disadvantage difference than using getter as I can tell, but maybe needing a Type cast is an disadvantage, however class object is much cleaner since almost no logic at all. Do you like me to switch to getter, I don't mind, if you prefer that way?

frontend/package.json
ianliuwk1019: Use ci now, had that thought too but forgot to change.

frontend/src/components/GrantAccess.vue
ianliuwk1019: Variable is renamed. Could you see if it is good? Thank you Basil.

client-code-gen/package.json
basilv: You could have a second npm script that just uses the local directory, and keep gen-api-docker for the CI build (maybe rename accordingly).

frontend/src/components/GrantAccess.vue
basilv: Good!

frontend/src/services/ApiServiceFactory.ts
basilv: It is still factory pattern either way, just making it simpler for the caller to have the getters.
E.g. factory.getApplicationApi() vs factory.getService(ServiceName.ApplicationApi) as FamApplicationApi

basilv: So I'd prefer the getters.

ianliuwk1019: No problem. I changed that so caller is easy to use this.

client-code-gen/package.json
ianliuwk1019: Yes, that's what I thought easiest option and I added the Windows option in (tested and worked for both Powershell and CMD prompt). Please have a look.

basilv (approved):

PR #360, opened 2023-01-10
fix: 104 - api  changes to support roles with forest client

franTarkenton:


franTarkenton:
This PR. is the updates to the api.  The create role end point has been modified so that if a requested role includes a forest client number, then the forest client table will be queried for the related forest client number.  If it does not exist it will be created.

ianliuwk1019 (approved):
PR looks great, Kevin. Thank you.

PR #364, opened 2023-01-13
347 apply security to all fam api endpoints

ConradBoydElliottGustafson:
- [x] Applied security to all the endpoints that we are using. 
- [x] Wrote a bunch of new code to secure the "User Role Assignment" endpoints (needed logic)
- [x] Made all the "Role" and "User" endpoints return 501 (since we are not using them)
- [x] Fixed all the unit tests
- [x] Refactored all the tests to live in the same directory


ianliuwk1019 (approved):
Thanks Conrad for the tricky security implementation.
I like your mock unit tests.
Appreciate your quality code although some SonarCloud code smell (that I understand due to some commented out code) otherwise, no blocking issue I can see, so approved.

PR #405, opened 2023-01-30
Fix/389 timestamp

MCatherine1994:
- update timestamp column to use timestamptz, to include the time zone
- update the default value of the timestamp column to use current_timestamp (it's the same as now(), now() is a traditional PostgreSQL equivalent to CURRENT_TIMESTAMP), so it includes a time zone


server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
```diff
@@ -0,0 +1,80 @@
...
+ALTER TABLE app_fam.fam_group ALTER COLUMN create_date SET DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE app_fam.fam_group ALTER COLUMN update_date SET DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE app_fam.fam_group_role_xref ALTER COLUMN create_date SET DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE app_fam.fam_group_role_xref ALTER COLUMN update_date SET DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE app_fam.fam_role ALTER COLUMN create_date SET DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE app_fam.fam_role ALTER COLUMN update_date SET DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE app_fam.fam_role_type ALTER COLUMN effective_date SET DEFAULT CURRENT_TIMESTAMP;
```
MCatherine1994, line 69: Currently there is no default value for the effective_date in the fam_role_type table, but I saw other tables who use effective_date (fam_app_environment and fam_user_type_code) both have a default value, so I added a default value here as well

server/backend/env-docker-compose.env
```diff
@@ -1,5 +1,5 @@
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD=postgres
+POSTGRES_PASSWORD=test
 POSTGRES_HOST=localhost
```
MCatherine1994, line 3: change to match the password in the docker-compose file in the root directory, for local Postgres setup

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
       - fam
 
   fam-flyway:
-    image: flyway/flyway:latest-alpine
+    image: flyway/flyway:latest
     container_name: fam-api-migrations
```
MCatherine1994, line 26: the "flyway/flyway:latest-alpine" image doesn't work for Mac, so changed

server/backend/env-docker-compose.env
ianliuwk1019: No objection here but I thought normally people use default-user "postgres" with password also as "postgres" for easy to remember and use. 
"test" probably was intended for api-proxy user. But see what other people say.

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
ianliuwk1019: I assume "effective_date" and "update_date" don't need default value as they will be updated by insert statement or application.

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
```diff
@@ -0,0 +1,80 @@
+-- Running upgrade V19 -> V20
+
+-- Change date type to timestamptz for create_date and update_date
```
ianliuwk1019, line 3: This is probably extra, or maybe no need to do it. From my understanding from Basil's safe concerns, would it be safer to also set postgres database timezone to UTC (as using "ALTER DATABASE...")?

docker-compose.yml
basilv: This is just a docker-compose change so won't impact anything other than local developer setup. I'm fine with this.

server/backend/env-docker-compose.env
basilv: Makes sense to be consistent, but I can see Ian's point. Again just for local so I have no personal preference here. To make Ian happy could change both passwords to just be 'postgres' instead of 'test'.

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
basilv: Database should already default to UTC.

ianliuwk1019:
Although not sure how soon the migration strategy would be changed, but should this one also include "Alembic" migration script to avoid (just for now) version conflict for V20?

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
basilv: Yes, I think we have update_date be null for newly created records, in which case there should be no default set for update columns.

basilv (changes requested):

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
MCatherine1994: yeah, I would agree, but I saw V4, V10, and V18 already set the default value for update_date and effective_date to LOCALTIMESTAMP. Maybe here I should cleanup all the default value?

server/backend/env-docker-compose.env
MCatherine1994: sure, I could update the docker-compose file to use password 'Postgres'

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
MCatherine1994: Based on our conversation will keep it as now.

server/flyway/sql/V20__fix_timestamp_to_timestamptz.sql
MCatherine1994: yeah, think we discussed this yesterday, for best practice, change to use timestamptz, so it's ensured and always come up with a timezone.

ianliuwk1019 (approved):

basilv (approved):

Issue #386, opened 2023-01-24
Blank browser screen on initial app launch in new browser


basilv:
Describe the Bug
When I launch the app by opening a new browser (e.g. after clearing history) and type in URL, I get a blank page. The web console shows the following error:
Uncaught TypeError: Cannot read properties of null (reading 'fam_cognito_region')

Expected Behaviour
The application screen displays properly. This does happen when I refresh the page.

Actual Behaviour
Blank screen is displayed. In the code, it appears it is initializing main.ts and specifically awsExports before env.js is called to initialize the environment. Perhaps move initializing of env.js into main.ts???

Additional context
Observed in Edge and Chrome.

ianliuwk1019:
PR is ready.
In summary:
From observations, this initial blank page only happens during user's "**first time**" (particularly when no browser caching from previous loading) accessing the FAM site.
Found that web browser stack can use "**preload**" for preloading some content for performance and that seems to solve the problem. Tested few times on browsers (Chrome/Edge) with clear caching and inPrivate mode, I observed it is consistent now with no blank page issue.
Hopefully it does solve the problem.

![image.png](https://images.zenhubusercontent.com/62a9ff93d8506f14758ef03a/5ef17d5e-a5b8-458b-af37-9bf9b03118ee)
As can be seen above, **"env.json"** is preloaded and ready first, and main **"js"** code is then queued and loaded later.

gormless87:
@basilv  - please review

PR #413, opened 2023-01-31
386-init-app-blank-page

ianliuwk1019:
This is for ticket #386.
From observations, this initial blank page only happens during user's "**first time**" (particularly when no browser caching from previous loading) accessing the FAM site.
Found that web browser stack can use "**preload**" for preloading some content for **performance** and that seems to solve the problem. Tested few times on browsers (Chrome/Edge) with clear caching and inPrivate mode, I observed it is consistent now with no blank page issue.

### DIFF
--- a/frontend/src/aws-exports.js
+++ b/frontend/src/aws-exports.js
@@ -1,3 +1,12 @@
+/**
+ * This aws-exports is the configuration for Amplify library to communicate with AWS Cognito.
+ * It uses generated (different in each environment) "env.json" static file that is
+ * being loaded and stored in user's browser storage at run time. (See env.js at index.html)
+ *
+ * A note for purpose of using optional ("env?.") is that this is not to skip values when "env"
+ * is not loaded. When there is no "env.json" loaded yet, the error message is strange; so give it
+ * optional "?" simply let the down stram Amplify to throw more meaningful error.
+ */
 const env = JSON.parse(window.localStorage.getItem('env_data'))
 
 const config = {
@@ -14,5 +23,5 @@ const config = {
     },
     federationTarget: 'COGNITO_USER_POOLS',
   };
-  
+
   export default config;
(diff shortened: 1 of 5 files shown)


frontend/src/aws-exports.js
```diff
@@ -1,15 +1,15 @@
 const env = JSON.parse(window.localStorage.getItem('env_data'))
 
 const config = {
-    aws_cognito_region: env.fam_cognito_region.value,
-    aws_user_pools_id: env.fam_user_pool_id.value,
-    aws_user_pools_web_client_id: env.fam_console_web_client_id.value, // This is App Client Id
+    aws_cognito_region: env?.fam_cognito_region.value,
+    aws_user_pools_id: env?.fam_user_pool_id.value,
+    aws_user_pools_web_client_id: env?.fam_console_web_client_id.value, // This is App Client Id
```
basilv, line 6: Not sure why this was made optional. Nothing will work if the env data is missing. Does it make for nicer error messages or something?

basilv (commented):
Nice solution, I'm unsure whether it will work on all browsers but the issue should be relatively rare to occur so I think it should be good. Just one question on the optional env?

frontend/src/aws-exports.js
ianliuwk1019: You always have good instinct Basil!

The solution so far is the only solution that works for me. I tried moving around Amplify initialization but nothing works as env.json always not catching up with initialization (loading timing). I seemed once see Vue/Vite talk about Webpack preload setting but was hesitant to try (I think we don't use webpack with Vite or maybe internally it is, don't know).

To answer you question, yes, I faced problem when moving Amplify init into AuthService component and break the test and had to use this "?/optional" for the aws-exports.js configuration. Then when pipeline was running it failed (test pass) on timeout from Amplify and showing message that makes sense more (something like it can not initialize with Cognito). So leaving it optional there would make down stream error to show up nicely.

basilv: Works for me then. I'd suggest adding a comment explaining this then, so future developers aren't left wondering.

ianliuwk1019: Comment added. Thanks Basil.

basilv (approved):

PR #414, opened 2023-01-31
fix: refactored test env variables

ConradBoydElliottGustafson:


server/backend/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.8
+
+RUN addgroup -S nonroot \
+    && adduser -S nonroot -G nonroot
```
ianliuwk1019, line 4: Hi Conrad, not sure why but I just checked out and tried to run (docker-compose up) and saw this error:
`executor failed running [/bin/sh -c addgroup -S nonroot     && adduser -S nonroot -G nonroot]: exit code: 1`

server/backend/local-dev.env
```diff
@@ -4,9 +4,6 @@ POSTGRES_HOST=localhost
 POSTGRES_DB=fam
 POSTGRES_DB_TEST=fam_test
 POSTGRES_PORT=5432
-TMP_FLYWAY_DIR=../../server/flyway/sql
-api_db_username=fam_user
-api_db_password=fam_user_password
```
ianliuwk1019, line 9: Not sure if these "api_db_username" and "api_db_password" are safe to be removed from environment file; backend config.py seems to be reading that. Maybe this "api_db_username" should instead be of value: "fam_proxy_api"?

server/backend/Dockerfile
ianliuwk1019: Hello Conrad, I commented out the two docker commands for "USER nonroot".
It seems to be running fine for me now but I don't know why the USER is needed.
Please try on your end to see if it still works for you.
Thank you.

server/backend/local-dev.env
ConradBoydElliottGustafson: TMP_FLYWAY_DIR is only used for the Alembic stuff that we are no longer using
api_db_username and api_db_password are indeed used in the config file, but it's still working without them because there are defaults. I'll add them back in for clarity.

server/backend/Dockerfile
ConradBoydElliottGustafson: It works fine without them, but the checks don't pass on Github because SonarCloud throws up a security error. I'll see if I can figure out how to disable that error.

server/backend/local-dev.env
ConradBoydElliottGustafson: Actually, I was right. api_db_username and api_db_password are not used by that docker container.

ianliuwk1019 (approved):

server/backend/Dockerfile
basilv: Its considered a security best practice to avoid Dockerfiles using root user. If the dockerfile is only used locally this doesn't matter in my opinion and we could just suppress the SonarCloud warning. But for dockerfiles used to build images used on AWS, I'd push for running as a different user.

PR #420, opened 2023-02-06
Update lambda timeout

basilv:


ConradBoydElliottGustafson (approved):
15 is a good temp number, but we should probably figure out the underlying issue as well.

ianliuwk1019 (approved):

PR #422, opened 2023-02-07
Logging and performance improvements

basilv:


basilv:
I'm uncertain about the len() syntax in my revised logging statements, please double-check that's right.

franTarkenton:
> I'm uncertain about the len() syntax in my revised logging statements, please double-check that's right.

Looks right to me!

ianliuwk1019 (approved):
Looks good! Thanks for performance tweaking.

PR #423, opened 2023-02-07
SQL logging and improving get user role assignment performance

basilv:


basilv:
Added eager loading to have get user role assignments API call now use a single DB query.

ianliuwk1019:
Good to know there is an option "lazy="joined" from SqlAlchemy.
Just wondering, does that make internally the magic sql different now and no N database queries behind the scene (how does the sql looks like)?

server/backend/api/app/database.py
```diff
@@ -9,6 +9,9 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# Log SQL queries from SQLalcheny
+logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
```
ianliuwk1019, line 13: Nick says we can use `engine = create_engine(SQLALCHEMY_DATABASE_URL, echo=**True**)`

basilv: Yes, the documentation suggested this was more for local development than production (as echo = true writes to standard out), but then other documentation suggested it was equivalent to the change I did, so it wasn't entirely clear which was best. I figured the approach I picked was clearer to understand.

basilv:
(And yes, now just a single database query for get user role assignments API call.)

ianliuwk1019 (approved):
This is really useful. Thanks Basil!

server/backend/api/app/database.py
franTarkenton: ` LOGGER = logging.getLogger(__name__)` will create a logger for the same name as the module, in this case that would be `api.app.database`.  The logging config can all be defined in the `https://github.com/bcgov/nr-forests-access-management/blob/main/server/backend/api/config/logging.config`.  If you want to control the logging for 'sqlalchemy.engine' module, I think you should be able to create another logging config section in the logging.config file.

I'm guessing ya'll are trying to turn down the log level.. I'd start by changing the log level for the root logger to INFO, here: https://github.com/bcgov/nr-forests-access-management/blob/0c35e9818e945d6407acf7f72e5eda03050c1fdb/server/backend/api/config/logging.config#L14

Hope this helps... just caught my attention, may misunderstand what you guys are trying to do.  Ping me if you have any questions. (discord / teams)

PR #431, opened 2023-02-09
fix/339 - Return Abstract FAM Role in the Get Role Assignment Endpoint

NickSaglioni:
- Adjusting get_application_role_assignments to return parent_role info and optimizing joins;
- Updating schema.py
- Changing frontend to display parent role name when there is one


ianliuwk1019 (approved):
Looks good. Thanks Nick.

PR #433, opened 2023-02-10
FOM: add custom:idp_display_name mapping and callback_url2

ianliuwk1019:


basilv (dismissed):
The dev_fom_oidc_client is missing the dev PR-specific URLs. Where you planning to see if a wildcard works? I also don't really get why the oidcdebugger is included, we should probably document that somewhere if it going to be 'officially' provides as a tool for all consumer applications. But I'll approve this PR anyways, as the main change to add display name looks fine, and we can always tweak the callback URL lists later.

ianliuwk1019:
> The dev_fom_oidc_client is missing the dev PR-specific URLs. Where you planning to see if a wildcard works? I also don't really get why the oidcdebugger is included, we should probably document that somewhere if it going to be 'officially' provides as a tool for all consumer applications. But I'll approve this PR anyways, as the main change to add display name looks fine, and we can always tweak the callback URL lists later.

Thanks Basil! I mentioned this is not possible from Cognito to use wildcard for redirect URL for security reason in the ticket. So for now we may live without dev.... 
I leave the 'oidcdebugger' in as it might be helpful, but I didn't put it into prod. I will discuss this with Conrad.

MCatherine1994 (approved):

Issue #430, opened 2023-02-09
Add redirect callback url and display name for FOM in the infrastructure config file 


MCatherine1994:
Add the redirect callback and sign out urls to the config file for FOM client, so it won't disappear after redeployment. 

Add attribute mapping for display name vs custom:idp_display_name

ianliuwk1019:
For Cognito "redirectSignin" Url setting for FOM (for admin frontend, only admin needs signin). Confirm these URL with Basil:

Dev (question):
https://fom-dev.nrs.gov.bc.ca (? or is it dynamic route based on PR number?)
or just use http://localhost:4200?

Test:
https://fom-test.nrs.gov.bc.ca

Prod:
https://fom.nrs.gov.bc.ca

PR #436, opened 2023-02-15
Feat/logout chain

MCatherine1994:


ConradBoydElliottGustafson (commented):
I feel unqualified to approve this PR (because front-end). ;-)

@ianliuwk1019 can you please review?

ianliuwk1019:
> I feel unqualified to approve this PR (because front-end). ;-)
> 
> @ianliuwk1019 can you please review?

No problem at all, give me some time.

frontend/src/services/AuthService.ts
```diff
@@ -1,56 +1,91 @@
...
+    domain: env?.fam_cognito_domain.value,
+    region: env?.fam_cognito_region.value,
+    userPoolWebClientId: env?.fam_console_web_client_id.value,
+  };
+  const keycloakConfig = {
+    url: "https://dev.oidc.gov.bc.ca/auth",
+    realm: "ichqx89w",
+    siteMinderUrl: "https://logontest7.gov.bc.ca",
+  };
+
```
ianliuwk1019, line 57: Do we need these settings as environment variables (or in a file) as well since it is only coded for KeyCloak '**dev**'. I guess there is also 'test' keycloak, correct (just asking, not sure)?

frontend/src/services/AuthService.ts
```diff
@@ -1,56 +1,91 @@
...
+    keycloakConfig.realm +
+    "/protocol/openid-connect/logout?post_logout_redirect_uri=" +
+    cognitoLogoutUrl;
+
+  const siteMinderLogoutUrl =
+    keycloakConfig.siteMinderUrl +
+    "/clp-cgi/logoff.cgi?retnow=1&returl=" +
+    keycloakLogoutUrl;
+
+  router.push(siteMinderLogoutUrl);
```
ianliuwk1019, line 76: Does the `router.push()` works!? or maybe it does work for router push to trigger browser redirect (I thought it does not work that way previously)?

ianliuwk1019 (changes requested):
There is a build fail error on 'Type Error', could you also look into that?
Thank you Catherine.

frontend/src/services/AuthService.ts
MCatherine1994: it works, it will redirect to our home page

frontend/src/services/AuthService.ts
MCatherine1994: yes, sorry I missed this part, maybe we need to setup some env variables based on different environment. Or get them from backend

frontend/src/services/AuthService.ts
```diff
@@ -1,56 +1,91 @@
...
-    Auth.signOut()
-    removeFamUser()
-    console.log("User logged out.")
-    router.push('/')
+  Auth.signOut();
+  removeFamUser();
+  console.log("User logged out.");
+
+  // todo: refactor this part, should get these config variables from backend
+  const env = JSON.parse(window.localStorage.getItem("env_data"));
```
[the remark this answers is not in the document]
MCatherine1994: yeah, just like FOM, we get all these config values from backend api, not sure if we want to do the same for FAM, or maybe in another ticket

frontend/src/services/AuthService.ts
```diff
@@ -1,56 +1,76 @@
...
+  Auth.signOut();
+  removeFamUser();
+  console.log("User logged out.");
+
+  // todo: refactor this part, should get these config variables from backend
+  const env = JSON.parse(window.localStorage.getItem("env_data") || "");
+  const cognitoConfig = {
+    domain: env?.fam_cognito_domain.value,
+    region: env?.fam_cognito_region.value,
+    userPoolWebClientId: env?.fam_console_web_client_id.value,
```
ianliuwk1019, line 50: This "?" optional was used for special reason at aws-exports; I think it is better removing "?" optional if env is not present.
Otherwise looks good to me!

frontend/src/services/AuthService.ts
```diff
@@ -1,56 +1,76 @@
...
+  };
+
+  const postLogoutUrl = window.location.origin;
+  const cognitoLogoutUrl =
+    `${cognitoConfig.domain}.auth.${cognitoConfig.region}.amazoncognito.com/` +
+    `logout?client_id=${cognitoConfig.userPoolWebClientId}&logout_uri=` +
+    postLogoutUrl;
+
+  const logoutUrl = env?.frontend_logout_chain_url.value + cognitoLogoutUrl;
```
ianliuwk1019, line 59: Same comment above.

ianliuwk1019 (changes requested):

ianliuwk1019 (approved):

MCatherine1994:
> I feel unqualified to approve this PR (because front-end). ;-)
> 
> @ianliuwk1019 can you please review?

Hi @ConradBoydElliottGustafson Conrad, I added some env variables in the terraform config, do you mind take a quick look? Thanks!

MCatherine1994:
I wonder if we want to setup the logout url for FOM as well in the terraform config `infrastructure/server/oidc_clients_fom.tf`, wondering what's the base url for FOM?

https://fom-test.apps.silver.devops.gov.bc.ca/admin? and https://fom-prod.apps.silver.devops.gov.bc.ca/admin?

frontend/src/aws-exports.js
```diff
@@ -18,7 +18,7 @@ const config = {
         domain: `${env?.fam_cognito_domain.value}.auth.ca-central-1.amazoncognito.com`,
         scope: ['openid'],
         redirectSignIn: `${env?.front_end_redirect_base_url.value}/authCallback`, // For some reason, vue nested path (/cognito/callback) does not work yet.
-        redirectSignOut: `${env?.front_end_redirect_base_url.value}`,
+        redirectSignOut: `${env.frontend_logout_chain_url.value}${env?.front_end_redirect_base_url.value}`,
```
ianliuwk1019, line 21: It's very minor, but let's add `?.` to the `${env` for the first part of the url.

ianliuwk1019 (changes requested):

ianliuwk1019:
> I wonder if we want to setup the logout url for FOM as well in the terraform config `infrastructure/server/oidc_clients_fom.tf`, wondering what's the base url for FOM?
> 
> https://fom-test.apps.silver.devops.gov.bc.ca/admin? and https://fom-prod.apps.silver.devops.gov.bc.ca/admin?

It's good to add logout rul in I think. These are the base url for FOM below, but we won't be able to set it up for FOM Dev environment as it is based on its PR number (value changes) and I indicated in ticket #430 that AWS Cognito does support wildcard for redirect url.:
```
Dev:
	https://fom-[pr#].apps.silver.devops.gov.bc.ca

Test:
	https://fom-test.nrs.gov.bc.ca

Prod:
	https://fom.nrs.gov.bc.ca
```

infrastructure/server/oidc_clients_fom.tf
```diff
@@ -7,6 +7,9 @@ resource "aws_cognito_user_pool_client" "dev_fom_oidc_client" {
     "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/",
     "http://localhost:4200/admin"
   ]
+  logout_urls                                   = [
+    "${var.frontend_logout_chain_url}http://localhost:4200/admin/not-authorized?loggedout=true"
+  ]
```
MCatherine1994, line 12: I assume all terraform files can read the variable "var.frontend_logout_chain_url"?

basilv: Yes, they should be able to. The Terraform CI jobs would fail if they can't find a variable.

basilv (dismissed):

ianliuwk1019 (approved):

infrastructure/server/oidc_clients_fam.tf
```diff
@@ -11,8 +11,8 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
     "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/"
   ]
   logout_urls = [
-    "${var.front_end_redirect_path}",
-    "http://localhost:5173"
+    "${var.frontend_logout_chain_url}${var.front_end_redirect_path}",
+    "${var.frontend_logout_chain_url}http://localhost:5173"
```
ConradBoydElliottGustafson, line 15: Hmm. Is the localhost redirect going to make it into production? I think it is. Not sure how to make sure this only appears in dev and test. Any ideas?
I guess the way to do it is to put it into the Terragrunt files instead of in the infrastructure file. That way we can add the localhost to dev and test, but not production.

ConradBoydElliottGustafson (changes requested):
Let's try to keep localhost redirects and logout urls out of production. See comment on oidc_clients_fam.tf

infrastructure/server/oidc_clients_fam.tf
ianliuwk1019: Good catch!
Thinking.... maybe we don't even need localhost one...

MCatherine1994: um... but in the callback_urls we also have localhost? 
Screen Shot 2023-03-02 at 12 45 13 PM

ConradBoydElliottGustafson: Yeah, I missed that. Is we leave it in that file, it will go to production. It's a security risk. Both the callback_urls and the logout_urls should not have localhost in production. One could argue that the logout_urls doesn't really matter so much from a security perspective, but the localhost one is definitely a security risk.

MCatherine1994: Thanks for the explaination Conrad! I'll think a bit how to do this

ConradBoydElliottGustafson: Should be able to move the whole "list" to the terragrunt file and then specify it individually per environment. The syntax is the same I believe.

infrastructure/server/oidc_clients_fam.tf
```diff
@@ -3,16 +3,22 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
...
     "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/"
   ]
-  logout_urls = [
-    "${var.front_end_redirect_path}",
-    "http://localhost:5173"
+  logout_urls = var.fam_console_idp_name == "PROD-IDIR" ? [
+    "${var.frontend_logout_chain_url}${var.front_end_redirect_path}"
+  ] : [
+    "${var.frontend_logout_chain_url}${var.front_end_redirect_path}",
+    "${var.frontend_logout_chain_url}http://localhost:5173"
```
MCatherine1994, line 21: Hi @ConradBoydElliottGustafson @ianliuwk1019, how about define the value based on a condition? Shall we do it this way? We have an existing variable fam_console_idp_name, and its values are "DEV-IDIR", "TEST-IDIR", "PROD-IDIR" based on different environment, shall we check that, and then decide if include localhost or not?

ianliuwk1019: Hi Catherine, looks interesting! TF can do this syntax as well.
I might prefer to code them([...]) in terragrunt files (easier to find them and, add and remove) but leave it for you and Conrad to decide :)

terraform/prod/terragrunt.hcl
```diff
@@ -20,8 +20,16 @@ generate "prod_tfvars" {
...
+  frontend_logout_chain_url = "https://logon7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri="
   front_end_redirect_path = "https://fam.nrs.gov.bc.ca"
   local_frontend_redirect_path = "http://localhost:5173"
+  fam_callback_urls = [
+    "https://fam.nrs.gov.bc.ca/authCallback",
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/"
+  ]
+  fam_logout_urls = [
+    "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://dev.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=https://fam.nrs.gov.bc.ca",
+  ]
```
ianliuwk1019, line 32: Hi Catherine, this "https://dev.loginproxy....." looks like is for "**dev**" and should not be in /**prod**/terragrunt.hcl.
Also, not sure if "https:/logon**test**7.gov.bc.ca" is wrong in /prod/terragrunt.hcl.
Could you verify and modify it to be production environment?
Thank you Catherine.

terraform/test/terragrunt.hcl
```diff
@@ -16,7 +16,18 @@ generate "test_tfvars" {
...
+  fam_callback_urls = [
+    "https://fam-tst.nrs.gov.bc.ca/authCallback",
+    "http://localhost:5173/authCallback",
+    "http://localhost:8000/docs/oauth2-redirect",
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/"
+  ]
+  fam_logout_urls = [
+    "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://dev.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=https://fam-tst.nrs.gov.bc.ca",
+    "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://dev.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=http://localhost:5173"
+  ]
```
ianliuwk1019, line 30: Hi Catherine, this "https://dev.loginproxy....." looks like is for "dev" and should not be in /**test**/terragrunt.hcl.
Could you verify and modify it to be production environment?
Thank you Catherine.

ianliuwk1019 (changes requested):
Hi Catherine, just need few adjustment.
There is also this value (from key: "**frontend_logout_chain_url**") "https://logon7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=" repeated several time in terragrunt files. I am not sure if this can be in one place (by using terragrunt "**input**" or "**local**"? I am not so sure the usage) but we can leave it for next time if too much trouble.

ianliuwk1019 (approved):
Thanks Catherine! Approved! :)

ConradBoydElliottGustafson (approved):
Nicely done! Sorry for being so picky

PR #437, opened 2023-02-15
Add write permission for mapping on "custom:idp_display_name"

ianliuwk1019:
- Remove "custom:idp_display_name" from general min list.
- Add "custom:idp_display_name" to oidc_clients_fom.tf


ianliuwk1019:
> I think the minimum read list and the minimum write list should stay "minimum". FAM doesn't need the display name (for now) and adding it to the minimum lists takes away our ability to have a "default" list of attributes that works. It would be better to add that attribute specifically for the FOM clients for the read_attributesj and write_attributes properties.

Yup, I was wondering that but thought maybe display_name is quite often that many apps needs. Sure, I will find a way to do that. Thanks Conrad!

ConradBoydElliottGustafson (approved):
Looks awesome!

PR #447, opened 2023-02-21
Try restrict allow origin

ianliuwk1019:


basilv (changes requested):
On second thought, doesn't API gateway have to be open to users coming from anywhere. The front-end of FAM URL is where users go to load the front-end, but traffic to the API doesn't come from there it comes from the user's browser.

ianliuwk1019:
Close this one in favor of diving and investigating more at the ticket #472

PR #448, opened 2023-02-22
Feat/367 save user (API Requester) with cognito user

ianliuwk1019:
This PR contains:
- extracts "username" from access_token claim as a requester to API. This is Cognito's user id.
- obtains "requester" (username) as router "Depends".
- saves API requester into db various tables' create_user field.
- Flyway script v22 to increase create_user/update_user length for saving this Cognito user id.


NickSaglioni (approved):
Everything seems to be ok. Thanks.

PR #459, opened 2023-02-24
Fix/394: Frontend code quality improvement: using vue3 composition api

NickSaglioni:
- Changing Dialog.vue from option api to composition api;
- Correctiong typo in GrantAccess.vue;

### DIFF
--- a/frontend/src/components/Dialog/Dialog.vue
+++ b/frontend/src/components/Dialog/Dialog.vue
@@ -35,29 +35,13 @@
     </vue-final-modal>
 </template>
 
-<script>
-export default {
-    inheritAttrs: false,
-    emits: ['confirm'],
-    props: {
-        message: {
-            type: String,
-            default: '',
-        },
-        title: {
-            type: String,
-            default: '',
-        },
-        confirmText: {
-            type: String,
-            default: 'Confirm',
-        },
-        cancelText: {
-            type: String,
-            default: 'Cancel',
-        },
-    },
-};
+<script setup lang="ts">
+const props = defineProps({
+    message: String,
+    title: String,
+    confirmText: { type: String, default: 'Confirm' },
+    cancelText: { type: String, default: 'Cancel' },
+});
 </script>
 
 <style scoped>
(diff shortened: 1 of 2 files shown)


frontend/src/components/Dialog/Dialog.vue
```diff
@@ -35,29 +35,13 @@
     </vue-final-modal>
 </template>
 
-<script>
-export default {
-    inheritAttrs: false,
-    emits: ['confirm'],
```
MCatherine1994, line 41: Hi Nick, saw it's trying to emit the "confirm" method, we might need to declare this emit in order to call this method. https://vuejs.org/guide/typescript/composition-api.html#typing-component-emits

frontend/src/components/Dialog/Dialog.vue
```diff
@@ -35,29 +35,14 @@
...
-            default: 'Confirm',
-        },
-        cancelText: {
-            type: String,
-            default: 'Cancel',
-        },
-    },
-};
+<script setup lang="ts">
+const emit = defineEmits(['confirm']);
```
MCatherine1994, line 39: Hi Nick, sorry I checked the documentation again, seems like we don't need to do the emit declaration, if we just use $emit in the html template. Sorry about that, think we could remove this. https://vuejs.org/guide/components/events.html#event-arguments

frontend/src/components/Dialog/Dialog.vue
```diff
@@ -35,29 +35,14 @@
...
-            type: String,
-            default: 'Confirm',
-        },
-        cancelText: {
-            type: String,
-            default: 'Cancel',
-        },
-    },
-};
+<script setup lang="ts">
```
MCatherine1994, line 38: We want to add `inherit-attrs="false"` here, <script setup inherit-attrs="false" lang="ts">

NickSaglioni: Thanks! I have changed it.

MCatherine1994 (approved):

frontend/src/components/Dialog/Dialog.vue
NickSaglioni: No problem, I have reverted back this change.

PR #467, opened 2023-03-03
Feat/464: Bypass Cognito Login UI

MCatherine1994:
Bypass cognito login UI

### DIFF
--- a/frontend/public/env.json
+++ b/frontend/public/env.json
@@ -29,6 +29,11 @@
         "type": "string",
         "value": "http://localhost:5173"
     },
+    "fam_console_idp_name": {
+        "sensitive": false,
+        "type": "string",
+        "value": "DEV-IDIR"
+    },
     "target_env": {
         "sensitive": false,
         "type": "string",
--- a/frontend/src/services/AuthService.ts
+++ b/frontend/src/services/AuthService.ts
@@ -34,15 +34,10 @@ async function login() {
         https://docs.amplify.aws/lib/auth/social/q/platform/js/
         https://docs.amplify.aws/lib/auth/advanced/q/platform/js/#identity-pool-federation
     */
-    // Auth.federatedSignIn();
 
     const env = JSON.parse(window.localStorage.getItem('env_data') || '');
 
-    window.location.href =
-        `https://${env?.fam_cognito_domain.value}.auth.${env?.fam_cognito_region.value}.amazoncognito.com/` +
-        `oauth2/authorize?identity_provider=${env?.target_env.value.toUpperCase()}-IDIR&` +
-        `redirect_uri=${env?.front_end_redirect_base_url.value}/authCallback&response_type=CODE&` +
-        `client_id=${env?.fam_console_web_client_id.value}&state=&scope=openid`;
+    Auth.federatedSignIn({ customProvider: env?.fam_console_idp_name.value });
 }
 
 async function logout() {
--- a/frontend/src/services/AuthService.ts
+++ b/frontend/src/services/AuthService.ts
@@ -2,6 +2,7 @@ import router from '@/router';
 import type { CognitoUserSession } from 'amazon-cognito-identity-js';
 import { Auth } from 'aws-amplify';
 import { readonly, ref } from 'vue';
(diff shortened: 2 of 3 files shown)


frontend/src/services/AuthService.ts
```diff
@@ -34,7 +34,15 @@ async function login() {
...
         https://docs.amplify.aws/lib/auth/advanced/q/platform/js/#identity-pool-federation
     */
-    Auth.federatedSignIn();
+    // Auth.federatedSignIn();
+
+    const env = JSON.parse(window.localStorage.getItem('env_data') || '');
+
+    window.location.href =
+        `https://${env?.fam_cognito_domain.value}.auth.${env?.fam_cognito_region.value}.amazoncognito.com/` +
+        `oauth2/authorize?identity_provider=${env?.target_env.value.toUpperCase()}-IDIR&` +
```
MCatherine1994, line 43: The target_env variable in deployment is actually an injected variable, in the `infrastructure/server/outputs.tf` file, it says as the following, so I assume its value will be "dev" or "test" or "prod"? If so, I just make it to upper case and the identity provider will be "DEV-IDIR" or "TEST-IDIR" or "PROD-IDIR"
```
output "target_env" {
  description = "dev, test, or prod in AWS"
  value = var.target_env
}
```

basilv: Hmm, this is only temporary at best until BCeID users are added to FAM. And this solution seems like a bit of a hack.
For example, assuming the identity provider name could be avoiding by passing in the IDIR identity provider name into the env_data.

frontend/src/services/AuthService.ts
```diff
@@ -34,7 +34,15 @@ async function login() {
         https://docs.amplify.aws/lib/auth/social/q/platform/js/
         https://docs.amplify.aws/lib/auth/advanced/q/platform/js/#identity-pool-federation
     */
-    Auth.federatedSignIn();
+    // Auth.federatedSignIn();
+
+    const env = JSON.parse(window.localStorage.getItem('env_data') || '');
```
basilv, line 39: The EnvironmentSettings class should be used instead as that abstracts accessing environment-specific data.

frontend/src/services/AuthService.ts
```diff
@@ -34,7 +34,15 @@ async function login() {
         https://docs.amplify.aws/lib/auth/social/q/platform/js/
         https://docs.amplify.aws/lib/auth/advanced/q/platform/js/#identity-pool-federation
     */
-    Auth.federatedSignIn();
```
basilv, line 37: Not using Auth.federatedSignIn seems to bypass the Amplify library entirely, at least for login. Is there no way to pass in the identityProvider option into the signin method? Documentation suggests this is possible - e.g. 
After configuring the OAuth endpoints (Cognito Hosted UI), you can integrate your App by invoking Auth.federatedSignIn() function. Passing LoginWithAmazon, Facebook, Google, or SignInWithApple on the provider argument (e.g Auth.federatedSignIn({ provider: 'LoginWithAmazon' }))

basilv (changes requested):
@MCatherine1994 Review done, changes suggested.

frontend/src/services/AuthService.ts
MCatherine1994: Updated to use customProvider option as mentioned in this example: 
https://github.com/aws-amplify/amplify-js/issues/10900

frontend/public/env.json
```diff
@@ -29,6 +29,11 @@
         "type": "string",
         "value": "http://localhost:5173"
     },
+    "fam_console_idp_name": {
+        "sensitive": false,
```
MCatherine1994, line 33: This is an existing variable in the terraform

basilv (approved):
I'm fine with this. I'm also fine if you want to try moving the customProvider setting into the AWS Config as per Ian's suggestion.

MCatherine1994:
> I'm fine with this. I'm also fine if you want to try moving the customProvider setting into the AWS Config as per Ian's suggestion.

I didn't find a solution for setup that in the Amlify config method. I checked in their GitHub repos issues, most of the people are using the customProvider in the signin method to bypass the Cognito UI.

PR #487, opened 2023-03-10
Feat/453: Onboard SPAR

MCatherine1994:
- onboard spar, create spar oidc client in terraform
- add spar application, application client for all env, spar admin role for all env, and two testing roles "user_write" and "user_read" (got from SPAR team) for all env in db
- update flyway infrastructure to pass cognito spar client id
- update pipeline and docker-compose to put flyway placeholder for the spar client id, just random strings
- update the test that gets all applications, update the asset number when we add apps

### DIFF
--- a/infrastructure/server/oidc_clients_spar.tf
+++ b/infrastructure/server/oidc_clients_spar.tf
@@ -1,0 +1,105 @@
+resource "aws_cognito_user_pool_client" "dev_spar_oidc_client" {
+  access_token_validity                         = "5"
+  allowed_oauth_flows                           = ["code"]
+  allowed_oauth_flows_user_pool_client          = "true"
+  allowed_oauth_scopes                          = ["openid", "profile", "email"]
+  callback_urls                                 = [
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/",
+    "http://localhost:3000/dashboard",
+    "http://localhost:3000/silent-check-sso"
+  ]
+  logout_urls                                   = [
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/",
+    "http://localhost:3000/"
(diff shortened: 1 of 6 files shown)


MCatherine1994:
Hi @ConradBoydElliottGustafson, the flyway test failed for auth function, I wonder if we should put some placeholder for the spar client id in both pipeline and docker compose file as for fom? are those random numbers?
image

Screen Shot 2023-03-10 at 12 43 12 PM
Screen Shot 2023-03-10 at 12 43 24 PM

ConradBoydElliottGustafson (approved):
Looks awesome! This PR will be the example for going forward with onboarding.

infrastructure/server/oidc_clients_spar.tf
```diff
@@ -0,0 +1,105 @@
...
+  allowed_oauth_flows                           = ["code"]
+  allowed_oauth_flows_user_pool_client          = "true"
+  allowed_oauth_scopes                          = ["openid", "profile", "email"]
+  callback_urls                                 = [
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/",
+    "http://localhost:3000/dashboard",
+    "http://localhost:3000/silent-check-sso"
+  ]
+  logout_urls                                   = [
+    "https://oidcdebuggersecure-3d5c3f-dev.apps.silver.devops.gov.bc.ca/",
```
MCatherine1994, line 12: Don't forget to add the logout chain!! Fix this in another pull request https://github.com/bcgov/nr-forests-access-management/pull/495

MCatherine1994:
For the sql script to do insert, better to add ("on conflict (application_name, etc...") do nothing) for later easier fixing migration

MCatherine1994:
When create the new version of the flyway script, be careful about what's the current version is, and if main branch has another version, be careful!!

Issue #336, opened 2022-12-13
Filter FAM Front-end by Environment 


ArogeG:
**As a** *FAM user with access to multiple FAM environments **I want** to view FAM applications by environment **So That** *I can easily provision access to the correct environment*

**Additional Context**
-To support the decision to create multiple DEV, TEST, and PROD IDP configuration in the Prod environment , the front-end needs to be designed to allow users select the environment they intend to work on.
- Create three single select buttons for Prod, Test, and Dev
- Filter the list of application based on the button selected. 

**Acceptance Criteria**
- [ ] Given that the single select buttons are in place, when a user select an environment, then the application list should be filtered to display only applications in the selected env that the user has access to administer. 

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

![image.png](https://images.zenhubusercontent.com/62a1029ce9497c39c087ab8b/06460c3d-f3ea-4f61-aeba-1f9674a9bc89)

NickSaglioni:
A conversation was held on 03/03 with myself and @OlgaLiber2 about this implementation and two possibilities were discussed. The first one was to implement this task as it is described in this ticket. The other possible implementation would be to have a multiple selection for cases were a specific user would need to be granted access to multiple environments at the same time like, for instance, PROD and TEST. This second implementation would need a major application refactoring as the grant access screen and backend API currently only expect a single application at a time. Olga also raised the possibility to change the current app workflow for granting access, but that needs to be verified and validated first. For now, the implementation described in the task was chosen. @ArogeG

NickSaglioni:
@MCatherine1994 suggested to use the bootstrap package for vue3 for this implementation to create a common radio button component.

https://bootstrap-vue.github.io/bootstrap-vue-next/getting-started/#installation-vue-js

PR #488, opened 2023-03-13
feat: #336 filter fam front-end by environment

NickSaglioni:
Adding bootstrap-vue-next package and creating new component radio group


frontend/src/components/SelectApplication.vue
```diff
@@ -26,45 +37,49 @@ onMounted(async () => {
         return Promise.reject(error);
     }
 });
+
+function valueChanged(event: any) {
```
MCatherine1994, line 41: The parameter of valueChanged function should be the selected environment, which is a string. And maybe we should give it a better name

function filterEnv(selectedEnv: string) {
}

frontend/src/components/common/RadioGroup/RadioGroup.vue
```diff
@@ -0,0 +1,40 @@
...
+        default: [],
+    },
+    initialValue: {
+        type: String,
+        required: true,
+        default: '',
+    },
+});
+const emit = defineEmits(['change']);
+const valueFilter = ref<string>('');
```
MCatherine1994, line 28: let's rename this to "selectedValue"

frontend/src/components/common/RadioGroup/RadioGroup.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+    <div class="row">
+        <b-form-radio-group
+            id="radio-group-1"
```
MCatherine1994, line 4: I wonder if we need "id" and "name" here, do they need to be unique?

frontend/src/services/ApplicationState.ts
```diff
@@ -5,7 +5,7 @@ import { computed, ref } from 'vue';
 export const applicationsUserAdministers = ref<FamApplication[]>([]);
 
 // The application selected by the user to admin
-export const selectedApplication = ref<FamApplication>();
+export const selectedApplication = ref<FamApplication | null>();
```
MCatherine1994, line 8: let's change "null" to "undefined", so it won't complain the type error when we use it

MCatherine1994:
maybe we don't need to put RadioGroup into a folder? it can stay in the common folder directly?

frontend/src/services/ApplicationState.ts
NickSaglioni: Usually undefined is when a variable is declared but a value has not been assigned. Do you prefer to use it instead of null for this case?

frontend/src/components/SelectApplication.vue
```diff
@@ -26,45 +37,49 @@ onMounted(async () => {
...
+    environmentFilter.value = event;
+    selectedApplication.value = null;
```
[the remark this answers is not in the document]
NickSaglioni: Same comment as above.

frontend/src/components/common/RadioGroup/RadioGroup.vue
```diff
@@ -0,0 +1,40 @@
...
+            name="radio-options"
+            @change="valueChanged"
+        ></b-form-radio-group>
+    </div>
+</template>
+
+<script setup inherit-attrs="false" lang="ts">
+import { ref, onBeforeMount, type PropType } from 'vue';
+const props = defineProps({
+    values: {
```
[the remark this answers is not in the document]
NickSaglioni: Ok, I will change this.

frontend/src/components/common/RadioGroup/RadioGroup.vue
NickSaglioni: Ok, I will change this.

NickSaglioni:
> maybe we don't need to put RadioGroup into a folder? it can stay in the common folder directly?

Sure, no problem. It is just a personal organization taste. I will change this.

frontend/src/services/ApplicationState.ts
MCatherine1994: yeah... maybe we don't need to change this, I saw a type complain before, but I just checked again, it disappeared... maybe I just need a refresh

frontend/src/components/common/RadioGroup/RadioGroup.vue
NickSaglioni: They are not needed, removing those. Thanks for spotting it.

frontend/src/components/SelectApplication.vue
```diff
@@ -26,45 +37,49 @@ onMounted(async () => {
         return Promise.reject(error);
     }
 });
+
+function filterEnv(event: string) {
```
MCatherine1994, line 41: Hi Nick, shall we not call it event? even sounds like an event, not a string :  )) thanks!!

NickSaglioni: Thanks. This has been changed.

frontend/tsconfig.json
```diff
@@ -1,7 +1,14 @@
 {
     "extends": "@vue/tsconfig/tsconfig.web.json",
-    "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
+    "include": [
+        "env.d.ts",
+        "src/**/*",
+        "src/**/*.vue",
+        "tests/**/*.ts",
+        "components.d.ts"
```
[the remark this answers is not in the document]
NickSaglioni: Thank you for spotting it. "components.d.ts" is needed for the bootstrap-vue-next package. I have removed tests because it was an unwanted change.

frontend/src/components/SelectApplication.vue
```diff
@@ -6,9 +6,20 @@ import {
     isApplicationSelected,
     selectedApplication,
 } from '@/services/ApplicationState';
-import { onMounted } from 'vue';
+import { onMounted, ref } from 'vue';
 import router from '../router';
 
+const ALL_APPS = 'All';
```
ianliuwk1019, line 12: Hi Nick, `const environments[0]` is already having value `All`.
Probably no need to define this ALL_APPS const anymore?

ianliuwk1019 (changes requested):

MCatherine1994 (approved):

frontend/src/components/SelectApplication.vue
```diff
@@ -26,45 +37,49 @@ onMounted(async () => {
...
+                            v-for="app in applicationsUserAdministers.filter(
+                                (item) =>
+                                    item.app_environment?.includes(
+                                        environmentFilter.toUpperCase()
+                                    )
+                            )"
+                            :value="app"
+                        >
+                            {{ app.application_description }}
+                        </option>
```
ianliuwk1019, line 80: Hi Nick,
I feel putting this javascript filtering code in the <script></script> area with a function might be cleaner, not at the html <template>.
Also you probably don't need "v-if" and "v-else" condition. It can be done in one function filtering. Just move your `environmentFilter == ALL_APPS` inside filter to return 'true' for all element no matter what it is.

frontend/src/main.ts
```diff
@@ -16,6 +16,9 @@ import router from '@/router';
 import 'bootstrap';
 import './assets/styles/styles.scss';
 
+import 'bootstrap/dist/css/bootstrap.css'
```
ianliuwk1019, line 19: Do we need to import this bootstrap.css again? FAM already uses bootstraps style. I think it is imported in styles.scss file already.
Could you check?

frontend/src/components/common/RadioGroup.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+    <div class="row">
+        <b-form-radio-group
+            v-model="selectedValue"
+            :options="props.options"
+            @change="valueChanged"
+        ></b-form-radio-group>
+    </div>
+</template>
```
ianliuwk1019, line 9: Hi Nick,
I am not sure it is necessary we need to create a wrapper around the library. 
Isn't it easier to just use original <b-form-radio-group> with the doc instruction provided?
I don't see we are styling it differently either.
The doc also can have <b-form-radio-group> displayed in a row just fine.
Or maybe I am missing some points here?

ianliuwk1019 (changes requested):

frontend/src/main.ts
NickSaglioni: Thanks, yes, it is imported already. Removing this.

frontend/src/components/common/RadioGroup.vue
NickSaglioni: The point is to make it easier to style using the carbon template in the future, but if we agree that it is not necessary, I can remove it.

frontend/src/components/SelectApplication.vue
NickSaglioni: This const is to make it clear that is for the all_apps option, not array item 0 which might be confusing, specially if we change the app order in the future.

frontend/src/components/common/RadioGroup.vue
MCatherine1994: Hi Ian, sorry I didn't get this, what means a wrapper around the library? like the line <div class="row"> ?

frontend/src/components/common/RadioGroup.vue
```diff
@@ -0,0 +1,38 @@
...
+
+<script setup inherit-attrs="false" lang="ts">
```
MCatherine1994, line 11: maybe we don't need the "inherit-attrs" here

frontend/src/components/SelectApplication.vue
NickSaglioni: I have changed this as suggested.

frontend/src/components/common/RadioGroup.vue
NickSaglioni: It was removed

MCatherine1994 (approved):

PR #494, opened 2023-03-14
438 tag based deploy

ConradBoydElliottGustafson:


server/backend/pytest.ini
```diff
@@ -15,6 +15,6 @@ markers =
 
 env =
     COGNITO_REGION=ca-central-1
-    COGNITO_USER_POOL_ID=ca-central-1_yds9Vci8g
-    COGNITO_CLIENT_ID=75g60i0g3rt0npuvvho1ucvjm8
+    COGNITO_USER_POOL_ID=ca-central-1_SMbUfdCC3
+    COGNITO_CLIENT_ID=7oi2u0st9ov66aahdk02at9f7g
```
MCatherine1994, line 19: just put a comment here, looks like the cognito credentials are only needed for sqlite testing, we could consider remove them once we rewrite all the tests using test container

MCatherine1994 (approved):

.github/workflows/reusable_terraform_server.yml
```diff
@@ -111,22 +115,15 @@ jobs:
...
-          if ((echo ${{ github.ref_name }} | wc -c ) > 12)
-          then
-              commit=${{ github.ref_name }}
-              db_cluster_snapshot_identifier="pre-flyway-${{ github.ref_name }}-${commit:0:7}"
-          fi
-          # remove non alphanumeric characters
+          # is less than 63 characters and has no non alphanumeric characters
+          refname=${{ github.ref_name }}
+          commit=${{ github.sha }}
+          db_cluster_snapshot_identifier="pre-flyway-${refname:0:12}-${commit:0:7}"
```
ianliuwk1019, line 121: Thanks Conrad for fixing that "commit" bug previously.
However, is there a reason to provide the upper bound "12" or should it be based on "length"?
If there is a branch name less than 12, then I wonder if it will have index-out-of-bound error (not sure in terraform)?
Thanks.

ConradBoydElliottGustafson: You're right, and I didn't test that condition. I'm not really sure what that GHA syntax does. Maybe we should merge this and then open that as a small defect ticket if necessary?

ianliuwk1019: > You're right, and I didn't test that condition. I'm not really sure what that GHA syntax does. Maybe we should merge this and then open that as a small defect ticket if necessary?

That's good idea, thanks Conrad!

PR #495, opened 2023-03-15
Feat/453 onboard spar add logout chain

MCatherine1994:


ConradBoydElliottGustafson (approved):
Wow, this is a tricky one for sure. Looks like you nailed it.

MCatherine1994:
> Wow, this is a tricky one for sure. Looks like you nailed it.

I totally forgot about it last time, I'll put a comment in the last pull request when we onboard SPAR, so I won't forget next time  :: ))

Issue #516, opened 2023-03-20
Duplicate and Environment Conflict when Creating Role Assignment in PROD for FOM TEST for Client 


ArogeG:
**Describe the bug**
When a role assignment record exists in an application's environment (DEV), the same role assignment cannot be created in the another environment(TEST). The system returns a duplicate record notification. 

When the record is deleted in one environment(DEV) and re-created in a different environment(TEST), the record is displayed in the previous environment (DEV)

**To Reproduce**
1. Go to 'FAM Dev'
2. Click on 'FOM DEV' application on the Select Application screen
3. Create a role assignment with a client # for FOM
4. Save the record
5. Navigate back to Select application on FAM
6. Click on 'FOM TEST' application 
7. Create the same Role assignment with the same client #

The screen returns an error message: _Duplicate record cannot be created_

8. Delete the same role assignment from FOM Dev application on FAM
9. Recreate the Record for FOM TEST application, FAM notifies that record has been successfully created

**Record is not displayed on the Manage Access screen**

10. Navigate back to the FOM Dev application on FAM Dev

**Record can be seen on the Manage Access Screen.**

**Expected behavior**
**user assignment should be unique across multiple environments**
**Recreated records should be created for the environment in which the record is being created **

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

![image.png](https://images.zenhubusercontent.com/62a1029ce9497c39c087ab8b/1617614c-0837-4c1e-941d-d064b90c9825)

![image.png](https://images.zenhubusercontent.com/62a1029ce9497c39c087ab8b/6764a709-b2c1-43f9-80c9-3e6153efe26b)

basilv:
Nice find. We just need to fix logic (either db constraint or business rule) to check environment as well. But I'm confused why that wasn't already done - the application should be part of what's checked for duplicate, and the application is scoped to environment. Unless we are missing application from the check?

ianliuwk1019:
Backend code fix and new tests are added into PR #520.

ArogeG:
Thanks @ianliuwk1019 is this deployed to Dev yet?

ianliuwk1019:
Yes it is currently deployed to dev as the pr is merged and closed, but I moved this ticket back for your review @ArogeG . Also FOM's thickline is fixed as well.
(logout - need to do it in FAM) is still in progress. So FOM test still has problem for logout.

ArogeG:
[...]

PR #520, opened 2023-03-21
fix: #516 Duplicate and Environment Conflict when Creating Role Assignment in PROD for FOM TEST for Client

ianliuwk1019:
This PR is to fix bug described in #516.
* The bug is identified as a wrong query result on crud_role.get_role_by_role_name(), which is missing one more filtering parameter (application_id) as now FAM role is environment(application) specific (dev/test/prod).
* Previous FAM fam_role table had initial unique constraint as (role_name) only. Then some time ago, this constraint is updated to (role_name, application_id) after FAM has multiple application environments(such as fom_dev/fom_test/fom_prod) for each FAM environment and the function was missed to upgrade along that change.

### DIFF
--- a/server/backend/api/app/constants.py
+++ b/server/backend/api/app/constants.py
@@ -22,5 +22,11 @@ FAM_PROXY_API_USER = "fam_proxy_api"
 
 COGNITO_USERNAME_KEY = "username"
 
+FOM_SUBMITTER_ROLE_NAME = "FOM_Submitter"
+FOM_APP_DESC = "Forest Operations Map"
+FOM_ROLE_PURPOSE = "Grant a user access to submit to FOM"
+FOM_APP_DEV_NAME = "fom_dev"
+FOM_APP_TEST_NAME = "fom_test"
+
 # TODO: Only uses this before forest-client api integration from front-end.
 DUMMY_FOREST_CLIENT_NAME = "DUMMY_FOREST_CLIENT_NAME"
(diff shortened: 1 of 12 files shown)


basilv (dismissed):

server/backend/api/app/constants.py
```diff
@@ -22,5 +22,11 @@ class AppEnv(str, Enum):
 
 COGNITO_USERNAME_KEY = "username"
 
+FOM_SUBMITTER_ROLE_NAME = "FOM_Submitter"
+FOM_APP_DESC = "Forest Operations Map"
+FOM_ROLE_PURPOSE = "Grant a user access to submit to FOM"
+FOM_APP_DEV_NAME = "fom_dev"
+FOM_APP_TEST_NAME = "fom_test"
```
basilv, line 29: Why do we have app dev and app test names but no prod?

server/backend/api/app/constants.py
```diff
@@ -22,5 +22,11 @@ class AppEnv(str, Enum):
...
+
 # TODO: Only uses this before forest-client api integration from front-end.
 DUMMY_FOREST_CLIENT_NAME = "DUMMY_FOREST_CLIENT_NAME"
```
basilv, line 32: I believe this can be removed?

server/backend/api/app/constants.py
ianliuwk1019: I moved tests only constants out from backend constants file to avoid confusion.
Thanks for the catch!

server/backend/api/app/constants.py
ianliuwk1019: I moved tests only constants out from backend constants file to avoid confusion.
Thanks for the catch!

basilv (approved):

Issue #519, opened 2023-03-21
FAM Cognito logout setting for FOM (TEST) does not work.


ianliuwk1019:
While testing on FOM on (TEST )environment, the logout has an error popup.
Suspect Cognito setting for logout url might be wrong (or if correct, then need to check FOM openshift deployment template for this settings.)

Note: FOM (TEST) environment is configured to use FAM (PROD) Cognito with `**fom_test**` environment.
Also, double check Terraform Cognito logout url setting for PROD as well.

ianliuwk1019:
Interesting.....

Both (FOM Test, FAM PROD - fom_test) configurations are there, but it is a mismatch. The Cognito logout url fix was deployed last week.
On FOM Test - it is:
```
https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://test.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=https://fom-test.nrs.gov.bc.ca/admin/not-authorized?loggedout=true
```

But on FAM PROD (fom_test), it is:
```
https://logon.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=https://fom-test.nrs.gov.bc.ca/admin/not-authorized?loggedout=true
```

Checking on Terraform/Terragrunt configuration the settings seem to be fine (but looks to me on Cognito, **fom_test** is using terragrunt proudction setting):
This is terragrunt `test` setting:
```
  fam_logout_urls = [
    "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://test.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=https://fam-tst.nrs.gov.bc.ca",
    "https://logontest7.gov.bc.ca/clp-cgi/logoff.cgi?retnow=1&returl=https://test.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/logout?redirect_uri=http://localhost:5173"
  ]
```

This is terragrunt `prod` setting:
```
  fam_logout_urls = [
[...]

PR #521, opened 2023-03-23
fix: #519 fam cognito logout for fom TEST not working

ianliuwk1019:
The PR fixes logout_chain_url not matching with Cognito app clients environments ([app]_dev, [app]_test, [app]_prod)
* Migrate common logout chain urls (Cognito app client env specific) to common_vars.hcl.
* Setup terragrun "cognito_app_client_logout_chain_url" var as a map so oidc app client can reference to it with environments(dev, test, prod)
* Use "local.common_vars.inputs.idp_logout_chain_dev_url" to make values DRY on each environment's setting.
* Make sure oidc_client_fom.tf/oidc_client_spar.tf use environment specific value for dev/test/prod


basilv (approved):
This looks good to me, I like the common constants reducing duplication.

Issue #305, opened 2022-11-24
Back-end Integrate with Forest Client API  - Phase 1


gormless87:
As a an application access administrator I want to see a list of forest client So That I can select a forest client for an assignment.

Additional Context
-- Integrate Forest Client API with frontend and back-end for validation.
--Allow users to select and search for a forest client by forest organization name/client  and ID.
--If Forest Client does not exists, return a message "Forest Client cannot be Found".
--Security - must be logged FAM user in order for FC API to work

Acceptance Criteria

From the front end, can execute a search and get results returned from the Forest Client API

Definition of Done
Ready to Demo in Sprint Review
Does what I have made have appropriate test coverage?
Documentation and/or scientific documentation exists and can be found
Peer Reviewed by 2 people on the team
Manual testing of all PRs in Dev and Prod
Merged

Cloned from #179

gormless87:
Need to connect with Team Alliance on their Forest Client API

franTarkenton:
thinking this could be further broken down into two tickets
* wrap forest client api into our api
    * to support search of forest client based on forest client name or forest client number
* a ticket that would address, how and when we would integrate forest clients into our database.
    * thinking this would not take place during a search, but rather when we we add a new user/role assignment takes place

ianliuwk1019:
Got some information from Maria today:
- Direct link to Forest Client api-docs: https://nrfc-api-prod-backend.apps.silver.devops.gov.bc.ca/
- Need to request access through API Services Portal: https://api.gov.bc.ca/devportal/api-directory
- Currently the **Forest Client Api has not been published** yet @ API Services Portal due to some work on database vault.

Q: how to request access through API Services Portal for Forest Client Api if it is available?

ianliuwk1019:
Got some information from Maria today:
- Direct link to Forest Client api-docs: https://nrfc-api-prod-backend.apps.silver.devops.gov.bc.ca/
- Need to request access through API Services Portal: https://api.gov.bc.ca/devportal/api-directory
- Currently the **Forest Client Api has not been published** yet @ API Services Portal due to some work on database vault.

Q: how to request access through API Services Portal for Forest Client Api if it is available?

ianliuwk1019:
It may not be available next week.
Reply from Alliance Team:

> Hi Ian,
> 
[...]

PR #532, opened 2023-03-29
feat: #305 forest client api backend integration

ianliuwk1019:
Details for #305 PR:
* Add a class responsible for interfacing (make api calls to) Forest Client API.
* Use API Token from environment variable for setting up the request header.
* Add new FAM Forest Client - search endpoint by query parameter (client_number).
  - guarded with authorized user login.
  - for now restrict to exact 8-digits length (limitation from Forest Client API)
* Parse API return result for FAM search endpoint needs: "client_name", "forest_client_number", "status: code/description".
* Create custom crud business validator to validate creat user/role assignment before saving:
  - validate if the forest client exists.
  - validate if the forest client is active (ACT).
* Generate API Client for frontend use.
* Add new tests.
* API Token Setup for all environments. (Keys are set at repo first as suggested by Conrad)
* Adjust CI workflow so it can read GH environment secret token for Unit Testing.


server/backend/api/app/crud/crud_user_role.py
```diff
@@ -215,3 +229,42 @@ def find_or_create_forest_client_child_role(
...
+            fc_api = ForestClient()
+
+            # Locally stored (if any) for later use to prevent api calls again.
+            # Exact client number search - should only contain 1 result.
+            self.fc = fc_api.find_by_client_number(forest_client_number)
+            LOGGER.debug(f"Forest Client(s) retrieved: {self.fc}")
+
+    def forest_client_number_exists(self) -> bool:
+        # Exact client number search - should only contain 1 result.
+        return len(self.fc) == 1
```
basilv, line 259: So multiple results will cause this to fail, I'm thinking I'm okay with this.

basilv (changes requested):

server/backend/api/app/crud/crud_forest_client.py
```diff
@@ -1,6 +1,8 @@
+import json
```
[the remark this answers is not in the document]
ianliuwk1019: Good point, thanks Basil!
I guess I have 'crud' already baked in to hobby but in this case, I guess I can just moved extra code into router itself.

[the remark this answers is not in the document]
gormless87: Leftover from some Guy named Fran?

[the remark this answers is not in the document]
ianliuwk1019: Updated.
Hahah no Pete, it's leftover from the Empire master 'Fran's' apprentice. From now on, I should look at your Python coding style guide instead :P

client-code-gen/gen/model/fam-forest-client-status.ts
```diff
@@ -0,0 +1,39 @@
...
+     * @type {FamForestClientStatusType}
+     * @memberof FamForestClientStatus
+     */
+    'status_code': FamForestClientStatusType;
+    /**
+     * 
+     * @type {string}
+     * @memberof FamForestClientStatus
+     */
+    'description': string;
```
MCatherine1994, line 37: just wondering shall we give the description a real description? or it doesn't matter?

basilv (changes requested):
Didn't quite finish reviewing all the code but thought I should send over my review comments.

server/backend/api/app/routers/router_forest_client.py
```diff
@@ -0,0 +1,44 @@
...
+from fastapi import APIRouter, Query
+
+from .. import schemas
+
+LOGGER = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.get("/search", response_model=List[schemas.FamForestClient])
```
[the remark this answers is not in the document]
ianliuwk1019: Unlike other router that has the authentication check done at "router" for individual endpoint (and can be improved), this is done by include this at higher up at "main.py" for the whole "router" control with **FastAPI** "**dependencies**"
```
app.include_router(router_forest_client.router,
                   prefix=apiPrefix + '/forest_clients',
                   dependencies=[Depends(jwt_validation.authorize)],
                   tags=["FAM Forest Clients"])
```

server/backend/api/app/integration/forest_client/forest_client.py
```diff
@@ -0,0 +1,48 @@
+import logging
+from http import HTTPStatus
+
+import requests
+from api.config import config
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ForestClient():
```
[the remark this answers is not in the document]
ianliuwk1019: Thanks Basil. I don't know either from Python world... but seems to me Python does not borrow "service" that much from various tutorials or documentations.... However, I will rename this, I do like the sound of "ForestClientService".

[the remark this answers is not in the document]
ianliuwk1019: Renamed.

server/backend/api/app/integration/forest_client/forest_client.py
```diff
@@ -0,0 +1,48 @@
...
+    def find_by_client_number(self, p_client_number: str):
+        """
+        Find Forest Client(s) information based on p_client_number search query field.
+
+        :param p_client_number: Forest Client Number string (8 digits).
+                                Note! Current Forest Client API can only do exact match.
+                                '/api/clients/findByClientNumber/{clientNumber}'
+        :return: Search result (JSON) for a Forest Client information object.
+        """
+        url = f"{self.api_clients_url}/findByClientNumber/{p_client_number}"
```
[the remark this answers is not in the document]
ianliuwk1019: In this case I think it is not necessary. Our "router endpoint" for this one does have Pydantic validation built-in (variable Typing and length checks) to request object. And, FAM in this case only concerns the 'search' result from Forest Client API (status ok - with result found or not).

server/backend/api/config/config.py
```diff
@@ -143,3 +144,16 @@ def get_allow_origins():
     allow_origins = [get_env_var("ALLOW_ORIGIN")] if is_on_aws() else ["*"]
     LOGGER.info(f"allow_origins -- {allow_origins}")
     return allow_origins
+
+
+def get_forest_client_api_token():
+    api_token = get_env_var("FC_API_TOKEN") if is_on_aws() \
+        else "4r0dn5wC0JTs6N1tdw6NDzpjBRQwEJ6u"  # Test env.
```
basilv, line 151: Probably bad security - this token is now exposed to the entire internet.

server/backend/api/app/integration/forest_client/forest_client.py
```diff
@@ -0,0 +1,55 @@
...
+                                '/api/clients/findByClientNumber/{clientNumber}'
+        :return: Search result as List for a Forest Client information object.
+                 Current Forest Client API returns exact one result or http status
+                 other than 200 with message content. The intent for FAM search is for
+                 wild card search and Forest Client API could be capable of doing that
+                 in next version.
+        """
+        url = f"{self.api_clients_url}/findByClientNumber/{p_client_number}"
+        LOGGER.debug(f"ForestClientService find_by_client_number() - url: {url}")
+        r = self.session.get(url)
```
basilv, line 44: Two error scenarios:
1. HTTP get fails entirely (no service running) - likely a python error like TCP connection timeout
2. The call returns HTTP status code 5xx indicating error in forest client.

basilv (changes requested):

server/backend/api/config/config.py
ianliuwk1019: Initially I treated this token in TEST environment isn't needed to be 'secure'.
You finally caught me, security officer!!

server/backend/api/app/integration/forest_client/forest_client.py
ianliuwk1019: Thanks Basil! As discussed, will try to see what happen in python and deal with 5xx error.

server/backend/api/config/config.py
ianliuwk1019: I removed the default TEST token now, officer... I will only share it with my team privately (if they give me a chocolate, I like chocolate)

server/backend/api/app/integration/forest_client/forest_client.py
ianliuwk1019: Python doc says: ```In the event of a network problem (e.g. DNS failure, refused connection, etc), Requests will raise a [ConnectionError](https://requests.readthedocs.io/en/latest/api/#requests.ConnectionError) exception``` so same as normal behavior, framework will handle that.
However, python best practice is we need to set "timeout" limit as there is no default one from python.

ianliuwk1019: Error handling for 400/500 and network error were addressed and discussed yesterday.
A note from handling 404 from ForestClient API as it uses this code to indicate return of "Client Not Found" so we handle this case in try-catch block and return empty for our search purpose.

server/backend/api/config/config.py
ianliuwk1019: Token removed.
Encountered using token from GH report secret for the CI run. Regular terraform plan seems to be fine.
Resolved from CI yaml template.

basilv (approved):

Issue #469, opened 2023-03-06
Convert existing tests to work against postgres database


MCatherine1994:
Copy and Modify the sqlite tests to work with postgres, and create more if needed to increase the code coverage. 

Suggestions:
- test all crud functions (could discuss if this is needed)
- test all routers

Remove the tests folder once finished. The tests folder holds all sqlite tests. Remember to move the `server/backend/tests/tests/jwt_utils.py` file to the new testspg folder, as that one is reusable

Task:
- Write tests
- Watch the running time for running all tests
- Check if could start the docker container just with postgres and flyway, skip the api service (maybe using Docker capability for compose file with 'extend', not completely unrelated two separate compose files)

MCatherine1994:
- Running time for the postgres testing still remains within 2 mins, didn't increase a lot when we added more tests.
- The test container doesn't support start the docker container with selected service, like just with postgres and flyway, skip the api service. I created a ticket in the test container repo testcontainers/testcontainers-python#315

PR #535, opened 2023-03-29
test: #469 add postgres tests

MCatherine1994:
Write postgres tests based on the sqlite tests example

Found a few crud methods are not in use at all:
- get_applications
- create_application
- delete_application
- createFamGroup
- get_roles
- get_users

And some crud methods only use for verifying test results:
- get_application_by_name
- get_application_id_from_name
- get_user

Router tests: integration testing
Crud method tests: unit testing, write test for each method that is not called directly by the router


server/backend/api/app/crud/crud_application.py
```diff
@@ -189,6 +189,9 @@ def get_application_role_assignments(
 
 
 def get_application_id_by_role_id(db: Session, role_id):
+    # question: why we need to go to fam_application table?
+    # the fam_role table has the application_id already
+
     application_id =\
         db.query(models.FamApplication.application_id)\
```
MCatherine1994, line 196: Got a question here, why we need to go to fam_application table? The fam_role table has the application_id already

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -157,6 +157,7 @@ def construct_forest_client_role_purpose(
     parent_role_purpose: str, forest_client_number: str
 ):
     LOGGER.debug(f"parent_role_purpose: {parent_role_purpose}")
+    # ???? is there a reason we put a ")" in the client purpose?
     client_purpose = f"{parent_role_purpose} for {forest_client_number})"
```
MCatherine1994, line 161: Wonder if there is a reason that we put a ")" in the client purpose? or just a typo?

server/backend/tests/tests/crud/test_crud_utils.py
```diff
@@ -42,35 +42,35 @@ def test_get_next(dbsession_fam_users, userdata2_pydantic, delete_all_users):
...
+#     ),
+#     (None, "something", "some_other_thing", None)
+# ])
+# def test_replace_str_list(str_list_to_test, str_to_replace, replace_with, expcted_str_list):
+#     result = crud_utils.replace_str_list(str_list_to_test, str_to_replace, replace_with)
+#     if result:
+#         for index, r in enumerate(result):
+#             assert r == expcted_str_list[index]
+#     else:
+#         assert result == expcted_str_list
```
MCatherine1994, line 76: These tests are the same for both postgres and sqlite tests, comment out in the sqlite tests to reduce duplicate.

server/backend/testspg/router/test_router_application.py
```diff
@@ -1,21 +1,213 @@
...
+    role_end_point = endPoint + f"/{TEST_FOM_DEV_APPLICATION_ID}/fam_roles"
+    token = jwt_utils.create_jwt_token(test_rsa_key, access_roles_fom_dev_only)
+    response = test_client_fixture.get(role_end_point, headers=jwt_utils.headers(token))
+    data = response.json()
+    # only return roles without a parent role
+    assert len(data) == 2
+
+    for app_role in data:
+        # return example: how we define what to return?
+        # the model has more columns
```
MCatherine1994, line 90: Have a questions how we define what to return in the router method?

server/backend/api/app/crud/crud_user_role.py
MCatherine1994: Removed that ")", looks like a typo

MCatherine1994:
Another question, not very related to the postgres tests, just found that, I wonder if we want to rename this api? Rename it from "/fam_applications/{application_id}/fam_roles" to "/fam_applications/{application_id}/roles", cause this api returns the application roles, not fam roles

image

server/backend/testspg/crud/test_crud_application.py
```diff
@@ -1,34 +1,167 @@
...
+    "update_user": "Tester",
+    "update_date": datetime.datetime.now(),
+}
 
-def test_get_applications(dbPgSession: sessionmaker):
-    apps = crud_application.get_applications(db=dbPgSession)
+
+def test_get_applications(db_pg_connection: Session):
+    apps = crud_application.get_applications(db=db_pg_connection)
     assert len(apps) == 7
```
ianliuwk1019, line 35: I wonder if this "7" is too magical to put into test... and will fail each time onboarding new app.

MCatherine1994: Yes, maybe I should change it to "> 1"?

MCatherine1994: Updated to > 1

server/backend/testspg/router/test_router_user_role_assignment.py
```diff
@@ -0,0 +1,418 @@
...
+    TEST_FOM_TEST_APPLICATION_ID
+
+LOGGER = logging.getLogger(__name__)
+endPoint = f"{apiPrefix}/user_role_assignment"
+
+FOM_DEV_ADMIN_ROLE = "FOM_DEV_ACCESS_ADMIN"
+FOM_TEST_ADMIN_ROLE = "FOM_TEST_ACCESS_ADMIN"
+ERROR_DUPLICATE_USER_ROLE = "Role already assigned to user."
+SUCCESS_RESPONSE = 200
+SUCCESS_RESPONSE_DELETE = 204
```
ianliuwk1019, line 25: Not sure if these http status codes are from previous written tests, we could use Python's http module `from http import HTTPStatus` for 200 and 204.

server/backend/testspg/crud/test_crud_application.py
ianliuwk1019: > I think it should be `assert len(apps) == 42` because 42 is the answer to all questions! https://en.wikipedia.org/wiki/42_(number)#:~:text=The%20number%2042%20is%2C%20in,period%20of%207.5%20million%20years.

Ah~ Nice to see you master Fran; didn't know you are also a mathmatician! Thanks for the number~ 42 will be forever in my lotto ticket from now on!

server/backend/testspg/crud/test_crud_role.py
```diff
@@ -0,0 +1,218 @@
...
+                    "application_id": TEST_FOM_DEV_APPLICATION_ID,
+                    "role_name": TEST_NEW_ROLE,
+                    "role_purpose": TEST_ROLE_PURPOSE,
+                    "create_user": "ANOTHER_CREATOR",
+                    "role_type_code": constants.RoleType.ROLE_TYPE_CONCRETE,
+                }
+            ),
+            db_pg_connection,
+        )
+    assert str(e.value).find("duplicate key value violates unique constraint") != -1
```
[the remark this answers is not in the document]
MCatherine1994: Hi Ian, yes this one is relied on the previous one "test_create_role". We don't have a delete role method, so for role tests I didn't write any cleanup.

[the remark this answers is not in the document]
ianliuwk1019: Hi Catherine, 
Previously in old "**tests**" there were some **fixtures** setup that **cleanup** test data after each function.
That might did the job previously ("tests" folder), that's one idea we can borrow; or I think in this case,
although there is no "**delete role**" written at 'crud' or 'router', we can still use **db session** with 
direct sql delete as well.

However, maybe there are better way. I am thinking we might need to change our foundation "**pgtests**" setup to **rollback** after each test (**NOT** destroy/recreate db) so we can write less fixture to cleanup. 

Also, if I checkout this pr branch and run in terminal for just one test file (under "testpg" folder) I should
be able to run tests individually and successful. But using these tow commands and compare you will see the failure:
"`pytest --ignore tests testspg/crud/test_crud_application.py`" => passed
"`pytest --ignore tests testspg/router/test_router_application.py`" => failed.

server/backend/testspg/crud/test_crud_application.py
```diff
@@ -1,34 +1,167 @@
...
+    )
+    assert app_by_id.application_id == TEST_APPLICATION_ID_FAM
+    assert app_by_id.application_name == TEST_APPLICATION_NAME_FAM
 
-def test_get_application_by_name(dbPgSession: sessionmaker):
-    apps = crud_application.get_applications(db=dbPgSession)
+    app_by_name = crud_application.get_application(
+        db=db_pg_connection, application_id=TEST_NOT_EXIST_APPLICATION_ID
+    )
+    assert app_by_name is None
```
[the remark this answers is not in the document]
MCatherine1994: yes, this might because I copy and paste a bit, updated

server/backend/testspg/router/test_router_user_role_assignment.py
MCatherine1994: yes!! updated

server/backend/testspg/crud/test_crud_application.py
```diff
@@ -1,34 +1,167 @@
...
-    assert apps[0].application_name == "FAM"
+    for app in apps:
+        assert hasattr(app, "application_id")
+        assert hasattr(app, "application_name")
+        assert hasattr(app, "application_description")
+        assert hasattr(app, "create_user")
+        assert hasattr(app, "create_date")
+        assert hasattr(app, "update_user")
+        assert hasattr(app, "update_date")
+        assert hasattr(app, "app_environment")
```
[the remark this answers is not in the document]
MCatherine1994: yeah....I was thinking to check the return result match the table schema, I'm fine with leave it or remove it

server/backend/testspg/crud/test_crud_role.py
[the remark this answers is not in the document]
MCatherine1994: I added the cleanup for every tests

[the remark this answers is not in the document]
MCatherine1994: For the reason why run this test individually it failed "pytest --ignore tests testspg/router/test_router_application.py" => failed, it's because for router tests, we didn't call the method to setup the pg container directly. When we run all the tests together, the pg container will be called by other tests, and will be setup ready. But when run the router tests individually, we missed that part. 

I added the "db_pg_container" in the router tests where we missed that.  In fact, we only need to add it to the first test method in each router test file, it will only setup once, but to reduce confusion, I added to all, so it looks like:
Screen Shot 2023-04-11 at 4 12 00 PM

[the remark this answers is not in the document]
ianliuwk1019: > I added the cleanup for every tests

Thank you Catherine!
Although I was thinking something else, but what you did is good enough I think. Thanks!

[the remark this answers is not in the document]
ianliuwk1019: > For the reason why run this test individually it failed "`pytest --ignore tests testspg/router/test_router_application.py`" => failed it's because for router tests, we didn't call the method to setup the pg container directly. When we run all the tests together, the pg container will be called by other tests, and will be setup ready. But when run the router tests individually, we missed that part.
> 
> I added the "db_pg_container" in the router tests where we missed that. In fact, we only need to add it to the first test method in each router test file, it will only setup once, and will be shut down once all tests done. But to reduce confusion, I added to all router tests that missed it, so it looks like: Screen Shot 2023-04-11 at 4 12 00 PM

For Pytest, adding fixture into each function when needed is fine (but not just at the first one) because Pytest will determine when to run the fixture by its scope. In this case, your db_pg_container is in scope="session" (as intended) then it is good. If this test file is written in another style in Pytest (like 'grouping' or have 'setup' before each test) then it can be more obvious to developers the container session is up and ready before all tests.

server/backend/api/app/crud/crud_application.py
ianliuwk1019: Seems to me this query can be improved; but we can touch that next time doing implementation I think.

ianliuwk1019 (approved):
Thanks for the great hard work!

Issue #462, opened 2023-02-27
SPAR user/role assigment import


gormless87:
**Describe the task**
Import a set of user assignments to roles from SPAR team

**Acceptance Criteria**
- [ ] users and roles imported
- [ ] confirm users can login
- [ ] confirm users have correct role assignments

**Additional context**
- SPAR team works with DBA to export users/roles from ADAM and resolve any data quality issues
- validated users provided by SPAR team
- related decision note to be written up:  How will users to be imported be validated --call the IDIM api for each entry OR send list to IDIM to validate?

MCatherine1994:
All the SPAR roles are abstract role, which requires a forest client number when doing the user role assignment, but not all the SPAR users have a forest client number

MCatherine1994:
Setup a meeting with Sabina to chat about the SPAR user and roles, wondering if the data is correct for those missing the forest client number, and also some records show "OBSOLETE" if we want to ignore those.

MCatherine1994:
Got the script to read users from csv file and transfer into FAM. Need to understand more about the SPAR user role requirement, will let Olga lead to understand the SPAR requirement. Script should be ready.

basilv:
I'm confused by the status of this. Is there anything to code review? I don't think its ready for code review, as we don't know what to do yet about the SPAR organizational units that aren't forest clients.

ArogeG:
Blocked by the SPAR organizational unit that have forest client and those that don't .

ArogeG:
Decision to repurpose this ticket for a generic script and then create a new ticket for the SPAR solution.

PR #555, opened 2023-04-20
feat: #462 Write script to transfer user from csv file into FAM

MCatherine1994:
- Write a common script that could read users and roles from csv file and transfer them into FAM, start with the SPAR user sheet


scripts/kc_data_transfer/src/KeyCloakTransfer.py
```diff
@@ -117,9 +119,7 @@ def __init__(self):
         #   FOM_SUBMITTER_TESTING_00001011, FOM_SUBMITTER_TESTING_00001012... etc
         self.fom_submitter_parent_role_name = "FOM_SUBMITTER"
         self.fom_reviewer_role_name = "FOM_REVIEWER"
-        self.fam = FamWrapper(
-            fom_forest_client_role_parent=self.fom_submitter_parent_role_name
```
MCatherine1994, line 121: the fom_forest_client_role_parent parameter is not in use at all, just removed it

scripts/kc_data_transfer/src/KeyCloakTransfer.py
```diff
@@ -285,8 +285,6 @@ def __init__(self, fom_forest_client_role_parent: str):
             "Content-Type": "application/json",
             "Authorization": "Bearer " + constants.FAM_JWT
         }
-        self.fom_forest_client_roles_parent = fom_forest_client_role_parent
-        self.fom_forest_client_role_prefix = f"{fom_forest_client_role_parent}_"
```
MCatherine1994, line 289: these two parameters are not in use at all, removed

scripts/kc_data_transfer/src/KeyCloakTransfer.py
```diff
@@ -328,7 +326,7 @@ def get_roles(self, app_id) -> List[FamRole]:
         data = resp.json()
         return data
 
-    def get_role(
+    def get_role_id(
```
MCatherine1994, line 329: this method previously called get_role,  but is not in use at all, changed it to get_role_id and let it returns the role_id

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
+import csv
+import logging
+from KeyCloakTransfer import FamWrapper
+import constants
+
+LOGGER = logging.getLogger(__name__)
+
+
+class AppToFam:
```
ianliuwk1019, line 9: Since this is for csv, could we have documentation to show one example here or the readme.txt (showing maybe a title row - if there is, and the first row data) so that programmers are clear what columns, delimiter and format are expected from the file. Using Python csv, the column order (or index) probably isn't important, so what we can document what column name are expected.

Oh, maybe also specify what Python version this script you are running at would be helpful if in future it does not work for some reason from csv package.

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
...
+            else:
+                LOGGER.error("failed to get user name and user type, " +
+                             f"please check the data format: {user}")
+
+    def read_file_data(self, path: str):
+        data = []
+        column_title = []
+
+        with open(path, newline='', encoding='utf-8-sig') as csvfile:
+            file_content = csv.reader(csvfile, delimiter=',')
```
ianliuwk1019, line 69: Would it be easier to use `csv.DictReader(csvfile)` instead? It directly translates csv into dictionary for you without you to construct your dataset. It could be problematic to parse data your own (for example, if data row columns, does not match title row columns, or index).
The [Python csv documentation](https://docs.python.org/3.8/library/csv.html#csv.reader) has this example that is easy to understand:
```
>>> with open('names.csv', newline='') as csvfile:
...     reader = csv.DictReader(csvfile)
...     for row in reader:
...         print(row['first_name'], row['last_name'])
...
Eric Idle
John Cleese

>>> print(row)
{'first_name': 'John', 'last_name': 'Cleese'}
```

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
...
+                          self.get_forest_client(user["For Organization"]))
+                    self.fam.create_user_role_assignment(
+                        user_type_code=user_type_code,
+                        user_name=username,
+                        role_id=role_id,
+                        role_type_code="C",
+                        forest_client_number=self.get_forest_client(
+                            user["For Organization"]
+                        ),
+                    )
```
ianliuwk1019, line 56: Just wondering, what happen when `fam.create_user_role_assignment` encounters problem and what we like this csv parsing process to behave in this case? Did this happen before?

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
...
+                else:
+                    LOGGER.error("failed to find role, " +
+                                 f"please check the data format: {user}")
+            else:
+                LOGGER.error("failed to get user name and user type, " +
+                             f"please check the data format: {user}")
+
```
ianliuwk1019, line 63: Same wondering comment above.
When something happens, do we let it continue or stop? I am thinking if we let it continue to process the next csv row data, is it good idea to "**write**" out to another csv file so it will collect all of the rows that encountered error (with what error) and we know which user is having problem.

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
...
+        # get username and type from the "User" column in the csv file
+        # it should be in the format of "USERSTYPE\USERNAME"
+        user_info = user.split('\\')
+        if len(user_info) == 2:
+            if "IDIR" in user_info[0]:
+                return "I", user_info[1]
+            if "BCEID" in user_info[0]:
+                return "B", user_info[1]
+
+        return None, None
```
ianliuwk1019, line 33: Interesting, I didn't know Python can returns two values together!!
Thanks.

ianliuwk1019 (dismissed):

scripts/kc_data_transfer/readme.md
```diff
@@ -40,14 +40,16 @@ exists in the ./scripts/kc_data_transfer directory it will automatically
...
+All variables start with "KC" are for transfering user from keycloak. For transfering user from csv file, use the last four variables.
 -   KC_HOST - url to keycloak instance (e.g. https://oidc.gov.bc.ca)
 -   KC_CLIENTID - The client id that has been configured as a service account
 -   KC_SECRET - The secret for the above client
 -   KC_REALM - (e.g. 'ichqx89w' for the NR custom realm)
 -   KC_FOM_CLIENTID=the name of the client used for fom
 -   FAM_JWT - Your JWT token (steal from the docs page after authenticating or from your browser, see the section below)
 -   FAM_URL - Base FAM API url (e.g. https://7j9h7vm7ag.execute-api.ca-central-1.amazonaws.com/v1)
--   FOM_APP_NAME_IN_FAM - The name of the FOM app in the FAM database (e.g. "FOM_DEV")
+-   APP_NAME_IN_FAM - The name of the app in the FAM database (e.g. "FOM_DEV")
```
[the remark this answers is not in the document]
MCatherine1994: good catch Ian!!!

scripts/kc_data_transfer/src/CsvTransfer.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def transfer_user_to_fam(self):
+        app_id = self.fam.get_fom_app_id()
```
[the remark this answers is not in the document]
MCatherine1994: yeah, that's a good idea. The script originally was just wrote for FOM, so the naming is more fom specific. I'll change it

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: yes, it happens before. So the script will print out all there users that need to be transferred, and once the script run finished, we compare the number of users that need to be transferred with the number of actually user added in FAM, the FAM frontend shows the total number of users for each application. And if  we feel not confident, we could also log out the status code, that also indicates success or fail
Screen Shot 2023-04-24 at 3 36 36 PM

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: yeah, currently we just print out the error row. Last time when we run the script for FOM, we just run for all the users a few times, it will ignore duplicate users. But I do agree if the user list is large, we might want to write the error ones out, so next time we run on a smaller list. It depends on how much effort we want to put here, ideally we shouldn't get this data format error.

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: I tried this method, this one requires we do everything while the csv file is open, so we need to write everything under "with open('names.csv', newline='') as csvfile:", or copy the the whole DictReader data, I feel it's okay to leave it as what we have now

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: Added that in the README, under "Run script -> Note" section, feel free to review

ianliuwk1019: Great, thanks for adding that in!

scripts/kc_data_transfer/src/CsvTransfer.py
ianliuwk1019: Hm... not really sure what you meant... since both method open a file for readying every row...; it is just instead of doing custom row parsing by the column index into a data set, we take advantage on cleaner approach using the second method with just using the "titles" and does not rely on specific column "index". We don't really need to write everything under the "open()". Assigning user role can still be out of the open() loop.

scripts/kc_data_transfer/src/CsvTransfer.py
ianliuwk1019: > yeah, currently we just print out the error row. Last time when we run the script for FOM, we just run for all the users a few times, it will ignore duplicate users. But I do agree if the user list is large, we might want to write the error ones out, so next time we run on a smaller list. It depends on how much effort we want to put here, ideally we shouldn't get this data format error.

It is preferable log to console and a file, we lose track easily if we need to compare large data.
See if you want to take a bit time improve on that. I think I can pass.

ianliuwk1019 (approved):

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: yeah, it's just that currently we call the "read_file_data" method to get all user data, where the "read_file_data" method read the file and return the data.  But if I use "csv.DictReader", then it must keep the file open while go through the data, so we can't have "read_file_data" as a separate function, everything need to be done while the csv file open, so like:
"""
def transfer_user_to_fam(self):
        app_id = self.fam.get_app_id()

        with open('names.csv', newline='') as csvfile:
            users = csv.DictReader(csvfile)
            for user in users:
                .....line39 - line64
"""
If I just changed the "read_file_data" function to use "csv.DictReader", it complains that "ValueError: I/O operation on closed file."

Screen Shot 2023-04-25 at 12 37 41 PM

scripts/kc_data_transfer/src/CsvTransfer.py
MCatherine1994: yes, I would agree. Let me try it

Issue #179, opened 2022-09-07
Front-end Integrate with Forest Client API


ArogeG:
As a an application access administrator I want to see a list of forest client So That I can select a forest client for an assignment.

Additional Context
-- Integrate Forest Client API with frontend and back-end for validation.
--Allow users to select and search for a forest client by forest client name and ID.
--If Forest Client does not exists, return a message "Forest Client cannot be Found".

**Acceptance Criteria**
When I type in a FC ID I want validation (via API) and see company name 

**Definition of Done**
 Ready to Demo in Sprint Review
 Does what I have made have appropriate test coverage?
 Documentation and/or scientific documentation exists and can be found
 Peer Reviewed by 2 people on the team
 Manual testing of all PRs in Dev and Prod
 Merged

ianliuwk1019:
Current backend has "**Create User/Role Assignment**" **endpoint** done for the first draft version which takes "**forest client number**" as is as parameter in request body (without forest client name). Saving the record into fam_forest_client "client_name" column is currently hardcoded dummy data.

When we start integration with Forest Client API, backend endpoint logic will needs some adjustment to:
- Validate "**forest_client_number**" parameter sent is correct by checking with Forest Client API.
- Obtain Forest Client "name" from Forest Client API to save into the database record's **client_name** column.

basilv:
I'm unsure of the design to store the client name - which table is that again?

ianliuwk1019:
The '**client_name**' column is at fam_forest_client table.
![image.png](https://images.zenhubusercontent.com/62a9ff93d8506f14758ef03a/6d2f91b2-ff5a-4b48-a27f-5690f5b35585)

basilv:
It seems weird to update this table from individual add-role-assignment transactions

basilv:
As discussed in today's scrum - plan is to have FAM backend retrieve forest client name from forest client API and store in this table as part of add-role-assignment transaction. 

Also, FAM front-end will need to search/look up forest clients, which will need to go through a FAM backend API to call system-to-system to forest client API using secure token.

ianliuwk1019:
I am not sure why the `client_name` column were being removed from the `fam_forest_client` table (might be due to we had no forest client api integration last year)
We might need to see if there is a need to use Forest Client Name for display in FAM (and bring back the column/logic). For example on this screen (or any other page in future):
[...]

PR #565, opened 2023-04-25
feat: #179 Front end integration with Forest Client API

NickSaglioni:
Adjust Grant Access screen with some layout changes and integrate the front end with the back end Forest Client API.


frontend/src/components/GrantAccess.vue
```diff
@@ -94,6 +101,27 @@ function toRequestPayload(formData: any) {
 function statusSelected(evt: any) {
     formData.value.role = JSON.parse(evt.target.value);
 }
+
+async function getForestClientNumber(evt: string) {
```
MCatherine1994, line 105: I would suggest to change the parameter "evt" to a better name, like call it "forestClientNumber", when we name the parameter, it's better to give a name that make more sense, like in this case, we know this function will take a string.

frontend/src/components/GrantAccess.vue
```diff
@@ -210,22 +241,56 @@ function statusSelected(evt: any) {
...
+                            'is-invalid': errors.forestClientNumber,
+                        }"
                     />
                     <ErrorMessage
                         class="invalid-feedback"
                         name="forestClientNumber"
                     />
                 </div>
+                <div class="col-md-3 d-flex align-items-end">
+                    <button
```
MCatherine1994, line 257: I just wondering if we want to switch to primevue outlined button, cause otherwise we have to change it in the future anyway

frontend/src/components/common/ForestClientCard.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+    <div>
+        <b-card class="mb-2">
```
MCatherine1994, line 3: also the card, maybe we could use the primevue card?

frontend/src/components/common/Tag.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+    <div
+        class="tag"
```
MCatherine1994, line 3: Primevue provides the tag component as well

MCatherine1994:
Thanks Nick! Overall looks good!! Just think we should update the "frontend/public/env.json" file to get the latest cognito credentials for local

MCatherine1994:
And just in the case the forest client number is not correct, the "verify" button is not aligned with the "forest client" input box
image

frontend/src/components/GrantAccess.vue
```diff
@@ -94,6 +101,27 @@ function toRequestPayload(formData: any) {
...
+async function getForestClientNumber(evt: string) {
+    loading.value = true;
+    try {
+        forestClient.value = (await forestClientNumberApi.search(evt))
+            .data as FamForestClient[];
+    } catch (err: any) {
+        return Promise.reject(err);
+    } finally {
+        loading.value = false;
+    }
```
ianliuwk1019, line 114: Maybe the frontend id done in certain way that I forgot already, but seeing catching an error and return `return Promise.reject(err);` seems a bit strange as the api should already be returned something equivalent as Promise.reject(). Just wondering here.
Also, in catch block, why isn't it set `loading.value` to `false`?

frontend/src/components/common/ForestClientCard.vue
```diff
@@ -0,0 +1,54 @@
...
+            <b-card-text>
+                <Tag
+                    :text="props.status.description"
+                    :active="props.status.status_code === 'A' ? true : false"
```
ianliuwk1019, line 7: Doesn't the `===` comparison already return you `true : false`?

frontend/src/components/common/ForestClientCard.vue
```diff
@@ -0,0 +1,54 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import '@/assets/styles/styles.scss';
+
+.invalid {
+    margin-top: 0.25rem;
+    font-size: 0.875em;
+    color: #dc3545;
+}
```
ianliuwk1019, line 47: Is there a global .invalid previously or added to global?
Or maybe you want to refactor this and also hardcoded color next time?

ianliuwk1019 (changes requested):

frontend/src/components/GrantAccess.vue
```diff
@@ -23,6 +29,7 @@ const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone defa
 const apiServiceFactory = new ApiServiceFactory();
 const applicationsApi = apiServiceFactory.getApplicationApi();
 const userRoleAssignmentApi = apiServiceFactory.getUserRoleAssignmentApi();
+const forestClientNumberApi = apiServiceFactory.getForestClientApi();
```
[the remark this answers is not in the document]
NickSaglioni: Thanks for spotting that. I will fix it.

frontend/src/components/GrantAccess.vue
NickSaglioni: It is the same logic as in `grantAccess()` method. Should it be changed in both cases? For the second question, the catch block doesn't need to set the `loading.value` because it is already in the finally block as it needs to be set to `false` either for successful calls as for errored ones.

frontend/src/components/GrantAccess.vue
NickSaglioni: Thanks, it will be changed as suggested.

frontend/src/components/GrantAccess.vue
ianliuwk1019: I would suggest please find out why you need to do nothing but only to return `Promise.reject(err)`, maybe there were already some certain way how it handle the error at frontend (but I kind of doubt that)... 
Feel feel to leave it as is if you could not find out the behavior and if we don't returns `Promise.reject(err))` does give you problem, but could you at next frontend ticket when you work on also try to see if this can be fixed all together? Thank you!

Sorry, I eyes are too big to see `finally` :P Thanks for pointing that out.

NickSaglioni: I'll leave as is for now and work on that in the future as suggested.

frontend/src/components/common/ForestClientCard.vue
NickSaglioni: This is going to be changed in the grant access flow task, the card component will be changed, so I will leave it as is for now.

frontend/src/components/common/ForestClientCard.vue
NickSaglioni: Thanks, I will fix it

ianliuwk1019:
Hi @NickSaglioni , could you just have a brief description at this PR, currently it is blank. I am not sure if that would not show up anything at release log...
Thank you.

ianliuwk1019 (approved):

MCatherine1994 (approved):

Issue #551, opened 2023-04-12
Landing page front-end development


NickSaglioni:
Create a Landing page as designed in Figma:

https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?node-id=226-85209&t=p47oPA2tKvX4iE6f-4

**Acceptance Criteria**
When I access FAM I should see first the Landing page as designed and navigate accordingly.

PR #572, opened 2023-05-03
feat: #551 landing page

NickSaglioni:
- Adding landing page and styling with primevue and fsa-theme;
- Creating a layout logic to handle pages that have different layouts;


frontend/src/assets/styles/styles.scss
```diff
@@ -5,3 +5,7 @@
 
 // Components
 @import './components/form-elements.scss';
+
+@import '@bcgov-nr/nr-fsa-theme/design-tokens/type-family.scss';
```
MCatherine1994, line 9: We don't need this one, the font design token already be called in the stylesheet

frontend/src/main.ts
```diff
@@ -28,6 +28,16 @@ import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
...
 
+import PrimeVue from 'primevue/config';
+
+// use bootstrap4 as default style
+import 'primevue/resources/themes/bootstrap4-light-blue/theme.css';
+import 'primevue/resources/primevue.min.css';
+import 'primeicons/primeicons.css';
+
+// import the component override style sheet for primevue
+import '@bcgov-nr/nr-fsa-theme/style-sheets/primevue-components-overrides.scss';
```
MCatherine1994, line 39: Let's remove this one, cause we already imported it in the `asses/styles/styles/scss` file. But we need adjust the import order, put `import './assets/styles/styles.scss';` here. Cause if we import the `./assets/styles/styles.scss` file first, and then import the default primevue style, it will apply the default one on top of our custom stylesheet. The import order of the style file is important

MCatherine1994:
I'm still getting that can't find font file error, but think I know where is wrong, I'll fix it in the common theme repo and release a new version. Thanks!!

MCatherine1994:
Also wondering since we already have a landing a page, when we logout, maybe no need to redirect to this "home" page first and then redirect to the landing page? Could we go directly to the landing page when a user logout?
image

MCatherine1994:
Just wondering since we already have the landing page to handle login, do we still need the login button on other pages like "http://localhost:5173/home"?
Screen Shot 2023-05-08 at 12 02 14 PM

NickSaglioni:
> Just wondering since we already have the landing page to handle login, do we still need the login button on other pages like "http://localhost:5173/home"? maybe this is a UX design question....
> Screen Shot 2023-05-08 at 12 02 14 PM

Home is the starting point page when the user is already logged in. It doesn't work redirecting to the landing page because it is supposed to only handle login, not to be the start point of the application. The Home Screen will be eventually replaced by the dashboard page in the future.

frontend/src/assets/styles/styles.scss
NickSaglioni: Thanks, I'll update it

MCatherine1994 (approved):

ianliuwk1019:
> > Just wondering since we already have the landing page to handle login, do we still need the login button on other pages like "http://localhost:5173/home"? maybe this is a UX design question....
> > Screen Shot 2023-05-08 at 12 02 14 PM
> 
> Home is the starting point page when the user is already logged in. It doesn't work redirecting to the landing page because it is supposed to only handle login, not to be the start point of the application. The Home Screen will be eventually replaced by the dashboard page in the future.

Just curious, so does it mean after landing page redirects for user signs in from Cognito, it redirects back to this 'home' page or `application` page?

frontend/src/App.vue
```diff
@@ -1,19 +1,25 @@
...
+const layout_component = shallowRef<Component>();
+watch(
+    () => route.meta.layout ?? 'SimpleLayout',
+    (layout) => {
+        layout_component.value = defineAsyncComponent(
+            () => import(`@/layouts/${layout}.vue`)
+        );
+    },
+    { immediate: true }
+);
```
ianliuwk1019, line 16: This is interesting. Why are we dynamically import components? Is there a need for changing layout design? I thought layout is consistent maybe I am not clear for new design...

NickSaglioni: We are going to have two layouts now, the landing page doesn't have a header and was not going to have a footer (but now it does). That's why we have this dynamic layout change. In the future, if we need a third layout, it can be easily implemented.

NickSaglioni:
> > > Just wondering since we already have the landing page to handle login, do we still need the login button on other pages like "http://localhost:5173/home"? maybe this is a UX design question....
> > > Screen Shot 2023-05-08 at 12 02 14 PM
> > 
> > 
> > Home is the starting point page when the user is already logged in. It doesn't work redirecting to the landing page because it is supposed to only handle login, not to be the start point of the application. The Home Screen will be eventually replaced by the dashboard page in the future.
> 
> Just curious, so does it mean after landing page redirects for user signs in from Cognito, it redirects back to this 'home' page or `application` page?

It goes to `application`.

frontend/src/main.ts
```diff
@@ -28,6 +27,14 @@ import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
 import { faTrashCan } from '@fortawesome/free-regular-svg-icons';
 import ErrorService from './services/ErrorService';
 
+import PrimeVue from 'primevue/config';
+
+// use bootstrap4 as default style
+import 'primevue/resources/themes/bootstrap4-light-blue/theme.css';
```
ianliuwk1019, line 33: Is bootstrap5 supported? Just wondering...

MCatherine1994: Hi Ian, no primevue doesn't support bootstrap 5, it has a list of [built in theme](https://primevue.org/theming/#builtinthemes), but since most of us like bootstrap more, we just picked that one. Hopefully later it can provide bootstrap 5, but since we need customization anyway, should be good for now

ianliuwk1019: Sounds good, thanks for explanation!

frontend/src/components/Landing.vue
```diff
@@ -0,0 +1,54 @@
...
+import logo from '../assets/images/bc-gov-logo.png';
+import Button from 'primevue/button';
+</script>
+
+<template>
+    <div class="full-width">
+        <div class="row landing-grid">
+            <div class="col-sm-6 col-md-7 col-lg-7">
+                <img :src="logo" alt="BCGov Logo" width="160" class="logo" />
+                <h1 data-testid="landing-title" class="landing-title">
```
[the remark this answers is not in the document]
NickSaglioni: This was from a testing that I was trying to do but didn't work, thanks for spotting it. It should be `id`. I will correct it.

ianliuwk1019 (approved):

MCatherine1994 (approved):

Issue #560, opened 2023-04-21
Application error when only has access to one application


MCatherine1994:
**Describe the task**
When the user only has access to one application, we will skip the "Select Application" page, and show the "Manage Access" page directly, but there will be an error message show on the screen
![image.png](https://images.zenhubusercontent.com/6201a0f0c6815e7dfafa3f78/8f1f7c5e-b635-44be-a8e1-b0ee7fa57cb2)
![Screen Shot 2023-04-21 at 9.32.24 AM.png](https://images.zenhubusercontent.com/6201a0f0c6815e7dfafa3f78/21f3bce2-be7d-479b-8422-fff1ce574875)

**Acceptance Criteria**
- [ ] No error show on the screen, when the user only has access to one application

ArogeG:
The error seems to come-up when the token is being refreshed.

MCatherine1994:
Found another case that could trigger this error:

When we are on the "Manage Access" page, and then we refresh the page, this error will shows again.

It's because refresh page will reset our local "selectedApplication" variable, and then triggers a redirect back to the select application page.

PR #573, opened 2023-05-03
fix: #560 Fix application error when refresh manage access page, and when only has access to one app

MCatherine1994:
- fix application error when refresh manage access page, and when only has access to one app refs: #560


frontend/src/components/common/RadioGroup.vue
```diff
@@ -24,7 +24,6 @@ const emit = defineEmits(['change']);
 const selectedValue = ref<string>('');
 onBeforeMount(() => {
     selectedValue.value = props.initialValue;
-    valueChanged();
```
MCatherine1994, line 27: remove this call, otherwise it triggers the filterEnv method to call everytime we redirect the page, which will cleanup the local storage, not need to have this call here

frontend/src/components/SelectApplication.vue
```diff
@@ -68,10 +71,14 @@ const filteredOptions = computed(() => {
                     <select
                         id="applicationSelect"
                         class="form-select"
-                        v-model="selectedApplication"
+                        :value="JSON.stringify(selectedApplication)"
+                        @change="(e: Event) => setSelectedApplication(e.target ? (e.target as HTMLTextAreaElement).value : null)"
```
ianliuwk1019, line 75: Looks good to me but not sure if you want to move the whole logic into <script> section?

frontend/components.d.ts
```diff
@@ -0,0 +1,33 @@
...
+    ManageAccess: typeof import('./src/components/ManageAccess.vue')['default']
+    NotFound: typeof import('./src/components/NotFound.vue')['default']
+    PageTitle: typeof import('./src/components/PageTitle.vue')['default']
+    RadioGroup: typeof import('./src/components/common/RadioGroup.vue')['default']
+    RouterLink: typeof import('vue-router')['RouterLink']
+    RouterView: typeof import('vue-router')['RouterView']
+    SelectApplication: typeof import('./src/components/SelectApplication.vue')['default']
+    Tag: typeof import('./src/components/common/Tag.vue')['default']
+  }
+}
```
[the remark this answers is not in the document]
MCatherine1994: I think this file is auto generated when we run the frontend, I'll add it in .gitignore

frontend/src/services/ApplicationState.ts
```diff
@@ -5,7 +5,20 @@ import { computed, ref } from 'vue';
 export const applicationsUserAdministers = ref<FamApplication[]>([]);
 
 // The application selected by the user to admin
-export const selectedApplication = ref<FamApplication | null>();
+export const selectedApplication = ref<FamApplication | null>(
+    localStorage.getItem('CURRENT_SELECTED_APPLICATION')
```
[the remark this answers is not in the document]
MCatherine1994: yeah, good idea, I'll define it in a variable

frontend/src/components/SelectApplication.vue
MCatherine1994: yeah, think it should be fine, moved into the script section

ianliuwk1019 (approved):

Issue #194, opened 2022-09-20
API End to End Test on FAM (Smoke test)


ArogeG:
As a developer, I want to make sure the deployment work so that when an end-point is called it gets through to the gateway and works fine. 
End to End test: Possibility of re-using the unit tests for End to End. 
This test is to check connectivity. -This can be done with one end-point that goes through the gateway and talks to the database. 

AC:
Tests run and pass in pipeline 
When an endpoint is called, it goes through the API gateway and the FAM database and provides the right response to the client.

PR #609, opened 2023-05-30
ci: #194 add deployment smoke test

ConradBoydElliottGustafson:
The 9 lines of BASH that tried to kill me


server/backend/api/app/routers/router_smoke_test.py
```diff
@@ -0,0 +1,30 @@
...
+    db: Session = Depends(database.get_db),
+):
+
+    """
+    List of different applications that are administered by FAM
+    """
+    LOGGER.debug(f"running router ... {db}")
+    fam_apps = db.query(models.FamApplication).all()
+    if len(fam_apps) == 0:
+        response.status_code = 417
```
ianliuwk1019, line 28: Curious on reason why using 4xx client error, not 5xx server error?

.github/workflows/reusable_terraform_server.yml
```diff
@@ -151,4 +151,25 @@ jobs:
...
+
+      - id: smoke_test
+        name: Smoke Test
+        working-directory: ${{ env.TG_SRC_PATH }}/${{ inputs.environment_name }}
+        env:
+          tfc_workspace: "${{ secrets.licenceplate }}-${{ inputs.environment_name }}"
+        run: |
+          if [[ "apply" == "${{ inputs.tf_subcommand }}" ]]
+          then
+            shopt -s lastpipe
```
ianliuwk1019, line 164: Hi Conrad, what does this do?

server/backend/api/app/routers/router_smoke_test.py
```diff
@@ -0,0 +1,30 @@
...
+from api.app.models import model as models
+
+
+LOGGER = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.get("", status_code=200)
+def get_applications(
```
basilv, line 17: Seems weird to have the method for smoke testing called get_applications

server/backend/api/app/routers/router_smoke_test.py
ConradBoydElliottGustafson: It's not actually an error if there is no data. Just an unexpected result. Doesn't matter because anything other than 2xx will fail the smoke test.

server/backend/api/app/routers/router_smoke_test.py
ConradBoydElliottGustafson: Good point. I'll rename it.

basilv (approved):
Looks good.

.github/workflows/reusable_terraform_server.yml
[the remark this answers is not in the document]
ConradBoydElliottGustafson: shopt -s lastpipe
it means that when you chain commands using the pipe operator | only the last command is assigned to the variable. The default behaviour in bash in Github actions is to put all the values in an array. The default behaviour in my Windows subsystem for Linux environment (Ubuntu) was the opposite. This took me 2 days to figure out!!!

server/backend/api/app/routers/router_smoke_test.py
ConradBoydElliottGustafson: Renamed

PR #611, opened 2023-05-30
chore: Remove unnecessary "@types/node": "^18.0.0" from frontend.

ianliuwk1019:
Not directly using @types/node, so remove it.
If we do a check with: "npm list @types/node", many other libraries in package.json have "@types/node@18.16.16" dependency already.


basilv (approved):
Looks good.

PR #603, opened 2023-05-23
chore(deps): update dependency requests to v2.31.0 [security]

bcgov-renovate:
This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| [requests](https://requests.readthedocs.io) ([source](https://togithub.com/psf/requests), [changelog](https://togithub.com/psf/requests/blob/master/HISTORY.md)) | minor | `==2.28.1` -> `==2.31.0` |

### Impact

Since Requests v2.3.0, Requests has been vulnerable to potentially leaking `Proxy-Authorization` headers to destination servers, specifically during redirects to an HTTPS origin. This is a product of how `rebuild_proxies` is used to recompute and [reattach the `Proxy-Authorization` header](https://togithub.com/psf/requests/blob/f2629e9e3c7ce3c3c8c025bcd8db551101cbc773/requests/sessions.py#L319-L328) to requests when redirected. Note this behavior has _only_ been observed to affect proxied requests when credentials are supplied in the URL user information component (e.g. `https://username:password@proxy:8080`).

**Current vulnerable behavior(s):**

1. HTTP → HTTPS: **leak**
2. HTTPS → HTTP: **no leak**
3. HTTPS → HTTPS: **leak**
4. HTTP → HTTP: **no leak**

For HTTP connections sent through the proxy, the proxy will identify the header in the request itself and remove it prior to forwarding to the destination server. However when sent over HTTPS, the `Proxy-Authorization` header must be sent in the CONNECT request as the proxy has no visibility into further tunneled requests. This results in Requests forwarding the header to the destination server unintentionally, allowing a malicious actor to potentially exfiltrate those credentials.

The reason this currently works for HTTPS connections in Requests is the `Proxy-Authorization` header is also handled by urllib3 with our usage of the ProxyManager in adapters.py with [`proxy_manager_for`](https://togithub.com/psf/requests/blob/f2629e9e3c7ce3c3c8c025bcd8db551101cbc773/requests/adapters.py#L199-L235). This will compute the required proxy headers in `proxy_headers` and pass them to the Proxy Manager, avoiding attaching them directly to the Request object. This will be our preferred option going forward for default usage.

### Patches
Starting in Requests v2.31.0, Requests will no longer attach this header to redirects with an HTTPS destination. This should have no negative impacts on the default behavior of the library as the proxy credentials are already properly being handled by urllib3's ProxyManager.

[...]

PR #599, opened 2023-05-17
chore(deps): update terraform terraform-aws-modules/rds-aurora/aws to v8

bcgov-renovate:
This PR contains the following updates:

| Package | Type | Update | Change |
|---|---|---|---|
| [terraform-aws-modules/rds-aurora/aws](https://registry.terraform.io/modules/terraform-aws-modules/rds-aurora/aws) ([source](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora)) | module | major | `7.7.1` -> `8.2.0` |

---

### Release Notes

<details>
<summary>terraform-aws-modules/terraform-aws-rds-aurora</summary>

### [`v8.2.0`](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/HEAD/CHANGELOG.md#&#8203;820-httpsgithubcomterraform-aws-modulesterraform-aws-rds-auroracomparev811v820-2023-05-24)

[Compare Source](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/compare/v8.1.1...v8.2.0)

##### Features

-   Upgrade aws provider version and update examples ([#&#8203;382](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/issues/382)) ([b792d84](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/commit/b792d84b68086e1458fb9ef0ee07dc982510f4ed))

##### Bug Fixes

-   Tag security group with correct name ([#&#8203;380](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/issues/380)) ([71d23db](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/commit/71d23db753184f416e06c0d0c6f4f73ec8d95184))

### [`v8.1.1`](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/HEAD/CHANGELOG.md#&#8203;811-httpsgithubcomterraform-aws-modulesterraform-aws-rds-auroracomparev810v811-2023-05-12)

[Compare Source](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/compare/v8.1.0...v8.1.1)

### [`v8.1.0`](https://togithub.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/HEAD/CHANGELOG.md#&#8203;810-httpsgithubcomterraform-aws-modulesterraform-aws-rds-auroracomparev802v810-2023-05-12)

[...]

PR #596, opened 2023-05-16
chore(deps): update dependency python-dotenv to v1

bcgov-renovate:
This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| [python-dotenv](https://togithub.com/theskumar/python-dotenv) | major | `==0.21.0` -> `==1.0.0` |
| [python-dotenv](https://togithub.com/theskumar/python-dotenv) | major | `==0.20.0` -> `==1.0.0` |

---

### Release Notes

<details>
<summary>theskumar/python-dotenv</summary>

### [`v1.0.0`](https://togithub.com/theskumar/python-dotenv/blob/HEAD/CHANGELOG.md#&#8203;100)

[Compare Source](https://togithub.com/theskumar/python-dotenv/compare/v0.21.1...v1.0.0)

**Fixed**

-   Drop support for python 3.7, add python 3.12-dev ([#&#8203;449](https://togithub.com/theskumar/python-dotenv/issues/449) by \[[@&#8203;theskumar](https://togithub.com/theskumar)])
-   Handle situations where the cwd does not exist. ([#&#8203;446](https://togithub.com/theskumar/python-dotenv/issues/446) by \[[@&#8203;jctanner](https://togithub.com/jctanner)])

### [`v0.21.1`](https://togithub.com/theskumar/python-dotenv/blob/HEAD/CHANGELOG.md#&#8203;0211---2023-01-21)

[Compare Source](https://togithub.com/theskumar/python-dotenv/compare/v0.21.0...v0.21.1)

**Added**

-   Use Python 3.11 non-beta in CI ([#&#8203;438](https://togithub.com/theskumar/python-dotenv/issues/438) by \[[@&#8203;bbc2](https://togithub.com/bbc2)])
-   Modernize variables code ([#&#8203;434](https://togithub.com/theskumar/python-dotenv/issues/434) by \[[@&#8203;Nougat-Waffle](https://togithub.com/Nougat-Waffle)])
-   Modernize main.py and parser.py code ([#&#8203;435](https://togithub.com/theskumar/python-dotenv/issues/435) by \[[@&#8203;Nougat-Waffle](https://togithub.com/Nougat-Waffle)])
-   Improve conciseness of cli.py and **init**.py ([#&#8203;439](https://togithub.com/theskumar/python-dotenv/issues/439) by \[[@&#8203;Nougat-Waffle](https://togithub.com/Nougat-Waffle)])
[...]

PR #612, opened 2023-05-30
fix: #592 renovate dashboard backend

ConradBoydElliottGustafson:
#592 

Batches up the changes from renovate: #603 #596 #588 

Leaving #599 out of this because it causes big problems and is a CI change, not a code change. It updates the terraform module that deploys our database. Needs a lot of investigation and testing.


basilv (approved):
Looks good.

Issue #594, opened 2023-05-16
Cleanup alembic remnants from backend code


gormless87:
**Describe the task**
Follow up from #474 - remove unused code

**Acceptance Criteria**
- [ ] unused code removed
- [ ] app still works
- [ ] tests still run
- [ ] pipeline updates to remove sql lite content
- [ ] documentation adjusted accordingly (readme)

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
This ticket is mostly completed at PR#614. Pipeline is running good as well.
The one thing that's not directly "**alembic**" related but indirectly "**sqllite**" related is the "model.py" for Python ORM.
This I feel should be separate into a new ticket itself as a direct implementation ticket follows from #474.
The ORM "model.py" itself probably contains just few tweaks from the past specific for **"sqlite" testing**. 
We can generate **a new ORM "model"** to use, but in new ticket, it will mostly be focusing on detail verification at the model: our test pipeline still runs and frontend all business functions are still working (hitting as many endpoints as we can), login auth and its returned roles still correct etc... as a new complete model generated from first try of "**sqlacodegen**" tool could potentially missed something and hidden (such as mentioned in ticket #474 from Catherine's investigation). 
@gormless87

gormless87:
@ianliuwk1019 - I noticed this issue:

Update postgres test container to start docker just with postgres and flyway service#584

IT seems you may have addressed it with  PR#614

ianliuwk1019:
Add #631 to separately dealing with generating new models for backend from sqlacodegen tool.

PR #614, opened 2023-06-01
chore: #594 backend cleanup alembic remnants

ianliuwk1019:
This PR #594 contains changes for:

* Remove server/backend/api/alembic migration folder.
* Remove server/backend/tests folder related to sqllite.
* Update some imports for tests.
* Remove sqllite tests from unit-test workflow.
* Update readme.
* Use lib "**pytest-dotenv**" to load our '**local-dev.env**' with **pytest.ini** to unify backend just need **one** .env file without duplicating the values in other places.
* Modify docker-compose.yml to "extends" from **docker-base-service.yml** so FAM can have separate **docker-compose-testcontainer.yml** also extends from base services but leaving **api service** out from the use in testcontainer workflow, so no need to spin out redundant 'api' container. This saves about 1/3 of running unittest workflwo time.
* Adjust ".github/workflows/reuseable_api_unittests.yml" to work from "**server/backend**" "**working-directory**".


basilv:
The PR should probably be named a 'chore' rather than a 'fix' in terms of updating version number and release notes, but that's a nitpicky point.

ConradBoydElliottGustafson:
> The PR should probably be named a 'chore' rather than a 'fix' in terms of updating version number and release notes, but that's a nitpicky point.

I think anything other than "fix" or "feat" does not bump the version number. Pretty sure we want a version number bump for this? Am I wrong?

basilv:
Hard to say, if we are only changing unit test setup, and there's nothing different with the actual deployed app then no need for a deploy or version change in theory. But lots of files have changed...

In practice, I'm fine with a version change for this, I just was trying to fit this within the release-notes model.

ianliuwk1019:
> Hard to say, if we are only changing unit test setup, and there's nothing different with the actual deployed app then no need for a deploy or version change in theory. But lots of files have changed...
> 
> In practice, I'm fine with a version change for this, I just was trying to fit this within the release-notes model.

Thanks Basil. I changed to "chore".

docker-compose.yml
```diff
@@ -2,58 +2,17 @@
...
-      - FLYWAY_PLACEHOLDERS_client_id_dev_forest_client_oidc_client=xxqiHFmwG8j1cVAz7NgtknaZOg
-      - FLYWAY_PLACEHOLDERS_client_id_test_forest_client_oidc_client=dm5Xkmomnq0gbwBiXiN5LgAnah
-      - FLYWAY_PLACEHOLDERS_client_id_prod_forest_client_oidc_client=KdnD2eGS3Zcx494p04yMFhDwSe
+    extends:
+      file: docker-base-services.yml
+      service: fam-flyway
     depends_on:
       fam-database:
         condition: service_healthy
-    networks:
```
ConradBoydElliottGustafson, line 55: Do we not need the network fam anymore?

ConradBoydElliottGustafson: I tested locally, seems to work fine without the network explicitly defined.

ConradBoydElliottGustafson (approved):
I rebased the branch for you.

Issue #649, opened 2023-06-13
Refactor Tests


ConradBoydElliottGustafson:
The tests are kind of a mess. Now that we have the testcontainer mechanism working and the alembic stuff is all gone, it's a good time to get these tests working properly.

[x] All tests should use a containerized database (unless they don't need a database)
[x] All tests should operate in a transaction that is rolled back after each test is run
[x] Tests should be able to work with either the testcontainer (self-contained) or with a persistent container (for speed)
[x] Get rid of the race condition in the test container set up (sleep 6 seconds is just a guess that depends on the system)

PR #650 auth function refactor
PR #651 api refactor

PR #650, opened 2023-06-13
ci(auth_function): #649 refactor tests

ConradBoydElliottGustafson:


server/auth_function/config.py
ConradBoydElliottGustafson, line 1: Production code should not have test code mixed in. Removed and used monkeypatching to make the behaviour change when testing.

server/auth_function/lambda_function.py
ConradBoydElliottGustafson, line 1: Refactored the lambda function for testability. No need for object-oriented here.

server/auth_function/pytest.ini
ConradBoydElliottGustafson, line 1: Same setup as the api tests. Reusing the local-dev.env file from the api.

server/auth_function/test/conftest.py
ConradBoydElliottGustafson, line 1: moved all the test fixtures to this file to simplify setup and get around some weird errors in VS Code

server/backend/local-dev.env
ConradBoydElliottGustafson, line 1: The only thing that needs root access to the DB is flyway. Everything else should connect with a proxy account

server/auth_function/test/lamda_function_test.py
ConradBoydElliottGustafson, line 1: Major refactoring of the tests. The main reason is so that they each run in a separate transaction that gets rolled back at the end of the test.

.github/workflows/reusable_auth_function_ci.yml
ConradBoydElliottGustafson, line 1: Changed the GH action to be the same as the API test. Previously it was depending on GHA to spin up a DB, now it is done by testcontainers library in code.

server/auth_function/pytest.ini
ianliuwk1019: Does another "pytest.ini" (in **server/auth_function/test/pytest.ini**) need to be removed?

ianliuwk1019 (commented):
Refactoring looks cool!
Just one (I think maybe) unused pytest.ini is left there.

server/auth_function/pytest.ini
ConradBoydElliottGustafson: Yes it does! Good catch

ianliuwk1019 (dismissed):

ianliuwk1019 (approved):

PR #668, opened 2023-06-27
ci: sonar cloud frontend

DerekRoberts:


ConradBoydElliottGustafson (changes requested):
Looks great! I'm just wondering about the file ci_auth_artifact.yml

My guess is that Derek wanted to remove the zip file stuff from the sonar process. It's good to have the zip file build tested on PR and on merge to main. I think we should also test the API zip file process as well. I'm fine with having it in a file separate from the sonar workflow, even though it means an extra checkout. These should run in parallel anyway.

My suggestion is to add the API zip file build process from "ci_backend.yml" back into the ci process so we know if the build is failing.

DerekRoberts:
@ConradBoydElliottGustafson The zip artifact is still there, but in its own workflow.  Want them rolled together/

DerekRoberts:
@ConradBoydElliottGustafson Lets get this merged to allow main branch comparisons.  Zip artifacts can be handled in the next PR.  Also, scope creep!

ConradBoydElliottGustafson (approved):
Will open another ticket to get the zip file for API added to the ci_auth_artifact.yml

PR #674, opened 2023-06-30
ci: adjust sonar for auth

DerekRoberts:
Sonar Clould isn't generating coverage for auth. This is confusing, because coverage appears to be picked up.

From Sonar run:
```
INFO: Python test coverage
INFO: Parsing report '/github/workspace/server/auth_function/coverage.xml'
```

Small adjustment made, since an error was coming from relative_files in .coveragerc in the [report] section.
```
.venv/lib/python3.11/site-packages/coverage/config.py:324: CoverageWarning: Unrecognized option '[report] relative_files=' in config file .coveragerc
    warn(
...
```

Update: Working!  Combination of .coverragerc and sonar params.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -38,74 +38,74 @@ jobs:
           projectBaseDir: server/auth_function
           args: >
             -Dsonar.exclusions=config.py
-            -Dsonar.host.url=https://sonarcloud.io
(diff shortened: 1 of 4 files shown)


ConradBoydElliottGustafson (approved):
All three "modules" are reporting correctly. We'll see again if they work after a merge to main.

PR #683, opened 2023-07-07
ci: 475 eliminate pet names

ConradBoydElliottGustafson:
This ticket gave me a LOT of problems. There is a bug in Terraform and/or AWS in that when you run it the API-Gateway and the API-Lambda both get renamed. Terraform is correct and they should get stitched back together, but they don't. The solution is to go into AWS and manually DELETE the API-Gateway completely, then re-run the jobs. You know it is working when the smoke test passes.


terraform/dev/terragrunt.hcl
ConradBoydElliottGustafson, line 1: These 2 variables are no longer being used and were issuing warnings. Deleted

terraform/prod/terragrunt.hcl
ConradBoydElliottGustafson, line 1: These 2 variables are no longer being used and were issuing warnings. Deleted

server/backend/api/app/routers/router_smoke_test.py
ConradBoydElliottGustafson, line 1: This broke when an underlying library was updated. Found the error and fixed it, as well as put in an exception log in case the test fails and we want to investigate why.

.vscode/launch.json
ConradBoydElliottGustafson, line 1: Figured out how to create a launch configuration for the API in VS Code for debugging. Just checked it in as a convenience for the devs.

.github/workflows/reusable_terraform_server.yml
ConradBoydElliottGustafson, line 1: Found the bug that was killing the smoketest script. Line 115 corrects the terraform output so that it can be parsed correctly by the bash script.

.github/workflows/tools_destruction.yml
ConradBoydElliottGustafson, line 1: Cleaned up an initial checkin to main for the tools_destruction script

infrastructure/frontend/cloudfront.tf
ConradBoydElliottGustafson, line 1: Got rid of random name

infrastructure/server/api-gateway.tf
ConradBoydElliottGustafson, line 1: Using the data variable for DRY

infrastructure/server/aurora-v2.tf
ConradBoydElliottGustafson, line 1: Random name not being used -- just cleanup

infrastructure/server/auth_lambda.tf
ConradBoydElliottGustafson, line 1: Got rid of random name

infrastructure/server/fam_api.tf
ConradBoydElliottGustafson, line 1: Got rid of random name

infrastructure/server/flyway.tf
ConradBoydElliottGustafson, line 1: Got rid of random names

MCatherine1994:
Hi Conrad, all looks good, thanks so much! I just want to get some clarification, so after we get rid of the random pet name and use a fixed name, when we rerun terrforam, we won't get complain that the name already is in use?

MCatherine1994 (approved):

Issue #653, opened 2023-06-19
Implement UI for notifications


OlgaLiber2:
**Describe the task**
Implement UI for all notifications in FAM

**Acceptance Criteria**
- [ ] All notifications in FAM, across all pages, are now displayed with the new UI. 

**Additional context**
- Add any other context about the task here.
- Or here

Issue #648, opened 2023-06-12
Create a profile component for FAM


OlgaLiber2:
Describe the task**
A clear and concise description of what the task is.

**Acceptance Criteria**
- [ ] Profile component exists and available on all pages

**Additional context**
- *![Capture-profile.PNG](https://images.zenhubusercontent.com/63b5e912a39d01413a894255/779bb6ee-61ab-44d9-bad8-22970f9a7b99)*

PR #699, opened 2023-07-14
feat: 616 frontend main page

NickSaglioni:
- New dashboard screen;
- Grant access flow pages redesigned with the new design system;
- Removed footer;
- Updated header;
- profile component created (#648);
- sidebar component created;
- new notifications (#653)


frontend/src/components/GrantAccess.vue
```diff
@@ -1,34 +1,40 @@
...
     FamApplicationRole,
     FamForestClient,
     FamUserRoleAssignmentCreate,
 } from 'fam-api';
 import { onMounted, ref } from 'vue';
-import { useToast } from 'vue-toastification';
 import { Form as VeeForm, Field, ErrorMessage } from 'vee-validate';
 import router from '@/router';
 import { number, object, string } from 'yup';
+import Button from 'primevue/button';
```
MCatherine1994, line 17: Since we already created a reusable button component under the common folder, let's change to use that one. Although we need to add more properties in that reusable button component.

frontend/src/components/common/cards/SummaryCard.vue
```diff
@@ -0,0 +1,81 @@
+<script setup lang="ts">
+import { ref, type PropType, onMounted } from 'vue';
+import Card from 'primevue/card';
+import Button from 'primevue/button';
```
MCatherine1994, line 4: let's use the button component under the common folder

frontend/src/components/common/Icon.vue
```diff
@@ -0,0 +1,54 @@
+<template>
```
MCatherine1994, line 1: As discussed, we'll remove this Icon file later in the new ticket, let's use the carbon icons.

frontend/src/components/common/Sidebar.vue
```diff
@@ -0,0 +1,90 @@
+<template>
```
MCatherine1994, line 1: Let's rename this file to NavBar, it's a little bit confused compare with the ProfileSidebar.

frontend/src/components/common/cards/ForestClientCard.vue
```diff
@@ -0,0 +1,73 @@
+<script setup lang="ts">
+import type { FamForestClientStatus } from 'fam-api/dist/model/fam-forest-client-status';
+import { ref, type PropType } from 'vue';
+import Card from 'primevue/card';
+
+const selected = ref('');
```
MCatherine1994, line 6: This variable is not in use anymore, let's remove it

frontend/src/components/dialog/Dialog.vue
```diff
@@ -1,103 +1,50 @@
...
-                <button type="button" class="btn btn-info" @click="close">
-                    {{ cancelText }}
-                </button>
-            </div>
-        </div>
-    </vue-final-modal>
-</template>
-
 <script setup inherit-attrs="false" lang="ts">
+import Dialog from 'primevue/dialog';
```
MCatherine1994, line 2: I think we can put this file under the common folder, no need to keep it separated.

frontend/src/assets/styles/styles.scss
```diff
@@ -1,9 +1,22 @@
...
+
 @import 'base';
 
 // Components
 @import './components/form-elements.scss';
 
 @import '@bcgov-nr/nr-fsa-theme/style-sheets/primevue-components-overrides.scss';
+
+:root {
+    --font-family: map.get(typeFamily.$type-family, 'sans');
```
ianliuwk1019, line 17: My css knowledge is limited. My understanding is here is to set "--font-family" **css variable** not the **css style**.
Curious that is this variable being used anywhere? Since at the line 21 below this line in the "body {}" it does not use this variable. So my impression on this line is it is only declare this variable but not being used. Or maybe I am wrong, could you explain a bit? Thank you.

MCatherine1994:
I got this error in the console after I logged in

image

frontend/src/components/common/NotificationMessage.vue
```diff
@@ -0,0 +1,62 @@
...
+    color: $text-primary;
+}
+
+.message-container {
+    position: relative;
+    align-items: center;
+}
+
+.p-message {
+    position: absolute;
```
MCatherine1994, line 58: we should remove this line

frontend/src/assets/styles/styles.scss
```diff
@@ -1,9 +1,22 @@
+@use 'sass:map';
+@use '@bcgov-nr/nr-fsa-theme/design-tokens/type-family.scss' as typeFamily;
+
 // Import all of Bootstrap's CSS
 @import '~bootstrap/scss/bootstrap';
 
+@import '../../../node_modules/primeicons/primeicons.css';
```
ianliuwk1019, line 7: Just for reference:
I think it is better not to use "../../.", however, if after build and packaged and it works, we can leave it for now.
It should be possible define a alias path for it.
Actually, if looking at line 5 for the '~bootstrap/scss/bootstrap', it is from "node_module/bootstrap" I believe.
And the configuration is probably set in vite.config.ts previously when Vue project was setup.
So if we are not using "bootstrap" we can remove it from configuration and the import and make an alias path to node_module. (but in my limited knowledge, vite.config.ts probably already has a "default" one to node_module. You will need to check if you like to change it.

frontend/src/components/common/PageTitle.vue
```diff
@@ -1,68 +1,35 @@
...
+        type: String,
+        required: true,
+    },
+});
 </script>
 
 <template>
-    <div class="title" v-if="props.displaySelectedApplication">
-        {{ selectedApplicationDisplayText }}
+    <div class="row">
```
MCatherine1994, line 15: why we need this classname row?

ianliuwk1019:
> I got this error in the console after I logged in
> 
> image

I think I also hit this error... but not too sure, I quickly turns to other action so didn't look into it.

ianliuwk1019 (changes requested):

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                class="dashboard-button"
+                label="Grant new access"
+                @click="router.push('/grant')"
+                ><Icon icon="AddIcon" medium class="icon-size-1"
+            /></Button>
+        </div>
+    </div>
+    <div class="page-body">
+        <div class="row">
+            <div class="col-12">
```
MCatherine1994, line 166: why we need these class row and col-12 here?

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                                :options="filteredOptions"
+                                optionLabel="application_description"
+                                placeholder="Choose an option"
+                                class="application-dropdown"
+                            />
+                        </div>
+                    </div>
+                </form>
+            </div>
+            <div class="col-12">
```
MCatherine1994, line 189: this as well, not sure if we need the classname

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                    v-if="useNotificationMessage.isNotificationVisible"
+                    severity="success"
+                    :msgText="useNotificationMessage.notificationMsg"
+                    icon="CheckIcon"
+                />
+            </div>
+        </div>
+    </div>
+    <div class="row h-auto" v-if="isApplicationSelected">
+        <div class="col-12">
```
MCatherine1994, line 200: and here, do we need the classname row, h-auto and col-12

MCatherine1994:
I'm actually don't like the big space here, maybe there is a reason I don't know. But we could chat about this and work for a bette dynamic solution to leave the space for messages. thanks!! 

image

ianliuwk1019:
I am not sure if it is just on my browser display or else.... first impression on viewing the app dashboard the first time locally, I suddenly feel the **font** on **label** seems quite small and lighter that I might need to pay more attention to read.
Maybe due to the **BC San font** default is like that? Should we make it slightly larger or darker? Not quite sure where is that's the issue for others.
Just a feedback, probably was design that way...

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                <div class="p-table-header">
+                    <h3>{{ selectedApplicationDisplayText }} users</h3>
+                    <span>
+                        This table shows all the users in
+                        {{ selectedApplicationDisplayText }} and their
+                        permissions levels
+                    </span>
+                </div>
+
+                <span class="p-input-icon-right">
```
MCatherine1994, line 211: this search input border is overlayed by the data table header, add the z-index to this class so it knows which one to show top. So I believe the style for this class is at line 289

.p-input-icon-right {
    ....
    z-index: 1;
    ....
}

image

frontend/src/components/GrantAccess.vue
MCatherine1994: think I'll create another ticket for this, don't want to add more work to this ticket, and it's not urgent

MCatherine1994:
Just for best practice and clean code, when we write import and define variables, let's group them in order.

For example, for import statements:
- import from vue for any package goes first
- import component from the ui library goes next (primevue component first, function next)
- import component from our local folder goes next
- then we import any local service call, api, state
- then we import type

When we define variables, we group them a bit. For example, let's put confirm and confirmRemoveMessage together so we know they are related. Put variables for api service and user role assignment ones together, so we know they are related as well. Otherwise if put them whatever, it's hard for us to review and understand the code in the future. 

Screen Shot 2023-07-20 at 9 07 08 PM

MCatherine1994:
I got type warning on this one
image

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+        return Promise.reject(error);
+    } finally {
+        loading.value = false;
+    }
+
+    if (isApplicationSelected) {
+        getAccessList();
+    }
+});
+const filteredOptions = computed(() => {
```
MCatherine1994, line 68: Let's rename it to the selectApplicationFilteredOptions, it confused me a bit if it's for the dropdown or for the data table

MCatherine1994:
Thinking of consistency, maybe let's have <template> at the top and then the <script setup>. These following components need to be adjusted:
- Button
- NotificationMessage
- PageTitle

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                                optionLabel="application_description"
+                                placeholder="Choose an option"
+                                class="application-dropdown"
+                            />
+                        </div>
+                    </div>
+                </form>
+            </div>
+            <div class="col-12">
+                <NotificationMessage
```
MCatherine1994, line 190: We use this NotificationMessage component, but we didn't import it.

MCatherine1994:
wondering if we remove the bottom border of the last row on purpose? 
image

MCatherine1994:
When no user role assignments found for a selected application, this error message is not correct.

image

MCatherine1994:
let's merge the main branch as well to keep up to date

MCatherine1994:
In the case I couldn't verify a forest client id, should submit button be disabled?
Screen Shot 2023-07-20 at 10 56 51 PM

MCatherine1994:
Looks like in the code we don't support search by identity type. We could consider support this search or update the displayed message

Screen Shot 2023-07-20 at 11 03 46 PM

MCatherine1994:
Wondering when we click on the data table header for sorting, the column size will change...

MCatherine1994:
Wondering when we assign a duplicate role, shall we get "Role already assigned to user." error message? This one looks a little bit general.

Screen Shot 2023-07-20 at 11 09 47 PM

MCatherine1994:
I assume we'll always get a default select for the user's domain, but sometimes I get this situation

image

frontend/src/components/common/cards/ForestClientCard.vue
NickSaglioni: Thanks, I'll remove it.

frontend/src/components/common/Sidebar.vue
NickSaglioni: Ok, this will be renamed.

frontend/src/components/dialog/Dialog.vue
NickSaglioni: Thanks, I'll move it to common folder.

NickSaglioni:
> um.... I'm actually don't like the big space here, but I know it's in the design system. Maybe we could confirm with Olga. thanks!
> 
> image

Would you prefer to have the table right below the dropdown? Actually we need to have some space there for the notifications to appear.

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                    <Column
+                        field="role.client_number.forest_client_number"
+                        header="Forest Client ID"
+                        sortable
+                    ></Column>
+                    <Column>
+                        <template #body="{ data }">
+                            <button
+                                class="btn btn-icon"
+                                @click="tryDelete(data)"
```
ianliuwk1019, line 272: Since this button is intended for deletion, I prefer not to prefix it with "try".
Maybe give it some context for the name then just "delete"; like "deleteUserAssignment" or something similiar.

MCatherine1994:
We need to add the following to the `frontend/src/assets/styles/styles.scss` file, to override the font family for component

.p-component  {
    font-family: map.get(typeFamily.$type-family, 'sans');
}

ianliuwk1019 (changes requested):

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+        });
+    } catch (error: unknown) {
+        router.push('/dashboard');
+        return Promise.reject(error);
+    } finally {
+        loading.value = false;
+    }
+}
+
+async function tryDelete(assignment: FamApplicationUserRoleAssignmentGet) {
```
ianliuwk1019, line 105: Need a better name.

frontend/src/store/ApplicationState.ts
```diff
@@ -37,3 +39,12 @@ export const selectedApplicationDisplayText = computed(() => {
...
 });
+
+export const useNotificationMessage = reactive({
+    notificationMsg: '',
+    isNotificationVisible: false,
+})
+
+export const useErrorDialog = reactive({
+    isErrorVisible: false,
+})
```
ianliuwk1019, line 50: Hmm... I am confused why "grantAccessFormData", "useNotificationMessage" and "useErrorDialog" belong to this "ApplicationState.ts" file as originally it seems to be holding state for "Application" but Notification and Dialog different. Can they be in different "state" file or is it better to design/rename the filename to be more generic?

ianliuwk1019 (changes requested):

frontend/src/components/GrantAccess.vue
```diff
@@ -119,211 +131,298 @@ async function getForestClientNumber(forestClientNumber: string) {
...
-                        id="grantAccessSubmit"
-                        class="btn btn-primary mb-3"
-                        :disabled="!meta.valid || invalidForestClient"
-                    >
-                        Grant Access
-                    </button>
-                </div>
+                    <div class="row gy-3 my-0" v-if="meta.valid">
+                        <div class="col-auto px-0">
+                            <Button
```
ianliuwk1019, line 399: It seems to me, when user enters "forestClientNumber" (it may be correct, or it may be incorrect number) and the "Verify" button help to verify the correctness. However, as soon as I enter 8 digits, this "Submit" button is active and can go to summary page (without user verifies the forestClientNumber).
Has this behavior being communicated to the team or should it be correct forestClientNumber before user can go next? 
Basically, this becomes a typical "Form Validation" question.

frontend/src/components/common/cards/SummaryCard.vue
```diff
@@ -0,0 +1,81 @@
...
+                </template>
+                <template #footer>
+                    <div class="card-footer">
+                        <Button
+                            type="submit"
+                            id="grantAccessSubmit"
+                            class="mb-3"
+                            label="Submit"
+                            v-on:click="$emit('submit')"
+                        ></Button>
```
ianliuwk1019, line 56: When I submit, it takes few seconds (but it is normal that response is slow).
However, there is no loading or some sort of overlay while waiting for its submit response.
User is puzzling that few seconds and can "**submit multiple times**" to confuse both backend/frontend.

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+    confirmRemoveMessage.userName = assignment.user.user_name;
+    useNotificationMessage.isNotificationVisible = false;
+    useNotificationMessage.notificationMsg = '';
+    confirm.require({
+        group: 'templating',
+        icon: 'none',
+        header: 'Remove Access',
+        rejectLabel: 'Remove',
+        acceptLabel: 'Cancel',
+        reject: () => {
```
ianliuwk1019, line 116: I thought this Confirm dialog is to "accept" the confirmation to remove access.
Why is the label for user to "accept" his action is labeling "Cancel" (at the "acceptLabel")?
Same as the callback function "reject: () =>" when user reject his action to remove access, the callback actually trying to remove the access. Isn't that contradict what user wants?
And that will make other developer to wrap the head around why the function/label are doing the opposite?

ianliuwk1019 (changes requested):

frontend/src/components/GrantAccess.vue
```diff
@@ -1,34 +1,40 @@
...
 import router from '@/router';
 import { number, object, string } from 'yup';
+import Button from 'primevue/button';
+import InputText from 'primevue/inputtext';
+import Dropdown, { type DropdownChangeEvent } from 'primevue/dropdown';
+import RadioButton from 'primevue/radiobutton';
 
 const FOREST_CLIENT_INPUT_MAX_LENGTH = 8;
 const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend when backend has the endpoint.
 let applicationRoleOptions = ref<FamApplicationRole[]>([]);
```
MCatherine1994, line 24: these state variables we can just use const to define

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                    />
+                </span>
+
+                <DataTable
+                    v-model:filters="filters"
+                    :value="userRoleAssignments"
+                    paginator
+                    :rows="5"
+                    :rowsPerPageOptions="[5, 10, 15, 20, 50, 100]"
+                    dataKey="id"
```
ianliuwk1019, line 225: Do we have "id" in the record for dataKey?

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+                        'role.parent_role.role_name',
+                        'role.role_name',
+                        'role.client_number.forest_client_number',
+                    ]"
+                    paginatorTemplate="RowsPerPageDropdown CurrentPageReport PrevPageLink NextPageLink"
+                    currentPageReportTemplate="{currentPage} of {totalPages} pages"
+                >
+                    <template #empty> No application selected. </template>
+                    <template #loading>
+                        Loading customers data. Please wait.
```
ianliuwk1019, line 239: Not sure if "customers" is good here than "users"

frontend/src/components/GrantAccess.vue
MCatherine1994, line 24: state variable can be defined as const

ianliuwk1019 (changes requested):

frontend/src/components/GrantAccess.vue
```diff
@@ -119,211 +131,298 @@ async function getForestClientNumber(forestClientNumber: string) {
...
+                                }}<span class="text-danger"> *</span></label
+                            >
+                            <Field
+                                id="roleSelect"
+                                class="form-select"
+                                name="role_id"
+                                aria-label="Role Select"
+                                v-slot="{ field, handleChange }"
+                                v-model="formData.role_id"
+                                @change="onRoleSelected"
```
MCatherine1994, line 302: I wonder if this method actually will be called

frontend/src/components/GrantAccess.vue
```diff
@@ -1,34 +1,40 @@
...
+import InputText from 'primevue/inputtext';
+import Dropdown, { type DropdownChangeEvent } from 'primevue/dropdown';
+import RadioButton from 'primevue/radiobutton';
 
 const FOREST_CLIENT_INPUT_MAX_LENGTH = 8;
 const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend when backend has the endpoint.
 let applicationRoleOptions = ref<FamApplicationRole[]>([]);
 let forestClient: FamForestClient[] | null = null;
 let loading = ref<boolean>(false);
+let userLoaded = ref<boolean>(false);
```
MCatherine1994, line 27: do we need this variable? or we can just check if formData.userId has a value to replace this variable?

frontend/src/assets/styles/layout.scss
```diff
@@ -1,13 +1,26 @@
...
+    display: block;
+    top: $header-height;
+    left: 250px;
+    right: 0px;
+    bottom: 0px;
+    overflow: auto;
+    padding-top: 42px;
+    padding-left: 40px;
+    padding-right: 30px;
+}
```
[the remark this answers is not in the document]
NickSaglioni: This is a really good point and you are right. We might have been misguided by Figma, I'll switch everything to relative. Thanks.

frontend/src/components/Dashboard.vue
NickSaglioni: This name is coming from what we had from ManageAccess.vue. You are right about the naming and I'll be changing it. Thanks.

frontend/src/components/Dashboard.vue
```diff
@@ -0,0 +1,321 @@
...
+        matchMode: FilterMatchMode.CONTAINS,
+    },
+    'role.client_number.forest_client_number': {
+        value: null,
+        matchMode: FilterMatchMode.CONTAINS,
+    },
+});
+const userRoleAssignments = ref<FamApplicationUserRoleAssignmentGet[]>();
+
+const confirmRemoveMessage = reactive({
```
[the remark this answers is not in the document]
NickSaglioni: It doesn't need to be reactive, thanks for spotting it.

frontend/src/components/Dashboard.vue
NickSaglioni: It will be changed. Thanks.

frontend/src/components/Dashboard.vue
NickSaglioni: According to @J0taFerreira this is because the PrimeVue component for the dialog has the button layout inverted than what we need. One idea could be to create a middle component that inverts it, making it more clear to who is using it. What do you think?

ianliuwk1019 (changes requested):

frontend/src/components/common/cards/SummaryCard.vue
ianliuwk1019, line 56: Does our design of SummaryCard needs to include "submit" button?
Also, this SummaryCard looks to me is quite specific for displaying ONLY for grantAccess data and the button id even says "grantAccessSubmit". Initially I thought "Summary"Card, is a "common" component that it may be used in different "summary"....

ianliuwk1019 (changes requested):

frontend/src/components/Summary.vue
```diff
@@ -0,0 +1,59 @@
...
+</script>
+
+<template>
+    <Dialog
+        v-model:visible="useErrorDialog.isErrorVisible"
+        :error="true"
+        header="Error"
+        text-first="This role cannot be assigned to this user."
+        text-second="Contact your administrator for more information."
+    ></Dialog>
```
ianliuwk1019, line 45: The dialog showing on my screen seems to be too big and lots of empty space.

frontend/src/components/Summary.vue
```diff
@@ -0,0 +1,59 @@
...
+
+    <div class="page-body">
+        <div class="row">
+            <SummaryCard
+                :data="(grantAccessFormData as FamUserRoleAssignmentCreate)"
+                @submit="handleSubmit()"
+            />
+        </div>
+    </div>
+</template>
```
ianliuwk1019, line 59: If I refresh the browser on this page, I will get:
`Error occurred: TypeError: Cannot read properties of undefined (reading 'user_name')`

ianliuwk1019 (changes requested):

frontend/src/assets/styles/layout.scss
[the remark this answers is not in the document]
MCatherine1994: Let's create another ticket to change all absolute px to relative rem. Figma dev mode support change px to rem.
image

frontend/src/components/Dashboard.vue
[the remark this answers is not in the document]
MCatherine1994: we'll keep it reactive for now to make sure we always get update when value changes

frontend/src/components/GrantAccess.vue
```diff
@@ -119,211 +131,298 @@ async function getForestClientNumber(forestClientNumber: string) {
...
+                                </Field>
+                                <ErrorMessage
+                                    class="invalid-feedback"
+                                    name="userId"
+                                />
+                            </div>
+                            <div class="col-md-3" v-if="!userLoaded">
+                                <Button
+                                    class="button p-button-tertiary p-button-outlined"
+                                    @click="getUserId('4')"
```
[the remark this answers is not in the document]
MCatherine1994: yeah we can just comment out this for now

frontend/src/components/GrantAccess.vue
```diff
@@ -119,211 +131,298 @@ async function getForestClientNumber(forestClientNumber: string) {
...
+                                        ? 'to ' + formData.userId
+                                        : ''
+                                }}<span class="text-danger"> *</span></label
+                            >
+                            <Field
+                                id="roleSelect"
+                                class="form-select"
+                                name="role_id"
+                                aria-label="Role Select"
+                                v-slot="{ field, handleChange }"
```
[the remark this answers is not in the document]
MCatherine1994: think it's the default method from the vee-form

frontend/src/components/common/ProfileSidebar.vue
```diff
@@ -0,0 +1,183 @@
...
+            <div class="sidebar-body">
+                <div class="img-wrapper">
+                    <img
+                        src="../../assets/images/tyrannosaurus-rex1.png"
+                        alt="User avatar"
+                    />
+                </div>
+                <div class="profile-info">
+                    <p class="profile-name">
+                        {{ userProfileInfo[0].userInfo.name }}
```
ianliuwk1019, line 42: User can be only one login user at a time I assume, so not sure why "userProfileInfo" is an array.

frontend/src/static/userProfileInfo.json
```diff
@@ -0,0 +1,11 @@
...
+    { 
+        "idir" : "oliber",
+        "userInfo" :  
+            {
+                "name" : "Olga Liber",
+                "avatar" : "../assets/images/tyrannosaurus-rex1.png",
+                "email" : "olga.liberchuk@bc.gov.ca"
+            }
+    }
+]
```
ianliuwk1019, line 11: We are not hardcoding a user to display.
FAM is already working for user to login. This file should be deleted.

ianliuwk1019 (changes requested):

frontend/src/components/common/Sidebar.vue
```diff
@@ -0,0 +1,90 @@
...
+    name: string;
+    items: [ISideBarItem];
+}
+
+interface ISideBarItem {
+    name: string;
+    icon: string;
+    link: RouteLocationRaw;
+    disabled: boolean;
+}
```
ianliuwk1019, line 77: Since you are renaming SideBar to SideNav, please don't forget to rename ISideBarData and ISideBarItem to be consistent.

frontend/src/static/sidebar.json
```diff
@@ -0,0 +1,30 @@
...
+            }
+        ]
+    },
+    {
+        "name": "Management",
+        "items": [
+            {
+                "name": "Settings",
+                "icon": "SettingsIcon",
+                "link": "#",
```
ianliuwk1019, line 25: This **"Management"** section does not seem to be implemented.
Should comment it out or remove it, unless it's been discussed to leave it there.

frontend/src/services/ErrorService.ts
```diff
@@ -1,39 +0,0 @@
...
-    if (axios.isAxiosError(error)) {
-        const err = error as AxiosError;
-        const axiosResponse = err.response;
-        const status = axiosResponse?.status;
-
-        const e401_authenticationErrorMsg =
-            'You are not logged in. Please log in.';
-        const e403_authorizationErrorMsg =
-            'You do not have the necessary authorization for the requested action.';
-        const e409_conflictErrorMsg = `${axiosResponse?.data.detail}`;
```
ianliuwk1019, line 20: Maybe previous **ErrorService** needs to be improved, but I wonder, if this ErrorService is now removed and in **main.ts** you also removed "**app.config.errorHandler**", how does the system now react to generic errors like above 401/403/409 etc? Does it mean, each component that connect to backend will need to handle them specifically or the system will have uncaught errors on console and won't show messages to the users?

frontend/src/main.ts
```diff
@@ -25,71 +16,26 @@ import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome';
...
-    vfmPlugin({
-        key: '$vfm',
-        componentName: 'VueFinalModal',
-        dynamicContainerName: 'ModalsContainer',
-    })
-);
-
-app.config.errorHandler = (err, instance, info) => {
-    ErrorService.onError(err, info);
-};
```
ianliuwk1019, line 91: See comment in deleted ErrorService.

ianliuwk1019 (changes requested):

frontend/src/services/ErrorService.ts
ianliuwk1019: Sorry, now I just saw you have a new "common/ToastMessage.vue" as an component.
However..... does it really work? I don't seem to see message....or maybe it does?

frontend/src/components/common/ToastMessage.vue
```diff
@@ -0,0 +1,82 @@
...
+        }
+        return;
+    }
+
+    showToastTopRight('error', genericErrorMsg.title, genericErrorMsg.text);
+};
+
+app.config.errorHandler = (err, instance, info) => {
+    onError(err, info);
+};
```
ianliuwk1019, line 71: Not really sure if messages will really show up using ToastMessage.vue....
Could you confirm those 4xx errors you coded above do work?
I kind of feel, maybe I am wrong, some configuration need to be registered as soon as possible in app before component is even initialized. Hope I am wrong, or this will break the behavior.

ianliuwk1019 (changes requested):

NickSaglioni:
> I got this error in the console after I logged in
> 
> image

This is fixed.

frontend/src/components/common/ProfileSidebar.vue
```diff
@@ -0,0 +1,183 @@
+<script setup lang="ts">
+import authService from '@/services/AuthService';
+import { useProfileSidebarVisible } from '../../store/useProfileVisibleStore';
+import userProfileInfo from '../../static/userProfileInfo.json';
```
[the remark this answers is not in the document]
MCatherine1994: I created another ticket for this profile improvement to get real user data. So far, let's just leave the sign out button there, and don't display any hardcode information.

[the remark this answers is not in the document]
ianliuwk1019: Agreed, we can remove hardcoded .json file and comment out the rest to only leave sign out there.

NickSaglioni:
> In the case I couldn't verify a forest client id, should submit button be disabled?
> Screen Shot 2023-07-20 at 10 56 51 PM

This has been fixed.

NickSaglioni:
> I got type warning on this one
> image

Replacing it back to .sort gives one code smell in Sonar Cloud.

ianliuwk1019 (changes requested):

server/backend/local-dev.env
```diff
@@ -9,5 +9,5 @@ COGNITO_CLIENT_ID=6jfveou69mgford233or30hmta
 COGNITO_USER_POOL_DOMAIN=dev-fam-user-pool-domain
 ENABLE_BCSC_JWKS_ENDPOINT=False
 BCSC_KEY_ID=thisisasecret
-FC_API_TOKEN=thisisasecret
-DB_STRING_OVERRIDE=postgresql+psycopg2://fam_proxy_api:test@localhost:5432/fam
\ No newline at end of file
+FC_API_TOKEN=4r0dn5wC0JTs6N1tdw6NDzpjBRQwEJ6u
```
ianliuwk1019, line 12: This should not be committed.

ianliuwk1019 (changes requested):

ianliuwk1019 (changes requested):

ianliuwk1019:
This part of dashboard is broken for user:
(When user did not select application but clicked Grant Access)
image

ianliuwk1019:
This is also broken for the user (See my comment about 401/403/409 that I felt Toast Message Component might not be working for previous all cases):
This is the case for http status 403 (Unauthorized operation).
image

NickSaglioni:
> This is also broken for the user (See my comment about 401/403/409 that I felt Toast Message Component might not be working for previous all cases):
> This is the case for http status 403 (Unauthorized operation). 401 Error for Toast message is also broken but is a bit trickier to find.
> image

@J0taFerreira could you please investigate this and provide some screenshots in case it is working as expected? Thank you.

MCatherine1994:
Get this warning when grant a user access
Screen Shot 2023-07-31 at 12 49 22 PM

MCatherine1994:
I guess this is the error 409 case, when the user already have the role, we need to tell the correct error message
Screen Shot 2023-07-31 at 12 53 22 PM
image

MCatherine1994:
feel this is not aligned... let's do a quick fix for this for now, and we could worry about the style later
```
/*----------- data table pagnition ---------------*/
.p-datatable .p-paginator-bottom .p-dropdown .p-dropdown-label.p-inputtext {
    padding-top: 7px;
}
```

Screen Shot 2023-07-31 at 1 12 12 PM

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -0,0 +1,402 @@
...
+    grantAccessFormData,
+    domainOptions,
+    FOREST_CLIENT_INPUT_MAX_LENGTH,
+    setGrantAccessFormData,
+    getGrantAccessFormData,
+} from '@/store/GrantAccessDataState';
+
+import type { FamApplicationRole, FamForestClient } from 'fam-api';
+
+interface IFormData {
```
MCatherine1994, line 28: why we need a new interface type? is it different than this type "import type { FamUserRoleAssignmentCreate } from 'fam-api'" in the GrantAccessDataState.ts file?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -0,0 +1,397 @@
...
+            then: () =>
+                string()
+                    .required('Forest Client number is required')
+                    .min(8, 'Forest Client ID must be 8 characters long'),
+        })
+        .nullable(),
+});
+
+const loading = ref<boolean>(false);
+const userLoaded = ref<boolean>(false);
```
MCatherine1994, line 53: do we use this variable somewhere? maybe remove it if it's not in use

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -0,0 +1,397 @@
...
+
+import type { FamApplicationRole, FamForestClient } from 'fam-api';
+
+const defaultFormData = {
+    domain: domainOptions.IDIR,
+    userId: '',
+    forestClientNumber: '',
+    role_id: null as number | null,
+};
+const formData = ref(defaultFormData); // clone default input
```
MCatherine1994, line 34: this need to be a hard copy, just copy the value,  "ref(JSON.parse(JSON.stringify(defaultFormData)))", otherwise when you reset the form in the "cancelForm()" method, it won't get reset

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -0,0 +1,393 @@
...
+    invalidForestClient.value = true;
+    forestClientData.value = null;
+}
+
+function resetForm() {
+    formData.value = defaultFormData;
+}
+
+function cancelForm() {
+    resetForm();
```
MCatherine1994, line 112: this still doesn't look correct to me..... when we click on the cancel, do we want to reset the form to be empty? then in this case, we might need to reset the grantAccessFormData in the store as well, otherwise whenever I click grant access again, it will populate the stored grantAccessFormData, then the reset here doesn't make any sense.

MCatherine1994:
this color doesn't look right....
image

MCatherine1994 (approved):

ianliuwk1019:
> > > This part of dashboard is broken for user:
> > > (When user did not select application but clicked Grant Access)
> > > image
> > 
> > 
> > This will need further investigation because it might need to refactor the sideNav component to support disabling links after component loading.
> 
> I'll create a ticket for this one, and Nick you can confirm with Olga to see if want to disable the grant access button when there is no application selected. #733

Thank you for the ticket Catherine! This will need to be fixed as soon as we can (as mentioned on the ticket before next release). Thanks.

ianliuwk1019:
> This is also broken for the user (See my comment about 401/403/409 that I felt Toast Message Component might not be working for previous all cases): This is the case for http status 403 (Unauthorized operation). 401 Error for Toast message is also broken but is a bit trickier to find. image

The 401/403 errors are still there as I can reproduced.
I discussed with @MCatherine1994  and created this ticket: #735.

ianliuwk1019:
Catherine has approved it so no need for my approval but did verify few critical ones.
I also chatted with Catherine that I also feel maybe it is better to let this PR go but our dev environment will not be in a good standing for the FAM frontend to be released. We should address several bugs as soon as we can, deploy to dev after that and let Olga or others to test and find bugs until the level we feel it is good for users. @gormless87

ianliuwk1019 (approved):

PR #723, opened 2023-07-26
feat: 575 implement bcgovpubcode

basilv:


bcgovpubcode.yml
```diff
@@ -0,0 +1,50 @@
+---
+data_management_roles:
+  data_custodian: Chantelle Abanilla
+  data_stewart: Pete Tollestrup
```
gormless87, line 4: data_steward:

gormless87 (changes requested):
Just one change from data_stewart to data_steward

gormless87 (approved):

Issue #679, opened 2023-06-29
Integrate fam-idim-lookup-proxy with FAM for IDIR verfification


gormless87:
Description:
Implement verify function for IDIR in FAM

AC:
- Can search IDIR user by username to verify for IDIR (backend - new API endpoint)
- if found: display a green checkmark (frontend)
- if not found:  display error message (frontend)

Additional context:
User card component implementation to follow in separate issue

ConradBoydElliottGustafson:
Confirmed that only Active IDIR accounts will return a result from the web service. As per Pathfinder SSO team.

ianliuwk1019:
For current design of IDIM-Proxy, it only has an already hashed string as an api-key in the header ("X-API-KEY") needed.
The value need to be set in both repos.

ianliuwk1019:
The "**requester**" (the user who login in FAM) is needed for backend to send a request to **IDIM-Lookup-Proxy**. FAM currently does not have direct information about this requester from **Cognito Access Token**. The access token does have "**username**" attribute (which mapped to "**cognito_user_id**" in **fam_user** database with this similar value: "**idir_b5ecdb094dfb4149a6a8445a01a96bf0@idir**" (The value gets inserted when user login).

While doing the integration, I found we can lookup the **requester** information by its "**cognito_user_id**" from the database. This has no problem in deployment environments, but locally will face problem because when user sign in at FAM frontend, the sign in user's "cognito_user_id" is saved in AWS DEV environment database, **NOT** on developer's local database. Developers will have problem doing user IDIR check (either from frontend FAM or swagger) because backend can't find who the requester is.

Few solutions are possible:
1. Passing entire token (contains not only access token, but id token) from frontend to our FAM resource server (backend). For this point, need to investigate how to adjust current backend jwt_validation code to take into account of the token structure change. (and possibly how it affect swagger UI)
2. Backend (somehow) exchange Access Token with AWS Cognito for ID Token. For this, more on how to integration with Cognito is required at jwt_validation code. Also for each request, it will need to call Cognito (maybe expensive/slow) and what if the service somehow isn't reachable.
3. Adjust token mapping on Access Token to provide necessary login user information. For this, not sure if possible, or by convention people do that but would possibly the easiest.

ianliuwk1019:
Just a experimenting to try passing a user GUID instead of username to see if IDIM webservice work but get 400.
[...]

PR #728, opened 2023-07-28
feat: #679 integrate idim lookup proxy

ianliuwk1019:
This ticket is for #679 for FAM to integrate with Fam-IDIM-Lookup-Proxy in Openshift and be able to search/verify IDIR User at FAM frontend/backend.

**Backend**:
* Add Idim-Proxy service to request IDIR serach to the proxy.
* Add router: router_idim_proxy, and add new enpoint "/idir" to search idir user.
* Add specific requests_http_error_handler for FastAPI so it does not get confused from exception back from request and throw 500 error.
* Refactor "jwt_validation.get_request_username" to be more clear and add comment.
* Add a new "Requester" class to retrieve the requester/user as a dependency for new "/idir" endpoint
	so it validates and able to use the requester for idim-proxy lookup.
* Add local requester data for local environment only to prevent development problem when lookup proxy without requester info in database.
* Configure idim-proxy base-url and api-key for github-workflow/terraform deployment.
* Add idim-proxy api key to github secret.
* Generate new api-client for frontend.
* Include generated IDIRBCeIDProxyApi to frontend ApiServiceFactory.
* Add service tests and router tests.

**Frontend**:
* Added search IDIR functionality for GrantAccess.
* Added UserIdentityCard to display searched result.
* Reset userId/verifiedUserIdentity when user domain changed.
* Next button function renamed.
* Next button condiont adjustment: shows only when userId/forestClientNumber both verified.
* Remove unnecessary reactive declaration and use normal typescript declaration.


server/backend/api/app/integration/idim_proxy.py
```diff
@@ -0,0 +1,61 @@
...
+
+REQUESTER_ACCOUNT_TYPE_INTERNAL = "Internal"
+
+class IdimProxyService():
+    """
+    The class is used for making requests to search IDIR/BCeID information from IDIM Proxy API.
+    See environment setup (local-dev.env) for idim-proxy TEST api-docs.
+    """
+    TIMEOUT = (5, 20) # Timeout (connect, read) in seconds.
+    RETRY = 3
```
basilv, line 21: I think this warrants discussion. Retrying is problematic as it delays the response to the user. Is this recommended by the IDIM guide?

server/backend/api/app/jwt_validation.py
```diff
@@ -238,7 +238,11 @@ def get_access_roles(claims: dict = Depends(authorize)):
     return groups
 
 
-def get_request_username(claims: dict = Depends(authorize)):
-    requester = claims[COGNITO_USERNAME_KEY]
-    LOGGER.debug(f"Current requester for API: {requester}")
-    return requester
+def get_request_cognito_username(claims: dict = Depends(authorize)):
```
basilv, line 241: Why not call this cognito_user_id to avoid confusion.

server/backend/api/app/requester.py
```diff
@@ -0,0 +1,78 @@
...
+    # cognito_user_id => Cognito OIDC access token maps this to: username (ID token => "custom:idp_name" )
+    cognito_user_id: Union[str, None]
+    user_name: str
+    # "B"(BCeID) or "I"(IDIR). It is the IDP provider.
+    user_type: Union[str, None]
+    access_roles: Union[List[str], None]
+
+
+async def get_current_requester(
+    request_cognito_username: str = Depends(jwt_validation.get_request_cognito_username),
```
basilv, line 49: yes, this would definitely be clearer if it was called cognito_user_id

server/backend/api/app/routers/router_idim_proxy.py
```diff
@@ -0,0 +1,28 @@
...
+
+ERROR_EXTERNAL_USER_ACTION_PROHIBITED = "external_user_action_prohibited"
+
+LOGGER = logging.getLogger(__name__)
+
+router = APIRouter()
+
+@router.get("/idir", response_model=dict, dependencies=[Depends(internal_only_action)])
+def idir_search(
+    user_id: str,
```
basilv, line 18: probably clearer to call this something like 'search_user_id", to make it clear this is the user id we are searching for / looking up, instead of the current user.

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -19,7 +19,7 @@ def create_user_role_assignment(
     role_assignment_request: schemas.FamUserRoleAssignmentCreate,
     db: Session = Depends(database.get_db),
     token_claims: dict = Depends(jwt_validation.authorize),
-    requester: str = Depends(jwt_validation.get_request_username),
+    cognito_username: str = Depends(jwt_validation.get_request_cognito_username),
```
basilv, line 22: This is likely more accurately called cognito_user_id

server/backend/local-data.json
```diff
@@ -0,0 +1,9 @@
+{
+    "requester":
+    {
+        "cognito_user_id": "dev-idir_e72a12c916a44a9581cf39e5dcdffae7@idir",
```
basilv, line 4: This will have to be updated each time the dev environment is recreated.

server/backend/local-dev.env
```diff
@@ -10,4 +10,6 @@ COGNITO_USER_POOL_DOMAIN=dev-fam-user-pool-domain
 ENABLE_BCSC_JWKS_ENDPOINT=False
 BCSC_KEY_ID=thisisasecret
 FC_API_TOKEN=thisisasecret
-DB_STRING_OVERRIDE=postgresql+psycopg2://fam_proxy_api:test@localhost:5432/fam
\ No newline at end of file
+DB_STRING_OVERRIDE=postgresql+psycopg2://fam_proxy_api:test@localhost:5432/fam
+IDIM_PROXY_BASE_URL=https://nr-fam-idim-lookup-proxy-test-backend.apps.silver.devops.gov.bc.ca
+IDIM_PROXY_API_KEY=thisisasecret
```
basilv, line 15: We probably need to update a developer readme for instructions on how to set up their local environment from scratch.

basilv (changes requested):

.github/workflows/sonar.yml
```diff
@@ -69,6 +69,7 @@ jobs:
       - name: Tests and coverage
         env:
           FC_API_TOKEN: ${{ secrets.FOREST_CLIENT_API_API_KEY }}
+          IDIM_PROXY_API_KEY: ${{ secrets.IDIM_PROXY_API_API_KEY }}
```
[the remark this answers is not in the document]
ianliuwk1019: The Python pytests suites require these environment variables to run successfully. The env isn't for sonar and coverage (probably the workflow name is confusing).

server/backend/api/app/integration/forest_client/forest_client.py
```diff
@@ -18,7 +18,7 @@ class ForestClientService():
     """
     # https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
     # https://docs.python-requests.org/en/latest/user/advanced/#timeouts
-    TIMEOUT = (4, 10) # Timeout (connect, read) in seconds.
+    TIMEOUT = (5, 15) # Timeout (connect, read) in seconds.
```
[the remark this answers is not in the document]
ianliuwk1019: You are right, I felt its a bit long too.
Forest Client api seems a bit slow. Occasionally, I see forest-client-api requests were timeout (both frontend search and unit tests) but didn't investigate much. I don't know if it is just forest-client api service needs some time to wake up or something else. The only reason I increase this much is to hope the request goes smoothly. Frontend also has a timeout value (currently default to 20 seconds).

I search for typical timeout values, ours could still be in reasonable range, but I can change it back to previous.
Python:
image
.Net:
image

server/backend/api/app/integration/idim_proxy.py
ianliuwk1019: Ha, no, it is not from IDIM guide recommendation. Same thought as above and I was trying to be on the safe side just to add a retry since it is easy to add in Python. However, since you point out it "delays" to the user and our frontend probably didn't take into this consideration, maybe I should remove the retry. Can we discuss if there is a value for it when you have time? Thanks.

server/backend/api/app/jwt_validation.py
ianliuwk1019: :) Thanks, I actually prefer and will change it to be called "**cognito_user_id**".
It was only because the fact its name on Cognito attribute mapping is "**username**" so I refactored it from previous method name "get_request_username" to be more meaningful, but do prefer to call it "cognito_user_id".
Will change.
image

ianliuwk1019: Changed (get_request_cognito_user_id)

server/backend/api/app/routers/router_user_role_assignment.py
ianliuwk1019: Changed (get_request_cognito_user_id)

server/backend/api/app/routers/router_idim_proxy.py
ianliuwk1019: IDIM webservice is actually capable for searching not only userId, also username, lastname, firstname. However, our proxy currently only coded as userId only search to IDIM service.
For this router, the "user_id" in the argument actually will be translate to a "request parameter" not a "path parameter". So it will be "/search?user_id=ianliu" not "/search/user-id". It leaves more room to be expend in future if necessary without needing to refactor the name again.
But we can discuss.

ianliuwk1019: Discussed. Mark it resolved.

server/backend/api/app/integration/forest_client/forest_client.py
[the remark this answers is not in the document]
ianliuwk1019: Discussed. For now we lower the timeout (5,10) to 10 seconds the most. (lambda also has timeout of 15 seconds). If external service does give problem often, we will blame to the service or find alternative for user be aware (maybe frontend).

server/backend/api/app/integration/idim_proxy.py
```diff
@@ -0,0 +1,61 @@
...
+LOGGER = logging.getLogger(__name__)
+
+REQUESTER_ACCOUNT_TYPE_INTERNAL = "Internal"
+
+class IdimProxyService():
+    """
+    The class is used for making requests to search IDIR/BCeID information from IDIM Proxy API.
+    See environment setup (local-dev.env) for idim-proxy TEST api-docs.
+    """
+    TIMEOUT = (5, 20) # Timeout (connect, read) in seconds.
```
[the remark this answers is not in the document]
ianliuwk1019: Discussed. Same as forest-client-api service timeout.
For now we lower the timeout (5,10) to 10 seconds the most. (lambda also has timeout of 15 seconds). If external service does give problem often, we will blame to the service or find alternative for user be aware (maybe frontend).

server/backend/api/app/integration/idim_proxy.py
ianliuwk1019: Discussed and remove this. At this case we don't want to retry for user initiated request.

server/backend/local-dev.env
ianliuwk1019: New instruction added in Readme. Please see [here](https://github.com/bcgov/nr-forests-access-management/pull/728/files#diff-a3f3994eac52b3c914f626143baf4e4d2a3408088d077412ce005fcb0ee1a226). Thanks.

basilv (dismissed):

MCatherine1994:
I think this case won't happen often, but I just noticed that, if I enter the forest client number first, and then enter the username, the forest client number will be removed. Or whenever I update the username, the forest client number will be removed. Not sure if this is done by design.

image

ianliuwk1019:
> I think this case won't happen often, but I just noticed that, if I enter the forest client number first, and then enter the username, the forest client number will be removed. Or whenever I update the username, the forest client number will be removed. Not sure if this is done by design.
> 
> image

That's right, this was same behavior as before I noticed as well, but I don't know if they are by our service design. (I don't think service design taking care of this part). If you use our Dev or Prod environment (which without this IDIM search on userID field), if you enter forest client and verify it first, then enter username, the forestclient card will be removed and you need to click verify again, and user may not know what they need to do.
So, previously I adjust it slightly to hide "verify" button when it is verified, but when something is changed, the "verify" button shows again (for user to verify again). However, I remove that logic since we discuss we like the two fields/button to be consistent showing even after verifed.

MCatherine1994 (approved):

MCatherine1994 (approved):

Issue #718, opened 2023-07-21
Segregate database access with multiple proxys 


gormless87:
**Describe the task**
For auth lambda use database user with limited permissions separate from database user for api lambda

**Acceptance Criteria**
- [x] Create new appropriately named database proxy account for auth lambda
- [x] Analyze required permissions for api lambda proxy account
- [x] reduce privileges for api lambda
- [x] generate sql scripts to create new auth lambda user
- [x] create new AWS Secret for auth lambda user
- [x] auth lambda connections use lower privileged db user 
- [x] api lambda connections use higher privileged db user

**Additional context**
- https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Improving+Security (#9 level 1)
- Additional design work needed to clarify if the auth/api labmda split is sufficient, or if further segregation is warranted.

basilv:
I've started with a PR to create a separate auth lambda db user with restricted access.
The second step (not started yet) is to reduce the permissions of the API db user.

basilv:
Auth lambda work is done. Reducing access of API DB user is in progress, see PR https://github.com/bcgov/nr-forests-access-management/pull/745

basilv:
For the short term solution:
- Auth function: tests run with master db, since we want auth lambda user to be locked down as much as possible (and letting the tests run with auth lambda user requires opening up inserts on various tables)
- API function: tests run with API db user, the expansion of permissions to have tests run is reasonable - just some extra deletes. The work to refactor the tests to avoid these extra delete permissions has been descoped and moved to new issue #751.

PR #731, opened 2023-08-01
feat: Use seperate auth lambda database user (#718)

basilv:
For #718 (at least partially). Changes to existing API lambda user are planned to be kept in a separate PR.


infrastructure/frontend/cloudfront.tf
```diff
@@ -36,9 +36,7 @@ resource "aws_cloudfront_distribution" "web_distribution" {
   price_class         = "PriceClass_100"
 
   viewer_certificate {
-    acm_certificate_arn = "${var.cloudfront_certificate_arn}"
```
basilv, line 39: Temporary hack to deploy front-end

infrastructure/server/aurora-v2.tf
```diff
@@ -167,6 +168,40 @@ resource "aws_secretsmanager_secret_version" "famdb_apicreds_secret_version" {
 EOF
 }
 
+# Create Auth Lambda DB User Credentials
+
+variable "famdb_auth_lambda_user" {
```
basilv, line 173: The lamdba function resource name was suffixed with the environment. Should we do that for all resources by default?

ConradBoydElliottGustafson: It was a recommendation on the "rename" ticket. Doing this for all resources seems like a lot of change for no benefit. Definitely don't want to recreate the DB with a different name (data restore in prod is hard).

ConradBoydElliottGustafson (approved):
Looks awesome. Thanks for fixing frontend for tools environment!

ianliuwk1019 (approved):

ConradBoydElliottGustafson (commented):
I know this is merged already, but the tools_deployment.yml file still has the branch push on it. Just a quick cleanup required.

basilv:
I'm planning to use it on my next branch and will clean it up then.

Issue #735, opened 2023-08-02
FAM #616 change for Error Handling (401/403) is broken and does not behavior correctly.


ianliuwk1019:
**Describe the bug**
Current new frontend for #616 has a removal on **ErrorService** service and adjustment on using Toast Message (using **new component** and **new Toast Message service** with some store interaction). The intention was good but currently **it breaks previous working Error handling behavior for the app/and users**. 409 Error seems to be good from the new change (so far). However, the more server errors handling like **401 (user should be forced to logged out if it happens and after unsuccessful token refreshing)** and **403 (user should not be granted for the action it clicks)** are still broken and we **don't see the toast messages** inform the users what happened and the console has **unhandled promise rejection**.  

**To Reproduce**
It involves verifying user from backend on the jwt token. This will not be easy to be able to reproduce in DEV/TEST environment as you can't easily modify token.
So the easiest way to reproduce them is to run both frontend and backend locally.
And open the frontend page, while testing it, purposely change backend jwt_validation class to throw 401/403 back to the frontend to produce desired http error for testing frontend error handling behavior.

Use "uvicorn api.app.main:app --reload" for easy reload of backend change.

**Expected behavior**
- Correctly handle 401/403 error at frontend and no console error for **unhandled promise rejection**
   401: User should be logged out when token refreshing failed on retry.
   403: User should not be granted to perform the action(s).
- Toast Message dialog with relevant error message should popup to inform the user when 401/403 happens.

**Screenshots**
![image.png](https://images.zenhubusercontent.com/62a9ff93d8506f14758ef03a/473c2df8-2f51-451c-8396-0f34ccf509ff)
![image.png](https://images.zenhubusercontent.com/62a9ff93d8506f14758ef03a/b749cda4-c325-44ab-bb38-bcfccf6d392a)
![image.png](https://images.zenhubusercontent.com/62a9ff93d8506f14758ef03a/0c14fbaa-0c5e-4f22-9a31-d19a7fb66c98)

**Additional context**

PR #746, opened 2023-08-04
fix: #735 error handling fix

NickSaglioni:
- Fixing error handling on Dashboard page


NickSaglioni:
The main issue was the missing keyword `await` before calling the `async` method `getAppUserRoleAssignment()`. This has been fixed in the `onMounted` hook (and it was refactored) and also in the event `selectApplication`.

frontend/src/components/dashboard/Dashboard.vue
```diff
@@ -37,19 +37,23 @@ const userRoleAssignments = ref<FamApplicationUserRoleAssignmentGet[]>();
 
 onMounted(async () => {
     // Reload list each time we navigate to this page to avoid forcing user to refresh if their access changes.
-    try {
-        loading.value = true;
-        applicationsUserAdministers.value = (
-            await applicationsApi.getApplications()
-        ).data;
-        if (isApplicationSelected) {
-            getAppUserRoleAssignment();
```
MCatherine1994, line 46: we don't need to use "await" and "then" at the same time, the error is just caused by we didn't wait for the api call to get user role assignment, just change to "await getAppUserRoleAssignment();"

NickSaglioni: Thanks @MCatherine1994 . So, what are you suggesting, to keep the try/catch block with the `await` on  `getAppUserRoleAssignment()`?

MCatherine1994: yeah, I think that will do the same as your code, you need to test it though

NickSaglioni: I have changed as suggested.

MCatherine1994:
In the case the user is trying to grant himself a role, currently the error message is not displayed correctly in dev, could you please verify if the error message is correct in your branch? If not then we need to fix it

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcklPIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--214aa887970f907a5f14f3b308570adf616d2440/image.png)

MCatherine1994:
and also we need to confirm maybe with Olga, when we want to display error message in toast, when we want to display error message in Dialog. For example, for error 409, we display in dialog,  but for the rest we display in toast? should we keep consistent?

MCatherine1994:
And let's just use "Error" as the title for "409" error, cause after we deploy, we won't get anything from "axiosResponse.statusText".

frontend/src/components/common/ToastMessage.vue
```diff
@@ -5,11 +5,6 @@ import Toast from 'primevue/toast';
 import { useToast } from 'primevue/usetoast';
 import { IconSize } from '@/enum/IconEnum';
 
-import {
-    useNotificationMessage,
-    useErrorDialog,
```
MCatherine1994, line 10: Hi Nick, since we are not using the Dialog to display error anymore, let's check if we still need this "useErrorDialog" in state, or we can remove it completely. We might can also remove the Dialog component in the Summary page.

frontend/src/components/common/ToastMessage.vue
```diff
@@ -27,42 +22,18 @@ const onError = (error: any, info: string) => {
...
-            useErrorDialog.dialogMsg = axiosResponse.data.detail;
-            useErrorDialog.isErrorVisible = true;
-        }
+        const axiosResponseHasDetail = axiosResponse?.data.detail
+            ? axiosResponse.data.detail
+            : genericErrorMsg.text;
+
+        showToastTopRight(
+            'error',
+            axiosResponse ? axiosResponse.statusText : genericErrorMsg.title,
```
MCatherine1994, line 32: let's remove the  axiosResponse.statusText, cause status text is not coming back from the server, but from the axios library. We can just use the generic error title here.

frontend/src/components/common/ToastMessage.vue
```diff
@@ -27,42 +22,18 @@ const onError = (error: any, info: string) => {
     if (axios.isAxiosError(error)) {
         const err = error;
         const axiosResponse = err.response;
-        const status = axiosResponse?.status;
-
-        const e401_authenticationErrorMsg = {
-            title: 'Error',
-            text: 'You are not logged in. Please log in.',
-        };
```
MCatherine1994, line 35: Hi Nick, I think we still want to keep these the 401 and 403 error handling. We want the hardcode error messages, cause not all messages from backend are intended for frontend. So let's still keep these if conditions. But for the special case "can't grant self access", which is also a 403 error, we can check if the error detail code is "self_grant_prohibited", then we display the error description from the server. For the rest 403 error cases, we can still show the hardcoded message as previous. 
```
detail: {code: "self_grant_prohibited", description: "Granting roles to self is not allowed"}
```

I'll create a follow up ticket for further discussion about how server should return the error for frontend to handle. But for now, let's still follow the old way, but add the special case for  "can't grant self access". For 409 error, we don't need to hardcode, just show the error from the server.

frontend/src/components/common/ToastMessage.vue
NickSaglioni: I have removed because it is not used anymore.

frontend/src/components/common/ToastMessage.vue
NickSaglioni: Ok, using the generic error title.

frontend/src/components/common/ToastMessage.vue
NickSaglioni: Bringing it back and adjusting for `self_grant_prohibited` case.

MCatherine1994 (approved):

Issue #733, opened 2023-08-02
Disable grant access button when there is no application selected before next release


MCatherine1994:
**Describe the task**
Disable grant access button when there is no application selected

**Acceptance Criteria**
- [x] User cannot click on the grant access button from the side nav bar when there is no application selected

PR #749, opened 2023-08-08
fix: #733 Disable grant access button when there is no application selected

NickSaglioni:
- Adding the JSON data control ability from the `sideNav` component to the parent component;
- Checking if the `Grant Access` option should be enabled or disabled according to the value of `isApplicationSelected` state by adding a variable watcher.


MCatherine1994:
Let's merge the main branch into this branch

MCatherine1994 (approved):

ianliuwk1019:
Is this still a draft pr? Also update from main branch is needed.

NickSaglioni:
> Is this still a draft pr? Also update from main branch is needed.

Yes, Ian. I have put it in draft because I am working to solve a Sonar Cloud error reported.

MCatherine1994 (approved):

Issue #714, opened 2023-07-20
Cleanup FAM frontend code


MCatherine1994:
**Describe the task**
Since we did a lot of refactor and rewrite for the frontend pages, need a cleanup to remove unused components.

**Acceptance Criteria**
- [ ] remove all the unused components:
- Footer
- Breadcrumb as we might switch to use primevue’s breadcrumb
- Home
- ManageAccess
- SelectApplication
- Clean up route path options
- image
- styles
- package.json
- [ ] test application is still working

PR #756, opened 2023-08-09
fix: #714 cleanup frontend code

NickSaglioni:
1. Removed the unused components:
- Footer;
- Breadcrumb;
- Home;
- ManageAccess;
- SelectApplication.
2. Cleaned up all route paths
3. Test application is still working


MCatherine1994:
Hi Nick, sorry I forgot to mention, maybe we want to check the tokens in the default-theme as well, I wonder if those are for the old bootstrap style? do we still need them? thanks!!

NickSaglioni:
> Hi Nick, sorry I forgot to mention, maybe we want to check the tokens in the `frontend/src/assets/styles/default-theme.scss` as well, I wonder if those are for the old bootstrap style? do we still need them? Also this one `frontend/src/assets/styles/components/form-elements.scss` thanks!!
> 
> Also want to check the images, and packages in package.json

Assets and packages have been cleaned up.

frontend/package.json
```diff
@@ -14,11 +14,6 @@
     },
     "dependencies": {
         "@bcgov-nr/nr-fsa-theme": "^1.1.0",
-        "@fortawesome/fontawesome-svg-core": "^6.2.1",
-        "@fortawesome/free-regular-svg-icons": "^6.2.1",
-        "@fortawesome/free-solid-svg-icons": "^6.2.1",
-        "@fortawesome/vue-fontawesome": "^3.0.2",
-        "@popperjs/core": "^2.11.6",
```
MCatherine1994, line 21: I'm not sure where we use this @popperjs/core package previously, just want to make sure we tested everything still works after removed it, thanks!!

MCatherine1994:
let's merge from the main branch, and then do a clean install to get the clean package-lock.json

frontend/package.json
NickSaglioni: PopperJS is for tooltip and pop-ups, we are not using anywhere in the new design.

MCatherine1994 (approved):

Issue #561, opened 2023-04-24
Audit: Retain FAM Access Logs & Event Logs 


ArogeG:
As a security auditor, I want information on user activity in FAM including user logins/ user access and changes to user authorization so that I can review audit information 

Retain audit information on the following

- Log information on users logged in to FAM including time of login; application logged into; permissions provisioned at the time of login. 
- Event log information user role assignment changes including user that assigned/ unassigned roles; previous role assignment; and current assignment; Time when the change was made. 
- Tracks end user activity, including admins. Still relying on AWS to mitigate hacking exploits.

Security NFRs:
- Audit information should be stored somewhere read-only where an attacker who hacks the app can't modify.
- Plan like the audit information will need to be used in a legal proceeding.
- Retention period of audit logs to be defined, for MVP can likely just be retain permanently.

**Acceptance Criteria**
- [ ] Ability to export logs of users by applications  and the roles/permissions used
- [ ] Demonstrate that we can query the logs for investigation

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

Future task is to determine how audit and event logs will be displayed. 
More non-functional requirement is required for further logs

MCatherine1994:
Do we want to have all the logs in cloud watch first? or we want to store them somewhere? @basilv @ConradBoydElliottGustafson @ianliuwk1019

ianliuwk1019:
Not sure how long the cloud watch retains log information.
It would be nice we could have all logs on Cloud Watch or CloudTrail and be able to query or export about the log.

However, Basil point out in previous spike #566: "activity by specific FAM end users. For that we'd have to provide some audit logging capability within FAM itself.".

So for this ticket, I am not sure myself about the scope or should it be broken down into smaller tasks if we need to be able to save log files to somewhere, can have "requester" and other audit information together in saved log, be able to run query/export for specific user/application/event.

ianliuwk1019:
Added useful team reference for Security Improvement outlined from architect for item: "[Line 7: Lack of auditing capability](https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Improving+Security)"

ianliuwk1019:
Log looks like this currently (a GET user_role_assigments):
[...]

PR #766, opened 2023-08-10
feat: #561 audit for auth login 

MCatherine1994:
Add audit for auth login #561, with following information:

"logging_type": "Audit",
"event_type": "User Login",
"event_outcome": 1 of success, 0 for fail (same code as api audit log),
"user_name": IDP username,
"user_type": IDP user type,
"cognito_client_id": Application where the login happened
"access_roles": Role this user has in FAM


server/auth_function/lambda_function.py
```diff
@@ -29,10 +37,15 @@
...
     IDP_NAME_BCSC_DEV: USER_TYPE_BCSC_DEV,
     IDP_NAME_BCSC_TEST: USER_TYPE_BCSC_TEST,
-    IDP_NAME_BCSC_PROD: USER_TYPE_BCSC_PROD
+    IDP_NAME_BCSC_PROD: USER_TYPE_BCSC_PROD,
 }
 
 
+class AuditEventOutcome(str, Enum):
+    SUCCESS = 1
+    FAIL = 0
```
ianliuwk1019, line 46: Showing (1 or 0) on audit log could be hard to for audit person to understand (since this is for audit, not for debugging). 
(I understand this is copied from api-audit and it has the same; but initially I think I used "Success" and "Fail" to be more clear but I think api side was changed to 1/0).

Also if we like to use 1/0, then in Python, probably don't need (str) argument in the class (as initially previously the intention was for "string" enum, not the default enum (not 100% sure).

server/auth_function/test/conftest.py
```diff
@@ -277,8 +276,8 @@ def create_fam_child_parent_role_assignment(db_pg_transaction):
     cur.execute(insert_parent_role_sql)
 
     query = (
-        "select role_id from app_fam.fam_role " +
-        f"where role_name = '{parent_role_name}'"
+        "select role_id from app_fam.fam_role "
+        + f"where role_name = '{parent_role_name}'"
```
ianliuwk1019, line 280: Probably try without a need for "+".

ianliuwk1019 (approved):
Approve with minor comments.
Seems there are some sonar issue as well.

server/auth_function/lambda_function.py
MCatherine1994: I see... yeah, I just try to follow the api side, will wait before merge this one

server/auth_function/test/conftest.py
MCatherine1994: oh I feel that's the auto formatter

ianliuwk1019 (approved):

basilv (approved):

PR #767, opened 2023-08-10
fix: Enable dev deployment for push on main.

ianliuwk1019:
Previously was disabled for BCSC integration needs. Now enabled again.


basilv (approved):

ianliuwk1019:
Was about to merged but you were ahead of me. Thanks Basil!

Issue #478, opened 2023-03-09
Email Notification When Users are Added to an Application


ArogeG:
**As a** *FAM Application Administrator* **I want** *FAM to send an email notification from the FAM after a user access is provisioned* **So That** *a user can be informed of the access that was  provisioned.  

**Additional Context**

- A standard email should be sent to a user with information of the application and role that they have access for. 
- Uses the GC Notify email service for notification : https://documentation.notification.canada.ca/en/limits.html
-API to integrate GC Notify with FAM will need to be developed. 
- Functionality to support custom applications will be addressed in later versions
- Review [CHES](https://bcgov.github.io/common-service-showcase/services/ches.html)  with GC Notify service: https://notification.canada.ca/
**Acceptance Criteria**
- [x] Integrate GC Notify in the backend
-[ ]  set the send email to do not reply

**Email Message should read:**
Do not reply, this is an automated email that is not monitored

Hello, 

You have been granted access to Seed Planning and Registry Application (SPAR) [Application Description].

Please contact your administrator if you have any issues accessing the application.

...........................................................................................................................................................................................................

**Definition of Done**
- [ ] Ready to Demo in Sprint Review
- [ ] Does what I have made have appropriate test coverage?
- [ ] Documentation and/or scientific documentation exists and can be found
- [ ] Peer Reviewed by 2 people on the team
- [ ] Manual testing of all PRs in Dev and Prod
- [ ] Merged

ArogeG:
For the first iteration, we could begin the email Hello

ArogeG:
We can start work on the back-end to integrate with the email service. For the front-end, we will require the IDP webservice to call the name(s) and email address.

MCatherine1994:
Tried the GC Notify, setup an account, setup an email template and an api key, tested through curl command:
```
curl -X 'POST' \
  'https://api.notification.canada.ca/v2/notifications/email' \
  -H 'accept: */*' -H 'Authorization:ApiKey-v1 <my_api_key>' \
  -H "Content-Type: application/json" \
  -d '{ "email_address": "catherine.meng@gov.bc.ca", "template_id": "cd46fd74-7d79-4576-97ef-8b93297def24", "personalisation": {"username": "Catherine","appname": "FAM"}}'  
```


GC Notify only has a production server, CHES has dev/test/prod three server. They both support send files in email. The email content looks a little bit different:

An example email sent by GC Notify:
[...]

PR #789, opened 2023-08-14
feat: #478 add gc notify email endpoint

MCatherine1994:
Add gc notify email endpoint #478 
Add test, ignore the successful case, cause don't want to send the real email. GC notify only has production server. The api key is stored in the secret manager in AWS FAM prod


server/backend/api/app/schemas.py
```diff
@@ -300,8 +296,14 @@ class IdimProxyIdirInfo(BaseModel):
...
+            found=json_dict["found"],
+            userId=json_dict["userId"],
+            displayName=json_dict["displayName"],
         )
-        return info
\ No newline at end of file
+        return info
+
+
+class GCNotifyEmailParam(BaseModel):
```
MCatherine1994, line 306: This is the only change in this file, the rest are just done by the auto formatter

server/backend/api/config/config.py
```diff
@@ -167,6 +171,11 @@ def get_idim_proxy_api_key():
     return idim_proxy_api_key
 
 
+def get_gc_notify_email_api_key():
```
MCatherine1994, line 174: This is the only change in this file, the rest are just done by the auto formatter

basilv (approved):
Need to ensure the API key is added to the GitHub sections for Actions & Dependabot

server/backend/api/app/routers/router_gc_notify.py
```diff
@@ -0,0 +1,17 @@
...
+from api.app.integration.gc_notify import GCNotifyEmailService
+from api.app.schemas import GCNotifyEmailParam
+from fastapi import APIRouter
+
+LOGGER = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.post("/send", response_model=str)
```
basilv, line 12: In terms of the design, its better security not to expose the actual send_email endpoint, and just have other back-end processes generate (send) emails when required.

server/backend/api/app/integration/gc_notify.py
```diff
@@ -0,0 +1,53 @@
...
+        self.email_base_url = GC_NOTIFY_EMAIL_BASE_URL
+        self.headers = {
+            "Content-Type": "application/json",
+            "Authorization": "ApiKey-v1 " + self.API_KEY,
+        }
+
+        self.session = requests.Session()
+        self.session.headers.update(self.headers)
+
+    def send_email(self, params: GCNotifyEmailParam):
```
basilv, line 32: I don't see where the content for the email is supplied. Presumably the template_id relates to this somehow.

I also don't see where the from and reply-to email addresses are configured, maybe that's handled by the GC-Notify service.

Normally we'd want all emails to have a common subject prefix and a common body suffix.
E.g. email subject: FAM [ENV] - 
email suffix: This email was sent automatically by the FAM application from an unmonitored email account. Or, provide a valid reply-to if we actually want email interaction with end users (unlikely).

MCatherine1994: The email content and from are all set in GC Notify. It's just a simple message for now:

image
 

And the config settings:
Screen Shot 2023-08-14 at 3 47 59 PM

basilv: Okay, that's good, we should likely name the method send_granted_access_email

server/backend/api/app/routers/router_gc_notify.py
MCatherine1994: I could remove it for now, and we could just call the curd method when need to send the email. I believe that will in the process of create user role assignment.

server/backend/api/app/exception_handlers.py
```diff
@@ -23,9 +23,24 @@ async def requests_http_error_handler(request: Request, exc: HTTPError):
...
+        # this is the error format for gc notify
+        failureCode = response_text["errors"][0]["error"]
+
+    error_message = None
+    if "message" in response_text:
+        error_message = response_text["message"]
+    elif "errors" in response_text and "message" in response_text["errors"][0]:
+        # this is the error format for gc notify
+        error_message = response_text["errors"][0]["message"]
+
```
ianliuwk1019, line 40: Just a note here.
As you probably **will remove the "router" for GC notify** (for your testing purpose) later, I am thinking adding gc notify modification in "exception_handlers.py" here probably has no point.

I am quite positive (but not 100% sure), the exception_handlers.py is for **FastAPI**, so the exception you get from gc notify probably only will be propagated and caught (by **exception_handlers**) until reaching the "**router**". So if the router will be removed, you might need to handle exception in the email sending code itself or create a **custom exception** to throw for CRUD code to handle email sending errors.

MCatherine1994: Hi Ian, I think we use this general exception handler in all crud methods, so I changed in this general place, do you feel I need to handle the email error separately, or this should be okay?

Screen Shot 2023-08-15 at 10 02 26 AM

ianliuwk1019: Hi Catherine, 
The exception handler is for FastAPI (router), it only catches CRUD exception when there is exception propagates to the router or throw from CRUD (if CRUD does not handle it).
However, on the second thought, in the future when granting access (CRUD) interacts/uses the new gc_notify class, then it will become relevant. So leaving it there is fine.

The second point that I mentioned is because I see this and is not clear to me (maybe only I am confused) from the GC Notify instruction; want to see if you are aware if it needs some "**status**" checking to really make sure it sends (or maybe we don't care if it really sends...). So, if we care, possibly maybe we need to handle error specifically in the class, like retry...?
image
@basilv

MCatherine1994: I tested with invalid request, and it can catch the error successfully. It's hard to test the 5xx case though.

MCatherine1994: I got Ian's point. Wonder how we'd like to handle the error case, do we want to retry to send the email? or what we'll do?

MCatherine1994: I got Ian's point. Wondering how we'll handle the error case, if we want to try to resend the email? or what else we'll do?

MCatherine1994: Created a new ticket for integration in frontend and the error handling #799

ianliuwk1019 (approved):

ianliuwk1019 (approved):

Issue #696, opened 2023-07-13
Remove superfluous/unused AWS resources


gormless87:
**Describe the task**
Superfluous, undocumented, unused AWS resources have been deployed, presenting a higher-than-necessary attack surface. Catalog these resources and clarify purpose with cloud pathfinder team, deleting those that are unneeded.

**Acceptance Criteria**
- [x] List of ASEA resources 
- [x] Diff our TF resources with Cloudpathfinder resources
- [x] Request/discuss removal with Cloudpathfinder team

**Additional context**
- See also recommendation from Louis K (Investigate generating SBOM (cyclonedx format preferred) for eventual ingestion in ministry component aggregator & vulnerability mgmt. solution)
- https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Improving+Security
- 

Addendum: See https://apps.nrs.gov.bc.ca/int/confluence/x/fgQ9Cg for recommended inventory maintenance

basilv:
I enabled AWS Resource Explorer in the AWS Dev Workspace. After a short delay, this allowed retrieving the full list of resources (over 200) for the account. Next step is figuring out how to analyze this list.

basilv:
Findings:
1. Ensure that "all cloud regions" is selected when viewing resources to export. We have some resources in the Global region
2. We're unable to list (scan for) resources (e.g. certificate) in the us-east-1 region. Error is: 
Access denied: You don't have permission to call resource-explorer-2:ListViews in us-east-1.
Not a big issue as its just the cert missing, although technically from a security standpoint we're unable to identify cloud pathfinder / ASEA resources in other regions.
3. Not all resources created by our FAM team are tagged with "managed-by: terraform". This is either because a few resources (e.g. bastion host ec2) are manually created, or before terraform isn't configured to do the tagging - e.g. the S3 buckets for fam-flyway, fam-cloudfront, terraform-import, etc.
4.

basilv:
Findings:
- Network interfaces (for e.g. lambdas) are given random names, I think created new for each deployment, and old ones aren't being cleaned up by terraform. Really hard to review this via resource explorer, use EC2 Network Interfaces view instead.

basilv:
Finished prod analysis. Removed unused resources where possible. Spreadsheet with resource list (FAM-aws-prod-resources) uploaded to Improving Security confluence page. Ready for review.

PR #795, opened 2023-08-15
chore: #696 Add managed-by=terraform tags

basilv:
A number of FAM AWS resources are not created using the managed-by=terraform tag.


basilv:
Cancelling this work. Turns out not to be necessary for AWS resource analysis.

Issue #725, opened 2023-07-27
Add WAF through terraform


MCatherine1994:
**Describe the task**
Create WAF ACLs using terraform, and add WAF to the cloudfront and api gateway and cognito

**Acceptance Criteria**
- [ ] WAF ACLs has been created through terraform
     - One for Cloudfont with Global region, with 5 rules:
                - Amazon IP reputation list (AWS-AWSManagedRulesAmazonIpReputationList)
                - Core rule set (AWS-AWSManagedRulesCommonRuleSet)
                - Known bad inputs (AWS-AWSManagedRulesKnownBadInputsRuleSet)
                - Linux operating system (AWS-AWSManagedRulesLinuxRuleSet)
                - SQL database (AWS-AWSManagedRulesSQLiRuleSet)
     - One for Api Gateway with Canada region, with the same 5 rules as above
     - One for Cognito with Canada region, with 3 rules, because the other two rules block the access outside Canada, Nick and Jota can't access to FAM when added those two rules
                - Amazon IP reputation list (AWS-AWSManagedRulesAmazonIpReputationList)
                - Known bad inputs (AWS-AWSManagedRulesKnownBadInputsRuleSet)
                - Linux operating system (AWS-AWSManagedRulesLinuxRuleSet)
                
- [ ] WAF has been added to the cloudfont, api gateway and cognito through terraform
- [ ] Added a comment for the ACL for Cognito, explain why we can only have three rules

**Aditional info**
Test this out in the TOOLS environment - hack-away!

ianliuwk1019:
For web ACL for Cloudfront, AWS says:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBb2NWIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e961b6417d7f94ba47d0acc627a58c116ee1520b/image.png)
However, I am facing an error: 
```
│ Error: reading WAFv2 WebACLs: WAFInvalidParameterException: Error reason: The scope is not valid., field: SCOPE_VALUE, parameter: CLOUDFRONT
│ {
│   RespMetadata: {
│     StatusCode: 400,
│     RequestID: "4867c2ce-9e12-4931-b752-9c997aefd894"
│   },
│   Field: "SCOPE_VALUE",
[...]

PR #797, opened 2023-08-15
feat: #725 add waf terraform

ianliuwk1019:
Adding basic WAF to API-Gateway, Cognito and Cloudfront:

* Add aws "us-east-1" as second alias provider.
* Add Cloudfront WAF ACL with provider at us-east-1 region.
* Associate Cloudfront WAF ACL to aws_cloudfront_distribution.
* Add API-Gateway and Cognito WAF ACL.
* Associate Cognito WAF ACL with aws_cognito_user_pool.
* Associate API-Gateway WAF ACL with gateway "stage".


infrastructure/server/api-gateway.tf
```diff
@@ -66,10 +66,22 @@ resource "aws_api_gateway_deployment" "fam_api_gateway_deployment" {
   }
 
   rest_api_id = aws_api_gateway_rest_api.fam_api_gateway_rest_api.id
-  stage_name  = var.api_gateway_stage_name
+  stage_name  = var.api_gateway_stage_name  # Terraform warns about using "stage_name" in resource "aws_api_gateway_deployment".
+                                            # See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_deployment
+                                            # If possible in the future, use "aws_api_gateway_stage" resource instead. But may face issue about:
```
basilv, line 71: Did you create an issue in the backlog for this?

ianliuwk1019: Hi Basil, yes this is added into ticket #810 .

infrastructure/server/waf-firewall.tf
```diff
@@ -0,0 +1,226 @@
+locals {
```
basilv, line 1: I wonder if this file should be called waf-firewall-api-gateway, similar to how the waf-firewall-cloudfront is named?

infrastructure/server/waf-firewall.tf
```diff
@@ -0,0 +1,226 @@
...
+    # Example: "fam-dev-waf"
```
basilv, line 2: Remove this comment

infrastructure/server/waf-firewall.tf
```diff
@@ -0,0 +1,226 @@
...
+    web_acl_name_prefix = "fam-${var.target_env}-waf"
+}
+
+
+# API-Gateway WAF ACL
+locals {
```
basilv, line 8: Not sure why there are two locals blocks, can't this be combined into one.

terraform-frontend/terragrunt.hcl
```diff
@@ -43,5 +43,16 @@ provider "aws" {
     role_arn = "arn:aws:iam::$${var.target_aws_account_id}:role/BCGOV_$${var.target_env}_Automation_Admin_Role"
   }
 }
+
+# Additional provider configuration for us-east-1 region; resources can reference this as `aws.east`.
+# This is essential for adding WAF ACL rules as they are only available at us-east-1.
```
basilv, line 48: should clarify that "for adding WAF to Cloudfront", us-east-1 is needed.

basilv (changes requested):

infrastructure/server/waf-firewall.tf
ianliuwk1019: Hi Basil, my initially intention of the file is to have all WAF ACL rules defined in one file only, since they are just rules (not association) and since we don't have "modules" structure (such as "./modules/waf") in our Terraform file folder.
So the waf-firewall.tf file contains **both** the **api-gateway** rules and **cognito** rules to be in one location (and only split for Cloudfront due to limitation of our Terraform structure is server/frontend) .
However, if you prefer to split them I will do that.

infrastructure/server/waf-firewall.tf
ianliuwk1019: Yes, they can be combined.
I chose it to be separated since Terraform has feature for grouping "locals" for the same category and references "**common locals**" in different "locals" blocks so I define this locals to be close to the API-Gateway rule resource block where the value is actually being referenced.

MCatherine1994:
Thanks Ian, looks good to me!

terraform-frontend/terragrunt.hcl
ianliuwk1019: I can add the reference link here if that helps. The reason it has to be in us-east-1 is also documented in this ticket as well.

MCatherine1994 (dismissed):

terraform-frontend/terragrunt.hcl
ianliuwk1019: Added AWS reference link after this comment line..

infrastructure/server/waf-firewall.tf
ianliuwk1019: comment removed.

infrastructure/server/waf-firewall.tf
basilv: Ah, okay so this is really the WAF resources for the backend, while the other is the WAF resources for the front-end.

basilv (approved):

infrastructure/server/waf-firewall.tf
ianliuwk1019: Thanks Basil. I will mark this one resolved.

MCatherine1994 (approved):

Issue #800, opened 2023-08-16
Frontend test breaks the pipeline build


NickSaglioni:
The frontend pipeline build is currently broken by an error in the test file ProtectedLayout.spec.ts.

PR #801, opened 2023-08-16
fix: #800 Frontend test breaks the pipeline build

NickSaglioni:
Fixing error on file ProtectedLayout.spec.ts which breaks the pipeline build for the frontend.


MCatherine1994:
seems the npm build works fine locally

MCatherine1994 (approved):

Issue #739, opened 2023-08-03
Replace the close button for the error dialog to top right


MCatherine1994:
**Describe the task**
Fix the close button for the error dialog when assign duplicate role. Should we have the close button on the top right?
![image.png](https://images.zenhubusercontent.com/6201a0f0c6815e7dfafa3f78/00e2a851-f6f3-4056-abb5-5deac6bbe004)

**Acceptance Criteria**
- [ ] replace the close button to the top right

J0taFerreira:
That is weird, I tested it in three different browsers and it is ok for me. Can we talk about it so I can understand it better?![409_edge.png](https://images.zenhubusercontent.com/649090ffb7619f1bbb8a8889/035a6869-d440-4750-907c-bec9ab31d9a4)![409_firefox.png](https://images.zenhubusercontent.com/649090ffb7619f1bbb8a8889/57a72287-a658-4425-b338-9a90d2fece00)![409_google.png](https://images.zenhubusercontent.com/649090ffb7619f1bbb8a8889/0ae6d3d2-5f1f-4e7c-9692-49388b915161)

MCatherine1994:
When we run our code locally, the api call for duplicate user role assignment will return a status code "409 conflict", but in FAM dev, it only returns "409"

![image.png](https://images.zenhubusercontent.com/6201a0f0c6815e7dfafa3f78/10d30631-520d-446c-a353-bab5fe2f7720)

![image.png](https://images.zenhubusercontent.com/6201a0f0c6815e7dfafa3f78/73e4949e-402f-441e-b220-9cc81fef36f3)

ianliuwk1019:
Backend returns "**HTTPStatus.CONFLICT**" in code for FastAPI. I don't see we manipulate that. But as long as it is still a 409, it is correct. I have no idea that extra "Conflict" is coming from.
Does it create problem for the frontend? I assume not.

MCatherine1994:
yeah... frontend will think backend will return "409 Conflict", so "CONFLICT" will be the title of the popup modal. But it returns only on local, but not in FAM dev. Not sure why it's difference between local vs deployed version. I think the frontend should have a default title though, just in case we don't have error title returned from status code.

ianliuwk1019:
I see. Seeing the "Title" as "CONFLICT" isn't right to me. I would not believe it is a language from the service design.
[...]

PR #804, opened 2023-08-16
fix: #739 fix dialog icon to always display at top right

MCatherine1994:
- add a default error title for error dialog refs: #739
- add the style change in the common theme repo

### DIFF
--- a/frontend/src/components/common/ToastMessage.vue
+++ b/frontend/src/components/common/ToastMessage.vue
@@ -58,7 +58,7 @@ const onError = (error: any, info: string) => {
             );
         } else if (status == 409) {
             useNotificationMessage.isNotificationVisible = false;
-            useErrorDialog.dialogTitle = axiosResponse.statusText;
+            useErrorDialog.dialogTitle = axiosResponse.statusText || 'Error';
             useErrorDialog.dialogMsg = axiosResponse.data.detail;
             useErrorDialog.isErrorVisible = true;
         }


frontend/src/components/common/ToastMessage.vue
```diff
@@ -58,7 +58,7 @@ const onError = (error: any, info: string) => {
             );
         } else if (status == 409) {
             useNotificationMessage.isNotificationVisible = false;
-            useErrorDialog.dialogTitle = axiosResponse.statusText;
+            useErrorDialog.dialogTitle = axiosResponse.statusText || 'Error';
```
ianliuwk1019, line 61: Previous code (which I didn't know why) should not have used Axios' "**statusText**" for displaying to user on Dialog component. This is not a message/code/title from our backend, is purely Axio's http verbose, and probably depends on the library version.
I feel this **axiosResponse.statusText** should be removed and for now may just gave it a general suitable title based on what status it encounters.

ianliuwk1019:
Wondering why do we need to change style in common theme repo? If the Dialog itself always default to a general title (if not passing any title) or if giving an " " empty title when no title is passed in, will that work?
Primvue probably has something available to adjust it so we don't need to override style, or not?

MCatherine1994:
> Wondering why do we need to change style in common theme repo? If the Dialog itself always default to a general title (if not passing any title) or if giving an " " empty title when no title is passed in, will that work?
> Primvue probably has something available to adjust it so we don't need to override style, or not?

yeah I was thinking about that before. But I feel it's okay to make the change. But we'll switch to use toast message to display errors anyway (in Nick's pr), I wonder maybe there's no need for this pr. Maybe I can just close it.

frontend/src/components/common/ToastMessage.vue
MCatherine1994: I'm actually not so sure, cause when we run our code locally, it returns the statusText. But after deploy to AWS, it doesn't show anymore.

MCatherine1994:
Will close this one, we'll switch to use toast notification to show server errors in https://github.com/bcgov/nr-forests-access-management/pull/746

PR #807, opened 2023-08-17
fix: #561 update audit log

MCatherine1994:
- fix api audit log target user cognito user id 
- update auth audit log format to be the same as api log


server/auth_function/lambda_function.py
```diff
@@ -67,21 +67,25 @@ def lambda_handler(event: event_type.Event, context: Any) -> event_type.Event:
     """
 
     audit_event_log = {
-        "logging_type": "Audit",
-        "event_type": "User Login",
-        "event_outcome": AuditEventOutcome.SUCCESS,
+        "auditEventTypeCode": "USER_LOGIN",
+        "auditEventResultCode": AuditEventOutcome.SUCCESS.name,
+        "requestingUser": {},
```
ianliuwk1019, line 72: Although I do actually prefer your previous "key" with snake_case (normally I don't like snake_case) and "value" in plain English, I don't object on the changes.
Not sure what Basil/Pete like to be consistent.
I heard there is no standard format, but I see few examples that they all seem to have at least a "value" for event "Audit" only. I think probably it is for easy search and distinguish than other logging purpose. Maybe we don't need to do that... I don't know what's better actually.
Here is one example from AWS Consul(but not for our purpose):
image

MCatherine1994: yeah, it's just that Conrad's api log all use this format, for consistency I changed

ianliuwk1019 (approved):

Issue #814, opened 2023-08-22
Frontend - Pipeline Build and Deploy Slow Down Significantly


ianliuwk1019:
**Describe the bug**
This is an observation in pipeline and also can be seen in locally.
After probably task #704 (switch to carbon icons) was merged to main, frontend build and deployment slow down significantly.
From normally **2~ 3 minutes** to now can be observed between **13~20 minutes** (normal Github action slowness probably isn't the main factor at this issue).

Observation also includes:
- The frontend production build (locally and at pipeline) produces lots of import-artifacts at "dist/assets" folder (more than **7000** small files).
- During frontend deployment, these small files are contributing to the large amount of AWS bucket creation and large increase of transfer time to AWS buckets.

Possible reason:
- It is very likely due to the frontend Icon.vue component is using "**dynamic import**" for Carbon Icons.
  [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBck1qIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--b6de1b07d5c5ab3c6249fedf135a3377482a1dfa/image.png)
   "**Vite**" probably does not know how to resolve this (and thus include everything ?? no tree shaking?? Just a guessing).
   The intention having one component to dynamically loading icon was good but maybe it isn't the scenario to fit in this case for loading icons.

Root cause and solution are not immediately clear yet but need to be investigated for possible mitigations .
- Can "Vite" have build option to optimize this dynamic icon imports issue, plus a way for icons packing into a small bundle?
- Is it possible to still have "Icon.vue" but not using dynamic imports for icon to resolve the problem (so limit this issue within Icon.vue and no impact for other codes)?
- Don't use dynamic imports for icons for this case (basically partially re-do #704)?
- Other options?

**To Reproduce**
Please observer past pipeline run mentioned at "Additional context":
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcmtqIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--d1f15cddec152d312f7400899f17ccebf7bca2c9/image.png)
[...]

PR #816, opened 2023-08-22
fix: #814 update dynamic icon import to fix pipeline slow issue

MCatherine1994:
- update frontend dynamic import icons to only import the icons we used refs: #814

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -14,7 +14,6 @@
                 "axios": "0.26.1",
                 "bootstrap": "^5.2.3",
                 "fam-api": "file:../client-code-gen/gen",
-                "primeicons": "^6.0.1",
                 "primevue": "^3.27.0",
                 "vee-validate": "^4.7.3",
                 "vue": "^3.2.38",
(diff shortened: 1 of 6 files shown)


ianliuwk1019:
I checked if Carbon also provides a "@Type" package.
Looks like it does have types for [@types/carbon__icons-react](https://www.npmjs.com/package/@types/carbon__icons-react)  but it does not have https://www.npmjs.com/package/@types/carbon__icons-vue...
Seems Vue is still not fully supported by Carbon with typing after years...

ianliuwk1019:
The PR looks good. Thanks Catherine!
I have researched into vite option to optimize a little on our frontend.
Could you also add this option in "vite.config.ts" (vite is a compile/packaging tool for Vue for production), it does not impact your change but it will do a better bundling in this case for dynamic import for async components.
If you are not sure where to add, I can quickly show you tomorrow or I can add it in your branch.
```
        build: {
            chunkSizeWarningLimit: 1600,
            outDir: "dist",
            sourcemap: true,
            rollupOptions: {
                output: {
                  inlineDynamicImports: true,
                },
            }
        },
```

ianliuwk1019 (approved):

ianliuwk1019 (approved):

Issue #750, opened 2023-08-08
Frontend: minor fixes


NickSaglioni:
Creating a ticket for @basilv 's comments on the frontend:

1. Hover over the profile icon shows "Log Out", but opening the user profile has an option "Sign out". The two should be consistent.
2. The app title in the titlebar is "Forest Access Management" but the dropdown to select an application has "Forests Access Management" (plural vs singular for Forest). The two should be consistent.
3. There's disabled content in the left side panel - management settings, support, need help. In my opinion this disabled content should be removed before going to prod if it isn't being enabled.
4. I miss not seeing the current user's name anywhere. (Maybe this isn't part of the new design.)

NickSaglioni:
1 - Changing the text to 'Profile', which makes more sense to the displayed icon;
2 - The correct is 'ForestS' in plural. Fixing the frontend;
3 - After discussing this with @OlgaLiber2, the disabled content will be removed for now;
4 - Adding the logged user_name info in the profile component.

PR #819, opened 2023-08-23
fix: #750 frontend minor fixes

NickSaglioni:
1 - Changing the text to 'Profile', which makes more sense to the displayed icon;
2 - The correct is 'ForestS' in plural. Fixing the frontend;
3 - The disabled content for the sideNav will be removed for now;
4 - Adding the logged user_name info in the profile component.


MCatherine1994:
Hi Nick, please check the code smell in sonar cloud analysis, and I'm not sure if you added the user_name to the profile.

NickSaglioni:
> Hi Nick, please check the code smell in sonar cloud analysis, and I'm not sure if you added the user_name to the profile.

Hey Catherine, the code smells are bits of code that were commented out because we will not be showing the disabled options in the sideNav component. But once we are able to show them, they can be uncommented back. Should we remove it even though is going to be redone in the future?

The user name is being displayed in the profile sidebar component, could you please check again?

image

Issue #755, opened 2023-08-09
Cleanup primevue stylesheet, and update custom classname


MCatherine1994:
**Describe the task**
Cleanup the primevue stylesheet in the common theme repo, remove the style changes just for our FAM frontend, and add to FAM. Rename our custom classname in FAM frontend, so we could know which ones are from primevue, which ones are custom just for FAM.

**Acceptance Criteria**
- [x] Remove the custom style change just for FAM from the common repo, add to FAM
    - for example, the access-table, and the side-nav, those are custom classname in FAM, and specific to FAM
    - remove those from common repo, and add to FAM
- [x] Rename all the custom classname to follow the pattern "custom-"
    - for example, the "p-access-table" will be "custom-access-table", so we know this is not a primevue classname, but a custom classname we defined. The classname in sidenav needs to change as well, cause we're not using the primevue side nav, but write our own one
    - check if there is any other custom classname in other components
- [x] Add missing style for dialog buttons, this is what left from the issue #616

PR #822, opened 2023-08-24
fix: #755 cleaning up stylesheet, and updating classes

J0taFerreira:
Cleaning up the common repo, updating classes


frontend/src/assets/styles/default-theme.scss
```diff
@@ -1,33 +1,36 @@
...
+
+//text
+$light-text-primary: map.get(lightTheme.$light-theme, 'text-primary');
+$light-text-secondary: map.get(lightTheme.$light-theme, 'text-secondary');
+$light-text-error: map.get(lightTheme.$light-theme, 'text-error');
+
+//button
+$light-button-tertiary: map.get(lightButton.$light-button-token-overrides, 'button-tertiary');
+$light-button-danger-primary: map.get(lightButton.$light-button-token-overrides,'button-danger-primary');
+$light-button-danger-hover: map.get(lightButton.$light-button-token-overrides,'button-danger-hover');
```
MCatherine1994, line 30: Hi Jota, let's check these variables if we actually use them anywhere? If not then we don't need to define them.

J0taFerreira: hello Catherine, I just removed the unused ones. Thanks!

MCatherine1994 (approved):

Issue #813, opened 2023-08-22
Investigate:  Frontend - Build Message: scss didn't resolve at build time


ianliuwk1019:
**Describe the bug**
This is an observation found for frontend build.
Not sure since when (and why) there are such messages (multiple lines)
```
vite v4.3.9 building for production...
transforming...

../assets/font/BCSans-Regular.woff referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime
```
from frontend build, and is it normal or is there implication that we should concern with. Didn't see them previously.

**To Reproduce**
This can be observed locally or at pipeline.
* Go to "frontend" directory.
* Do `npm run build` to build production package.
* Observe terminal message during build.
* Will see these:
```
> vite build

vite v4.3.9 building for production...
transforming...

../assets/font/BCSans-Regular.woff referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime

../assets/font/BCSans-Italic.woff referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime

../assets/font/BCSans-Bold.woff referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime

../assets/font/BCSans-BoldItalic.woff referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime

../assets/font/BCSans-Regular.woff2 referenced in /home/runner/work/nr-forests-access-management/nr-forests-access-management/frontend/src/components/Landing.vue?vue&type=style&index=0&scoped=976083d0&lang.scss didn't resolve at build time, it will remain unchanged to be resolved at runtime

[...]

PR #824, opened 2023-08-25
fix: #813 scss didn't resolve at build time

J0taFerreira:
* fixed vite not resolving nr-fsa-theme/assets/font on build time


ianliuwk1019:
Please see my comment on this ticket. @J0taFerreira

MCatherine1994:
Hi Jota! I took a deep look into this issue. And I think the error was not from our FAM repo. The problem is from this "body-type" import inside the landing page stylesheet in the common repo. Previously we were having the font path issue, so we decide to do all font changes locally in FAM. But we forgot to clean the import here, this body-type import actually is not in use. If we could remove it from the common repo, we won't have any errors on our side any more. 
Screen Shot 2023-08-28 at 11 46 36 AM

J0taFerreira:
The problem should be fixed in the common repo, no need for this pr anymore

Issue #827, opened 2023-08-25
FAM logout not working and lead to browser error overflow.


ianliuwk1019:
**Describe the bug**
Using the FAM 'Sign out' option causes browser to become unresponsive

**To Reproduce**
1. Login to FAM
2. Click on the user icon in the upper right corner
3. Click on 'Sign out'
4. Browser unresponsive

**Expected behavior**
User is signed out

**Screenshots**

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaDArIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--5f408c5925d87930ef172631b7341332095017b4/image.png)

When click 'Exit page':

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaDgrIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--564e1142a264921d7793ab4e39921fd1706e01d3/image.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Chrome, Firefox

**Additional context**

gormless87:
Looks good, works as as expected now.

PR #828, opened 2023-08-25
fix: #827  logout causing type error

J0taFerreira:
* fixed logout causing a type error leading to browser error overflow.


ianliuwk1019 (approved):
The fix looks ok. But I am wondering why " {{ authService.state.value.famLoginUser?.username }}" **famLoginUser** could be empty/underfined after login so you need to use "?" check. Maybe issue resided in deeper logic, but I will approve this for a quick fix.

Issue #773, opened 2023-08-11
Segregation by subnet


gormless87:
**Describe the task**
Create a new subnet for each deployment unit and lock down any unnecessary network traffic into or out of the subnet.

**Acceptance Criteria**
- [ ] subnet created for each deployment unit
- [ ] 
- [ ] third

**Additional context**
- Design recommendation [here ](https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Improving+Security#:~:text=Risk%20%2316%20%E2%80%93%20Shared%20Subnets)
- Or here

OlgaLiber2:
@catherinecarlson will merge

PR #841, opened 2023-08-30
feat: Custom network security groups to segregate lambdas #773

basilv:


MCatherine1994:
Just put a note here based on our walk through from last Friday so we won't forget, later on if we have time could refactor the "fam_app_sg" section to use the latest format as the "fam_data_sg".

infrastructure/server/network_security_groups.tf
```diff
@@ -0,0 +1,121 @@
...
+    tags = {
+        Name = "fam_app_sg"
+        managed-by = "terraform"
+    }
+
+    ingress {
+        from_port = 0
+        to_port = 0
+        protocol = "-1"
+        cidr_blocks = ["10.10.32.0/20", "10.10.128.0/20"]
```
MCatherine1994, line 16: just for curious Basil @basilv, I wonder how we know which ips we need to block? are these known bad ips?

terraform/tools/terragrunt.hcl
```diff
@@ -17,10 +17,10 @@ generate "tools_tfvars" {
   oidc_idir_idp_client_id = "fsa-cognito-idir-dev-4088"
   oidc_idir_idp_issuer = "https://dev.loginproxy.gov.bc.ca/auth/realms/standard"
   oidc_bceid_business_idp_client_id = "fsa-cognito-b-ce-id-business-dev-4090"
-  aws_security_group_data = "Data_sg"
+  aws_security_group_data = "fam_data_sg"
```
MCatherine1994, line 20: I wonder where we set the "Data_sg" and "App_sg" before? or are they default AWS security groups? And where could I find the network security in AWS console?
Screen Shot 2023-09-11 at 10 55 43 AM

Last question, do we change this for dev/test/prod as well?

MCatherine1994: Answer from the documentation: "Data_sg" and "App_sg" are ASEA-supplied network security groups

infrastructure/server/network_security_groups.tf
```diff
@@ -0,0 +1,121 @@
...
+resource "aws_security_group" "fam_data_sg" {
+    name = "fam_data_sg"
+    description = "FAM custom security group for data tier."
+    vpc_id = data.aws_vpc.selected.id
+    revoke_rules_on_delete = true
+    tags = {
+        Name = "fam_data_sg"
+        managed-by = "terraform"
+    }
+}
```
ianliuwk1019, line 55: Hi Basil, just curious... and probably you did this for a reason...
Why does ingress/egress rules were defined in the same block in "**fam_app_sg**" resource but rules are being separated from "**fam_data_sg**" resource?
Was just two different styles or maybe some difficulties you found?

infrastructure/server/network_security_groups.tf
```diff
@@ -0,0 +1,121 @@
...
+
+resource "aws_vpc_security_group_ingress_rule" "fam_data_sg_east_west" {
+  security_group_id = aws_security_group.fam_data_sg.id
+  referenced_security_group_id = aws_security_group.fam_data_sg.id
```
ianliuwk1019, line 59: Just to confirm, because I see these two id values are **the same** "**security_group_id**" vs "**referenced_security_group_id**", I am guessing Terraform using this "rule resource" syntax to referencing to itself to effectively means "**allow all traffic/ports, and all regions**" as well. (including west/east)?

infrastructure/server/network_security_groups.tf
basilv: the separate rules is the preferred style, but I didn't discover that initially. Plus for fam_data_sg I needed a rule to reference the same security group, not sure if that's possible with the inline format.

infrastructure/server/network_security_groups.tf
basilv: security_group_id is the parent security group that this rule belongs to.
referenced_security_group_id is the source security group's traffic (as the source) that this rule applies to.

Compare to the other rule for fam_data_sg that is referencing the fam_app_sg security group.

basilv: And yes, for other comment in teams - I likely forgot to update the other environments (dev/test/prod) terragrunt files like I did for tools.

MCatherine1994 (approved):

ianliuwk1019 (approved):

Issue #805, opened 2023-08-17
Update Ops guide to include audit functionality


gormless87:
**Describe the task**
Update the [FAM Operations](https://apps.nrs.gov.bc.ca/int/confluence/x/mKa6C) guide to include pertinent information about audit logs. 

--PO note: we may decide to have a standalone audit doc and reference it from the Ops guide

**Acceptance Criteria**
- [ ] description of log format, storage location, retention, access control, 
- [ ] description of how to export audit logs to third party in a secure manner
- [ ] 

**Additional context**
- See  #561 for initial audit log implementation
- See [STRA risk item #7](https://apps.nrs.gov.bc.ca/int/confluence/x/PKa9CQ)

MCatherine1994:
Added description for our current audit log: https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/FAM+Operations

Not sure what "access control" means in the acceptance criteria, and not sure where we want to export the audit logs?

MCatherine1994:
Found a small bug that, currently when we remove a user role assignment, the audit log shows the event type is "CREATE_USER_ROLE_ACCESS", will push a fix change it to "REMOVE_USER_ROLE_ACCESS"

PR #857, opened 2023-08-31
fix: #805 fix audit event type for remove user role access

MCatherine1994:
- fix audit event type for remove user role access #805

### DIFF
--- a/server/backend/api/app/routers/router_application.py
+++ b/server/backend/api/app/routers/router_application.py
@@ -14,7 +14,7 @@ router = APIRouter()
 
 @router.get("", response_model=List[schemas.FamApplication], status_code=200)
 def get_applications(
-    response: Response,
+    # response: Response,
     db: Session = Depends(database.get_db),
     access_roles: dict = Depends(jwt_validation.get_access_roles)
 ):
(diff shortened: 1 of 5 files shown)


ianliuwk1019 (dismissed):

server/backend/requirements.txt
```diff
@@ -1,6 +1,7 @@
 boto3==1.28.24
 botocore==1.31.24
 fastapi==0.78.0
+httpx==0.24.1
```
MCatherine1994, line 4: Added this package as required from v0.87.0

server/backend/api/app/routers/router_application.py
```diff
@@ -24,8 +24,15 @@ def get_applications(
...
-        response.status_code = 204
+
+    # from fastapi v0.79.0, setting status_code to 204, 304, or any code below 200 (1xx) will remove the body from the response
+    # so when return response with status code 204, response will have no content return to user
+    # if we want to use status code 204 for empty case, we might want to apply to other methods as well
+    # and it will impact some test cases, casue the response cannot be conver to json in empty case
+
+    # if len(query_data) == 0:
+    #     response.status_code = 204
+
```
MCatherine1994, line 35: This is the only method set status code to 204 when no applications to return. I comment this out, and let it still return an empty string. But if we still want to use 204 with an empty body, we might want to apply to other methods as well to keep consistency and update the tests.

MCatherine1994:
Closed as duplicate as pull request #862

Issue #715, opened 2023-07-20
Component/integration test for Landing Page


MCatherine1994:
**Describe the task**
Create an integration test for the landing page. Need some learning of the testing library.

**Acceptance Criteria**
- [ ] test the functionality of each element in the landing page
- [ ] code coverage increase for the landing page component

PR #861, opened 2023-09-01
feat: #715 component integration test for landing page

NickSaglioni:
Adding integration test for Landing page


frontend/src/components/Landing.vue
```diff
@@ -22,15 +22,17 @@ import { IconSize } from '@/enum/IconEnum';
                 <Button
                     class="landing-button p-button-primary"
                     label="Login with IDIR"
-                    @click="authService.methods.login"
+                    data-testid="idir-button"
```
MCatherine1994, line 25: I feel there is no need to call it "data-testid", we can just pass an id. Shall we call them "login-idir-button" and "login-bceid-button"?

NickSaglioni: Sure, it works as well. I'll update as suggested. Thanks.

MCatherine1994 (approved):

Issue #848, opened 2023-08-30
Profile doesn't looks correct when sign out


MCatherine1994:
**Describe the task**
When click on the "Sign out" button to log out, instead of log the user out immediately, the username on the profile will disappear, and it paused for a second, and then log the user out. 

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdlZUIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--086d606ca7e9bacaffd52335e21b125aadd46df9/Screen%20Shot%202023-08-30%20at%2010.36.22%20AM.png)

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdmhUIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--7304fa704880c0cf8cce696f372d743e31e50f45/Screen%20Shot%202023-08-30%20at%2010.37.14%20AM.png)

**Acceptance Criteria**
- [ ] fix the username disappear behaviour in the profile section when logout

PR #865, opened 2023-09-05
fix: #848 user name disappearing when log out

J0taFerreira:
- fixed user name disappearing when log out was clicked
- button now have loading state
- button refactored to use our reusable button component

### DIFF
--- a/frontend/src/components/common/Button.vue
+++ b/frontend/src/components/common/Button.vue
@@ -1,16 +1,24 @@
-<script setup>
+<script setup lang="ts">
 import Button from 'primevue/button';
+import type { PropType } from 'vue';
+import { IconPosition } from '../../enum/IconEnum';
 const props = defineProps({
     label: {
         type: String,
         default: 'Click',
     },
+    iconPos: {
+        type: String as PropType<IconPosition>,
+        default: IconPosition.right,
+    },
 });
 </script>
 <template>
-    <Button class="nr-button"
-        ><span>{{ props.label }}</span> <slot class="icon"></slot
-    ></Button>
+    <Button class="nr-button">
+        <slot class="icon" v-if="iconPos === IconPosition.left"></slot>
+        <span>{{ props.label }}</span>
+        <slot class="icon" v-if="iconPos === IconPosition.right"></slot>
+    </Button>
 </template>
 <style lang="scss">
 .nr-button {
--- a/frontend/src/components/common/Button.vue
+++ b/frontend/src/components/common/Button.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import Button from 'primevue/button';
 import type { PropType } from 'vue';
-import { IconPosition } from '../../enum/IconEnum';
+import { IconPosition } from '@/enum/IconEnum';
 const props = defineProps({
     label: {
         type: String,
--- a/frontend/src/components/common/ProfileSidebar.vue
+++ b/frontend/src/components/common/ProfileSidebar.vue
@@ -67,6 +67,7 @@ const buttonLabel = computed(() => {
                 :label="buttonLabel"
                 :iconPos="IconPosition.left"
                 @click="logout"
+                :disabled="loading ? true : false"
             >
                 <Icon
                     icon="user--follow"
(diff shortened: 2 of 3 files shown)


frontend/src/components/common/Button.vue
```diff
@@ -1,16 +1,24 @@
-<script setup>
+<script setup lang="ts">
 import Button from 'primevue/button';
+import type { PropType } from 'vue';
+import { IconPosition } from '../../enum/IconEnum';
```
ianliuwk1019, line 4: Use better reference than "../..".

frontend/src/components/common/Button.vue
```diff
@@ -1,16 +1,24 @@
...
 });
 </script>
 <template>
-    <Button class="nr-button"
-        ><span>{{ props.label }}</span> <slot class="icon"></slot
-    ></Button>
+    <Button class="nr-button">
+        <slot class="icon" v-if="iconPos === IconPosition.left"></slot>
+        <span>{{ props.label }}</span>
+        <slot class="icon" v-if="iconPos === IconPosition.right"></slot>
```
ianliuwk1019, line 20: Can "v-else" work?

frontend/src/components/common/Button.vue
```diff
@@ -1,16 +1,24 @@
...
+import { IconPosition } from '../../enum/IconEnum';
 const props = defineProps({
     label: {
         type: String,
         default: 'Click',
     },
+    iconPos: {
+        type: String as PropType<IconPosition>,
+        default: IconPosition.right,
+    },
```
MCatherine1994, line 13: I'm not so sure why we need this iconPos property?

frontend/src/components/common/Button.vue
MCatherine1994, line 20: not so sure why we need these slot element?

frontend/src/components/common/Button.vue
J0taFerreira: Hello, Catherine! The sign out button has the icon on a before the text, to use this component I did this so we could just pass the side we want the icon

frontend/src/components/common/ProfileSidebar.vue
```diff
@@ -121,14 +136,25 @@ import { IconSize } from '@/enum/IconEnum';
...
-    color: #000;
+    background-color: transparent !important;
+    color: $text-secondary !important;
     cursor: pointer;
 }
 
 .sign-out:hover {
     background-color: transparent;
     background-color: #ffffff;
+    color-scheme: $text-secondary !important;
```
ianliuwk1019, line 147: I am not sure why this is "color-scheme", did you mean to use "color"?
Without this added, does that work?

frontend/src/components/common/ProfileSidebar.vue
```diff
@@ -121,14 +136,25 @@ import { IconSize } from '@/enum/IconEnum';
...
+}
+
+.sign-out:focus {
+    background-color: transparent;
+    background-color: #ffffff;
+    color-scheme: $text-secondary !important;
```
ianliuwk1019, line 153: Same comment above.

frontend/src/components/common/Button.vue
J0taFerreira: the slot is the icon, if the button receive left as props, the first slot (icon on will be before the text)  will be the one visible, else the second slot (icon will be after the text) will be the one visible.

frontend/src/components/common/Button.vue
MCatherine1994: Oh, I got why we need it, for the icon

frontend/src/components/common/Button.vue
MCatherine1994: yes, got it, thanks for the explanation

MCatherine1994 (dismissed):

ianliuwk1019 (changes requested):

frontend/src/components/common/Button.vue
J0taFerreira: I tried to use a v-else but it gave the " v-else/v-else-if has no adjacent v-if or v-else-if " typescript com·plaint.

frontend/src/components/common/ProfileSidebar.vue
J0taFerreira: it works the same but i'll use color to be more clear.

frontend/src/components/common/ProfileSidebar.vue
J0taFerreira: same comment as above

frontend/src/components/common/Button.vue
ianliuwk1019: Hmm... a bit surprised vue does not have this support.
We can only leave it as v-if then.

ianliuwk1019 (approved):

Issue #839, opened 2023-08-29
Fix browser console error for grant access form


MCatherine1994:
**Describe the task**
Getting this error from Chrome debugger console, need to check the form label. This error doesn't impact the functionality, but we better fix it.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcEZNIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--3279f3701dbe8e5699c5a699171d93c662a3b42b/Screen%20Shot%202023-08-29%20at%2011.43.49%20AM.png)

**Acceptance Criteria**
- [ ] fix the label id error for grant access form

PR #866, opened 2023-09-05
fix: #839 fix browser console error for labels

J0taFerreira:
- fixed console error for labels not having a match id.

One of the ways primevue recommends for labeling it's inputs is to use aria-labelledby with a matching span as label.
exemple:
`<span id="rb2">Two</span>
<RadioButton aria-labelledby="rb2" />`


frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -205,9 +205,9 @@ function roleSelected(evt: any) {
                 <form id="grantAccessForm" class="form-container">
                     <div class="row">
                         <div class="form-group col-md-3 px-0">
-                            <label for="domainInput" class="control-label"
-                                >Select user's domain</label
-                            >
+                            <p class="control-label-radio">
+                                Select user's domain
+                            </p>
```
ianliuwk1019, line 210: May it's not related to this ticket issue, but what's the main reason to change from "\<label>" tag to "\<p>" tag with custom class? Why not "\<div>" or "\<span>" tag, just curious.

Also is the intention for newly created custom css "**control-label-radio**" class only for this component?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -205,9 +205,9 @@ function roleSelected(evt: any) {
                 <form id="grantAccessForm" class="form-container">
                     <div class="row">
                         <div class="form-group col-md-3 px-0">
-                            <label for="domainInput" class="control-label"
```
ianliuwk1019, line 208: I can't find where the class "**control-label**" came form.
Perhaps this is old Bootstrap 3 class, not sure if for Bootstrap 5 this is "**[form-control](https://getbootstrap.com/docs/5.0/forms/form-control/)**".

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -246,16 +246,16 @@ function roleSelected(evt: any) {
...
-                                <span class="text-danger"> *</span></label
+                                <span class="text-danger"> *</span></span
                             >
                         </div>
 
                         <div class="row mt-0">
                             <div class="col-md-3 px-0">
                                 <Field
-                                    id="userIdInput"
+                                    aria-labelledby="userIdInput"
```
ianliuwk1019, line 258: As you remove the "id" from <Field>, does this input needs an "id"?
Should there be an "id" (or similar id attribute from PrimeVue) at line 265 "\<InputText>"?

I am not sure the original issue was to do with using "**aria-labelledby**", maybe it was just complaining about previous "\<label>" field (with for=) does not find a matching "id" input field. So perhaps the id was missing in line 265 is the cause?

ianliuwk1019 (changes requested):

ianliuwk1019:
I seem to see these (not error, but warning) still, seems to be related:
image

image

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Hello, Ian! Thanks for the review!
Well, it can be a div or a span. Probably make more sense to use a span and use just one class.

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: the primevue give us the following accessibility options: via label tag combined with id or using aria-labelledby, aria-label. The label tag combined with id show the error which was described  on the task

ianliuwk1019: Are you sure it is the **("label tag combined with id show the error")**? My understanding is combination "<label for=>" with form field's "id" are basic html usage. So with the original errors and warnings on console the cause seems to be that the "id" is placed at the wrong place, since <Field> isn't a normal html/vue component, but a validation library component, so probably it does not render id.
Could you try to verify this? Thank you.

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: If that "control-label" does not do anything, could you do a quick search and remove them (or replace them with "[form-control](https://getbootstrap.com/docs/5.0/forms/form-control/)" from Bootstrap 5) from our FAM frontend? Please run and verify after change. Thank you Jota.

ianliuwk1019 (changes requested):

MCatherine1994:
I don't feel we need these changes.... 

If we check the primevue component:
- for InputText, it has example how to use the label: https://primevue.org/inputtext/#floatlabel, so basically we need to give an id to the "InputText" element, not the "Field" element
- for Dropdown, it has example as well: https://primevue.org/dropdown/#floatlabel
- for RadioButton, there is no example, but I think we can just use one of the id

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -209,224 +209,210 @@ function roleSelected(evt: any) {
             <div class="row">
                 <form id="grantAccessForm" class="form-container">
                     <div class="row">
-                        <div class="form-group col-md-3 px-0">
-                            <label for="domainInput" class="control-label"
-                                >Select user's domain</label
+                        <label> Select user's domain </label>
```
MCatherine1994, line 212: The "for" attributes for the label must be the id of the [input, textarea, button, select, or object elements](https://stackoverflow.com/questions/18945774/validating-html-the-for-attribute-of-the-label-element-must-refer-to-a-form-con). Here we could set the label with "for='idirSelect' or for='becidSelect'", use one of the inputId of he RadioButton component, but I feel it's also okay to just leave it as this. We basically just need a title here, and just use the same label styling.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -209,224 +209,201 @@ function roleSelected(evt: any) {
...
+                                v-model="formData.userId"
+                                v-slot="{ field }"
                             >
-                        </div>
-
-                        <div class="row mt-0">
-                            <div class="col-md-3 px-0">
-                                <Field
+                                <InputText
                                     id="userIdInput"
```
MCatherine1994, line 249: For all the InputText field, we can add an id to it, and let the label for attributes use the InputText id.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -209,224 +209,201 @@ function roleSelected(evt: any) {
...
                             <UserIdentityCard
                                 :userIdentity="verifiedUserIdentity"
                             ></UserIdentityCard>
                         </div>
                     </div>
 
-                    <div class="form-group col-md-3 px-0">
-                        <label for="roleSelect" class="control-label px-0"
+                    <div class="row">
+                        <label
```
MCatherine1994, line 308: The primevue dropdown component doesn't support the id well, even I tried to give an id to the dropdown component, the label still shows the error message. Will remove the "for" attribute, just leave it as this.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -209,224 +209,201 @@ function roleSelected(evt: any) {
...
-                                    v-slot="{ field }"
-                                >
-                                    <InputText
-                                        :placeholder="
-                                            formData.domain === 'I'
-                                                ? 'Type user\'s IDIR'
-                                                : 'Type user\'s BCeID'
-                                        "
-                                        :validateOnChange="true"
-                                        class="w-100 p-inputtext"
```
MCatherine1994, line 277: For all primevue components, by default it already has a classname, we don't need to assign the default classname again. So the "p-inputtext" classname is already there by default, no need to add it again.

MCatherine1994:
I also did a cleanup and restructure of the elements in the grant access form component. When write the page elements, make sure we use the same pattern for the sections. For example, for this grant access form, each field should share the similar structure. For each field, it will be inside a div element with the classname "row" for margin. And then inside each div, it will be label and whatever form element. It will make our code more clear.

Screen Shot 2023-09-05 at 4 17 48 PM

Screen Shot 2023-09-05 at 4 18 25 PM

MCatherine1994 (dismissed):

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: Catherine already pushed some fixes for this.
That's basically the idea for <label> and input field id match I was getting you to.
Resolve this comment.

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: In this case for trying to resolve radios' group title, your previous change make sense except on using <p> tag for me.
But Catherine has adjust this and leave a comment. Please see her reasoning.
I feel it is acceptable. So I will mark this as resolved.
Later if we face the same and feel it is annoying, we can discuss and create a custom class again.

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: The two radios already have labels associated with the inputs ids previously, so this is like you said a group "title" for the radios. Previously Jota was trying to create a custom class for this special case, but I feel your reasoning on removing "for" attribute is reasonable in this special case.

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: I believe Catherine already took care of that

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: just a question. The No label associated with a form field possible improvements is not a problem?

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Removing it will not have an impact on accessibility?

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Agreed!

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: No, failed to check missed one, see line 44 on github:
image

J0taFerreira: oh, I'll gladly take care of that.

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: It will show a warning as well, but not an error. We could improve in the future if we have a better idea, but at this moment, I feel there's no need to define a custom class just because of this problem. Thanks!!

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: I think we can remove all the class here `class="control-label label px-0"`, please verify remove these won't impact anything, thanks!

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: should be okay

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: removed

ianliuwk1019 (approved):

PR #869, opened 2023-09-06
fix: #848 profile signout incorrect

J0taFerreira:
- fixed the error causing the pipeline to fail.
- updated the color variable

It's still unclear to me why it fail since the iconSize was using practically the same implementation.


MCatherine1994:
Looks like " iconPos" is a specific name, change it to any other name works

ianliuwk1019:
> Looks like " iconPos" is a specific name, change it to any other name works

Indeed, a strange case found!

ianliuwk1019 (approved):

Issue #751, opened 2023-08-08
Further restrict permissions of api lambda db user


basilv:
When working on #718, as a workaround to keep existing unit tests passing, permissions of the API lambda db user where made slightly broader than necessary (in V30__api_lambda_db_user_restrict_access) - extra delete permissions were granted on fam_user, fam_role, and fam_forest_client.

Objective of this issue is to eliminate the need for delete privilege for testing purposes.

This ticket is to adjust the unit tests so that these extra delete permissions can be removed. Some options:
1. have tests use database transactions that auto-rollback at the end, hopefully avoiding the need to manually delete records. (This won't work if testing code that explicitly commits, unless mocking out the commit logic.)
2. don't do any cleanup and change tests to accommodate this (within a single test run - across test runs I assume a fresh database is created each time). This might make tests more brittle though.
3. Run tests as master postgres user (but this will cause permission issues to not be caught during unit testing)
4. 

AC: Strive for option 1 above:
[  ] - change test to not delete records
[  ] - db migration to prune privileges

ianliuwk1019:
Just a developer note. 
Based on V30__ script, if permission is revoked, tests currently run will fail on:
```
================================================================= short test summary info ==================================================================
FAILED testspg/crud/test_crud_forest_client.py::test_create_forest_client - sqlalchemy.exc.ProgrammingError: (psycopg2.errors.InsufficientPrivilege) permission denied for table fam_forest_client
FAILED testspg/crud/test_crud_forest_client.py::test_get_forest_client - sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transacti...
FAILED testspg/crud/test_crud_forest_client.py::test_find_or_create - sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transacti...
FAILED testspg/crud/test_crud_role.py::test_get_roles - sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transacti...
FAILED testspg/crud/test_crud_role.py::test_get_role - sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transacti...
[...]

PR #870, opened 2023-09-06
feat: #751 further restrict api lambda user

ianliuwk1019:
From last V30__ flyway script, temporary DELETE permissions were given to API lambda db user granted on "fam_user", "fam_role", and "fam_forest_client". This pr fix tests and removes the extra DELETE grant.

- Function "**delete_user**" is removed from crud_user.py since now api lambda db user has no DELETE granted and the function was not actually being used (except was for test clean up).
- Fix tests: to use "**db_pg_session**" instead so each test gets rollback automatically and no need to clean up data.
- Add new **V31__** flyway script to revoke **DELETE** permission granted for API lambda db user on (fam_user, fam_role, fam_forest_client)


basilv (approved):
Took a quick look.

server/backend/testspg/crud/test_crud_user_role.py
```diff
@@ -102,15 +98,15 @@ def test_get_use_role_by_user_id_and_role_id(db_pg_connection: Session):
...
     found_user_role_xref = crud_user_role.get_use_role_by_user_id_and_role_id(
-        db_pg_connection,
+        db_pg_session,
         TEST_USER_ID,
         TEST_FOM_DEV_REVIEWER_ROLE_ID,
     )
     assert found_user_role_xref is not None
 
     # cleanup
     crud_user_role.delete_fam_user_role_assignment(
```
MCatherine1994, line 108: just curious, do we still need to call this delete method? or it can auto rollback?

server/backend/testspg/crud/test_crud_user_role.py
```diff
@@ -69,31 +65,31 @@ def test_create(db_pg_connection: Session):
...
```
MCatherine1994, line 75: same question here

server/backend/testspg/crud/test_crud_user_role.py
ianliuwk1019: Looks like I have missed tow of them. Thanks for spotting that.

MCatherine1994 (approved):

Issue #831, opened 2023-08-26
Dashboard Remove User Role Confirm Dialog Bug


ianliuwk1019:
**Describe the bug**
This is a rare case for request handling.
On Dashboard, user data table, when user clicks "**Remove**", if there is a network error or 403 not authorized error during requesting, a success message is showing up on dashboard but it shouldn't and the row disappears while the role assignment did not get deleted from backend.

**To Reproduce**
1. When on Dashboard, stop backend (mimic traffic issue) or edit backend logic to give an 403 error.
2. Click on "Remove" button on Dashboard.
3. Confirm dialog pops up and user clicks "Remove" to confirm.
4. Will see success message but Network tab on developer tool shows network issue or 403 error.

(For easy testing on deployment environment, the tester can try to **delete self role** and expect to see toast message pops up)

**Expected behavior**
* Should expect a toast error message instead.
* No success message on Dashboard.

**Screenshots**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcFUvIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--8d8661295d83696195e9f38f286db57ca2e674a9/image.png)

Remove self (correctly handled at backend with error returned, but frontend does not handle correctly), it leads to confusion:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcDFNIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--76bba252b6e77d8d847ff13619cfb3bb9a29ee3a/image.png)
 

**Additional context**
- Can be reproduced in local.

PR #871, opened 2023-09-07
fix: #831 remove user role confirm dialog bug

ianliuwk1019:
- Fix logic for not waiting for backend response and pops up incorrect success message.
- Fix exception was lost within PrimeVue Confirm Dialog and cannot be caught by Vue:
  (The internal cause was unknown, but likely out of our hand, somehow vue lost the chain of exception
   from the Confirm Dialog library)
  * Migrate previous logic out of the Confirm Dialog success handler.
  * Move Confirm Dialog into UserDataTable (it is a more appropriate place).
  * Emit delete event for parent Dashboard to do the backend deletion call and handle the rest,
    so exception can be properly caught and display toast message.


frontend/src/components/dashboard/Dashboard.vue
```diff
@@ -1,35 +1,26 @@
...
 import { ApiServiceFactory } from '@/services/ApiServiceFactory';
 import {
-    applicationsUserAdministers,
-    isApplicationSelected,
-    selectedApplication,
-    setSelectedApplication,
-    selectedApplicationDisplayText,
+applicationsUserAdministers,
+isApplicationSelected,
+selectedApplication,
```
MCatherine1994, line 14: maybe we need the space for better formatting?

MCatherine1994:
looks good to me, thanks Ian!!

frontend/src/components/dashboard/Dashboard.vue
ianliuwk1019: Ah~ interesting, that's due to auto formatting from vscode.
I don't think I have setup the format you guys have....
Thanks!

MCatherine1994 (approved):

Issue #867, opened 2023-09-05
FAM support Silva: return user email information in Cognito


MCatherine1994:
**Describe the task**
Team Silva wants to get the email information of the login user, we need to include that in the return data from Cognito. Update the read_attributes in the `infrastructure/server/oidc_clients_silva.tf` file to include email.

**Acceptance Criteria**
- [ ] Return email of the login user for Silva

MCatherine1994:
Need to make sure in our PIA or STRA, we're allowed to share these information to downstream applications

PR #872, opened 2023-09-07
fix: #867 add user email return to silva application

MCatherine1994:
- add user email return to silva application refs: #867


ianliuwk1019:
Looks good, but a question, I am not sure if the "**email**" in the "read_attributes" also is needed to be in "**write_attributes**" (compare to previous "custom:idp_display_name", it is also in **write_attributes**).

ianliuwk1019 (approved):

Issue #874, opened 2023-09-08
Frontend - Improvement on Unnecessary Try-Catch logic


ianliuwk1019:
Some frontend components functions seem to have unnecessary try-catch logic that in the catch block it does not even handle anything, which seems a bit strange. Not sure why, probably inherit from previous coding, but I feel we probably can get rid of most of them to make code more organized and compacted.
For example, the catch block only just to return a Promise.reject().
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcmlCIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e905b02f7ac57cb96daace6e2962ed3ba049e0e4/image.png)

Target Intention:
- Try to get ride of unnecessary try-catch (if possible)
- App still functions correctly: on flow and on error-handling.

ianliuwk1019:
Also take this opportunity to improve "loading" flag to be centralized with http-interceptors, since it is very closed related to current try-catch issue.

PR #875, opened 2023-09-11
fix: #874 frontend improvement no try catch

ianliuwk1019:
- Fix/Remove unnecessary try-catch in components.
- Organize some import using better reference.
- Create LoadingState store for central loading.
- Trigger "startLoading", "stopLoading" and "stopLoadingOnError" at both new request/response HttpInterceptors.
- Add loadingLabel at Button component and implement feature for Button to take care of label change when loading.


MCatherine1994 (approved):

MCatherine1994:
Looks good to me, thanks!!

PR #876, opened 2023-09-12
spike: 742 per-route guard exemple

J0taFerreira:
- Per-route guard on grantAccess page
- Used setTimeout to make sure that it would stay on dashboard until the function finished running since it was so quickly that was almost unnoticeable.

side note: Primevue have components such as ProgressSpinner and blockUi that might come in handy while the guard is doing it's function.

I made the simplest guard exemple a could and I'm available to answer any questions.

### DIFF
--- a/frontend/src/router/GrantAccessGuards.ts
+++ b/frontend/src/router/GrantAccessGuards.ts
@@ -1,0 +1,61 @@
+import { toRaw, ref } from 'vue';
+import authService from '@/services/AuthService';
+import { ApiServiceFactory } from '@/services/ApiServiceFactory';
+
+import {
+    domainOptions,
+    getGrantAccessFormData,
+    grantAccessFormData,
+    resetGrantAccessFormData,
+} from '@/store/GrantAccessDataState';
+
+import { selectedApplication } from '@/store/ApplicationState';
+
+import type { FamApplicationRole } from 'fam-api';
+
+let applicationRoleOptions: FamApplicationRole[];
+const loadingData = ref(false);
+const apiServiceFactory = new ApiServiceFactory();
+const applicationsApi = apiServiceFactory.getApplicationApi();
+const roles = toRaw(authService.state.value.famLoginUser!.roles);
+console.log(roles);
+
+const defaultFormData = {
+    domain: domainOptions.IDIR,
+    userId: '',
+    forestClientNumber: '',
+    role_id: null as number | null,
+};
+
+const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
+
+function resetForm() {
+    resetGrantAccessFormData();
+    formData.value = defaultFormData;
+}
+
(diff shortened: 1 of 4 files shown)


frontend/src/router/GrantAccessGuards.ts
```diff
@@ -0,0 +1,61 @@
...
+        } catch (error: unknown) {
+            return Promise.reject(error);
+        } finally {
+            console.log(loadingData.value);
+            next();
+        }
+    }, 3000);
+};
+
+export { getApplicationRoles, resetForm, applicationRoleOptions, loadingData };
```
ianliuwk1019, line 61: Just quick comments:
- In general my thought was the "router guard" which can do variety of things (like auth, navigation, meta info, validation etc...) is typically some system or generic stuff. So I would name the file a generic name. Unless you really have specific requirement for particular route/component.
- I believe the setTimeOut isn't necessary, probably because you did not use "async" at "beforeEnter".
- I already refactor try-catch, you should be able to take advantage of that.
- The "export" bunch of functions that are mostly focusing on component level logic, like formData etc, and logadingData look strange in the design. Better not to mix responsibilities.

frontend/src/router/GrantAccessGuards.ts
```diff
@@ -0,0 +1,61 @@
...
+    setTimeout(async () => {
+        try {
+            applicationRoleOptions = (
+                await applicationsApi.getFamApplicationRoles(
+                    selectedApplication.value?.application_id as number
+                )
+            ).data;
+
+            if (grantAccessFormData.value) {
+                formData.value = getGrantAccessFormData();
```
MCatherine1994, line 48: I feel these parts about assigning data no needs to be in the guards

ianliuwk1019 (commented):
Thank you for the example Jota.
I only gave quick comments.
I feel this probably need better design.

frontend/src/router/GrantAccessGuards.ts
J0taFerreira: Sure, I was just trying to do the simplest guard possible as an exemple

frontend/src/router/index.ts
```diff
@@ -45,16 +46,21 @@ const routes = [
         name: 'grant',
         meta: {
             title: 'Grant Access',
-            layout: 'ProtectedLayout'
+            layout: 'ProtectedLayout',
+            isRequiredAuth: true,
```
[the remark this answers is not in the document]
J0taFerreira: this isRequiredAuth was a meta that got leftover from some testing I was doing

frontend/src/router/GrantAccessGuards.ts
J0taFerreira: Thanks! I'll keep those in mind!

MCatherine1994:
We are done with this spike ticket, close this pr. A follow up ticket will handle the actual implementation.

PR #881, opened 2023-09-13
fix: Freeze important pydantic and SQLAlchemy dependencies version.

ianliuwk1019:
Freeze two important dependencies (app uses these) version, so they are consistent for installation on different systems to prevent breaking version installation for some local development.
Based on Catherine's working local version:
	pydantic==1.10.4
	SQLAlchemy==1.4.41

Also commit a branch for fixes for both dependencies on V2: fix/backend-dependencies-fix-pydantic2-sqlalchemy2

### DIFF
--- a/server/backend/readme.md
+++ b/server/backend/readme.md
@@ -115,7 +115,7 @@ If you have multiple python projects locally and you want to isolate your FAM de
 cd server/backend
 python3 -m venv venv
 . ./venv/bin/activate
-pip3 install -r requirements.txt
+pip install -r requirements.txt
 python serverstart.py
 ```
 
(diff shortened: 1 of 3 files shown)


server/backend/readme.md
```diff
@@ -128,7 +136,7 @@ Assuming you've already installed the dependencies from requirements.txt, you ju
 
 ```
 cd server/backend
-pip install -r requirements-dev.txt
+pip3 install -r requirements-dev.txt
```
MCatherine1994, line 139: I think it's okay to leave pip install, if the user only installed python3, pip and pip3 will do the same I believe

MCatherine1994 (approved):

server/backend/readme.md
ianliuwk1019: Thanks, I changed it back, although I thought they are probably the same, I am not entirely clear how python/python3/pip/pip3 work in Python world, confusing.

MCatherine1994 (approved):

Issue #849, opened 2023-08-31
Frontend: Adjust styling in Manage Permissions page (previously dashboard page)


NickSaglioni:
**Describe the task**
Adjust styling for the Manage Permissions page (previously dashboard page) per improved design.

**Acceptance Criteria**
- [x] Update referentes and naming from dashboard to manage permissions
- [x] Adjust table style
- [x] Adjust dropdown style
- [x] Display an icon and text according to the improved design when no application is selected, like the picture below:

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBc1JhIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e7ea0955a8d9b4d050c26898b658f4dfa3ba49ae/Screenshot%202023-08-31%20at%2013.46.41.png)

**Additional context**
Figma Design: https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=1773-102878&mode=design&t=V8XSihxV8ArvRHmU-0

ianliuwk1019:
When looking at PR (#890) for this ticket, the design for the user-role table shows these columns with no data (see highlight in red):
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBbEN0IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--50396208199def8023f32029b4ddb2e71b56404d/image.png)

These information (in **red**: "First name", "Last name", "Email") my understanding is from IDIM external search. Have we discussed and know in what strategy for how we obtain this information for the frontend (fetching externally for each user, caching, combinations, or else)?
We should have backend ready before showing these columns I think?
@OlgaLiber2 , @basilv  , @MCatherine1994

OlgaLiber2:
I think @MCatherine1994 will know what we are missing on the backend to be able to display this information.

ianliuwk1019:
I see the Figma design has lower case for ("**User name**", "**First name**", "**Last name**"). I see more common for using **U**pper case for the column header (all words: "**U**ser **N**ame", "**F**irst **N**ame", "**L**ast **N**ame"), should we use that?
@OlgaLiber2

ianliuwk1019:
Hi @J0taFerreira ,
[...]

PR #895, opened 2023-09-18
feat: 851 search bar in the manage permissions page

J0taFerreira:
- Moved  add user button (former Grant new access button) to the data Table.
- Search bar are now using the new design


MCatherine1994 (dismissed):

ianliuwk1019:
Hi Jota,
The button look pretty big and the search field "Search by keyword" is not the same height with the button.
Also there is a gap in between search and the column header row.
The button corner also look a bit strange, maybe discuss with designer if that's how it looks?
Could you adjust them? Thank you.
image

MCatherine1994:
yeah I think with the stylesheet in the common repo, it looks better

ianliuwk1019:
> > Hi Jota, The button look pretty big and the search field "Search by keyword" is not the same height with the button. Also there is a gap in between search and the column header row. The button corner also look a bit strange, maybe discuss with designer if that's how it looks? Could you adjust them? Thank you. image
> 
> Hello! the styling for the search field "Search by keyword" is on the common repo but I'll take a second look.

Oh, so the "search" field is in common repo, got it. My concern is strange looking on the button. So, maybe adjust button is easier than changing primevue style in common repo and leave it as is?

MCatherine1994:
we already published a new version for the new style, the new package version is changed in #849

J0taFerreira:
> > > Hi Jota, The button look pretty big and the search field "Search by keyword" is not the same height with the button. Also there is a gap in between search and the column header row. The button corner also look a bit strange, maybe discuss with designer if that's how it looks? Could you adjust them? Thank you. image
> > 
> > 
> > Hello! the styling for the search field "Search by keyword" is on the common repo but I'll take a second look.
> 
> Oh, so the "search" field is in common repo, got it. My concern is strange looking on the button. So, maybe adjust button is easier than changing primevue style in common repo and leave it as is?

I talked to Alison about the button styling and he said that since the search action has nothing to do with the button it should be the way it is on the design.

MCatherine1994 (approved):

Issue #722, opened 2023-07-25
Replace AWS certificates


gormless87:
**Describe the task**
TLS Certificate used RSA 2048 bit with SHA-256 too weak. Replace existing certificate in AWS Certificate Manager with one using ecdsa-256 algorithm.

**Acceptance Criteria**
- [ ] Terraform config updated
- [ ] ecdsa-256 based certificate in place in a rolling fashion DEV/TOOLS/TEST/PROD via deployment
- [ ] switch tools environment to use fam-tools.nrs.gov.bc.ca URL for cloudfront

**Additional context**
- Add any other context about the task here.
- Or here

PR #947, opened 2023-10-04
fix: #722 Update terraform CloudFront to use the new certificate

MCatherine1994:
update terraform cloudfront to use the new ECDSA P 256 certificate #722

looks like tools CloudFront already setup to use fam-tools.nrs.gov.bc.ca URL


terraform-frontend/tools/terragrunt.hcl
```diff
@@ -8,6 +8,6 @@ generate "tools_tfvars" {
   disable_signature = true
   contents          = <<-EOF
   cloudfront_vanity_domain = "fam-tools.nrs.gov.bc.ca"
```
MCatherine1994, line 10: ooks like tools CloudFront already setup to use fam-tools.nrs.gov.bc.ca URL

basilv (approved):

Issue #853, opened 2023-08-31
Frontend: Adjust Forest Client number input for multiple values


NickSaglioni:
**Describe the task**
Allow the Forest Client number input for multiple values and add to the forest client card.

**Acceptance Criteria**
- [x] The forest client number input accepts multiple values
- [x] The forest client card shows the added values
- [x] The `submit application` button is only enabled when there are only active forest client numbers
- [x] Display error message through the field input validation when trying to add an invalid or inactive forest client number

**Additional context**
Figma design: https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=1773-109273&mode=design&t=maQZcd3NsABNp8hp-0

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaDV0IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--2b643d0d7cafd1b8e120e9c03b96969180873a10/Capture-user%20access2.PNG)

NickSaglioni:
There has been some changes in the design for this task, the inactive items should display an error message through the input form validation and not be added to the list of numbers of the Forest Client Card.

MCatherine1994:
Hi @NickSaglioni, please merge the latest from main and improve the style from there, thanks!!

NickSaglioni:
The current backend API implementation only accepts one client ID per request, so there are two possible solutions for submitting multiple client IDs:

1. The backend API is adjusted to accept multiple IDs per request or a new separated API is created to allow this;
2. The frontend can loop through all the items in the client ID list and send one request per client ID through the current backend API;

After discussion, we have decided for now to proceed with the option 2. To accomplish that, the complexity of task has increased, because the frontend would have to handle the multiple API requests, including handling errors and it would also require adjustments on the notification component to display the IDs that were successfully added to one user and another one to display the IDs that were not added. As further discussed with the UX team, there would be also an warning notification for some special cases (trying to add an association that already exists, for instance), so this task should include this design below for the notification messages:

[...]

PR #956, opened 2023-10-09
feat: #853 forest client number multiple values

NickSaglioni:
Updating the forest client number to accept multiple value and populating the forest number card. The submission of the multiple values is being handled by the frontend as the backend API only accepts one forest client number per request. The successful and failed client numbers are displayed at the manage access screen by separated notification messages. Trying to add a forest client that has already been added shows a warning notification as designed in Figma.


MCatherine1994:
Hi Nick, I think you want to check the changes in the package-lock.json file, it should only has the version change for the nr-theme package. Now when I run npm install and run our application locally, many of the styles don't look correct. 

For example the login buttons on the landing page should next to each other:
Screen Shot 2023-10-11 at 10 46 13 AM

and the manage access page:
Screen Shot 2023-10-11 at 10 51 10 AM


not sure if it's just me though, thanks!!

MCatherine1994:
Sometimes if backend is slow or somehow, the button doesn't show loading

Screen Shot 2023-10-11 at 11 06 18 AM

frontend/src/components/grantaccess/ForestClientCard.vue
```diff
@@ -2,44 +2,63 @@
...
+import { FamForestClientStatusType } from 'fam-api';
 
 const props = defineProps({
     forestClientData: {
-        type: Object as PropType<FamForestClient>,
+        type: Object as PropType<FamForestClient[]>,
     },
 });
+
+const activeClientCheck = ref<boolean>(true);
```
MCatherine1994, line 15: I wonder if we already check if the client number is active or not before show this card, do we still need check it again here? In the GrantAccess file, we have the "forestClientInactive" for the same purpose I believe? and if the forest client is inactive, it won't be listed in the card?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -34,32 +33,32 @@ import {
...
     userId: '',
-    forestClientNumber: '',
+    forestClientNumberList: [],
+    forestNumberError: false,
     role_id: null as number | null,
 };
 const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
+
+const forestClientInactive = ref<boolean>(false);
+const forestClientDuplicate = ref<boolean>(false);
```
MCatherine1994, line 48: as another comment, we might need another variable here "forestClientInvalid", in the case the forest client number is not found

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -34,32 +33,32 @@ import {
...
 } from 'fam-api';
 
+import { IconSize } from '@/enum/IconEnum';
+
 const defaultFormData = {
     domain: domainOptions.IDIR,
     userId: '',
-    forestClientNumber: '',
+    forestClientNumberList: [],
+    forestNumberError: false,
```
MCatherine1994, line 42: this "forestNumberError" I'm not seeing we use it anywhere?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -139,9 +134,38 @@ async function verifyIdentity(userId: string, domain: string) {
...
-    forestClientData.value = (
-        await forestClientApi.search(forestClientNumber)
-    ).data;
+    const forestNumbers = forestClientNumber.split(/[ ,]+/);
+
+    for (const item of forestNumbers) {
+        if (isNaN(parseInt(item))) return;
+        await forestClientApi.search(item).then((result) => {
+            if (
+                result.data[0].status?.status_code ===
```
MCatherine1994, line 143: I think this is where the error comes from for invalid forest client. If not found, we need to handle it differently

frontend/src/components/grantaccess/GrantAccess.vue
NickSaglioni: Good idea, that makes sense.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -307,84 +339,117 @@ function roleSelected(evt: any) {
...
-                                v-bind:disabled="
-                                    formData.forestClientNumber?.length < 8 ||
-                                    LoadingState.isLoading.value
-                                "
-                            >
-                            </Button>
+                                >
+                                    {{
+                                        `Client ID ${
+                                            formData.forestClientNumber
```
MCatherine1994, line 414: I assume we should be able to see the the forest client number which is inactive or duplicate, but seems like this data is none, it didn't show as expected
Screen Shot 2023-10-11 at 1 48 55 PM

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -307,84 +339,117 @@ function roleSelected(evt: any) {
...
+                                                ? ' has already been added.'
+                                                : ''
+                                        }`
+                                    }}
+                                </small>
+                            </div>
+                            <div class="pr-1 col-2">
+                                <Button
+                                    outlined
+                                    aria-label="Verify forest client number"
```
MCatherine1994, line 428: maybe the aria-label and label should matched?

frontend/src/store/GrantAccessDataState.ts
```diff
@@ -1,35 +1,30 @@
...
 export const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend when backend has the endpoint.
 
-export const grantAccessFormData = ref<FamUserRoleAssignmentCreate | null>();
+export const grantAccessFormData = ref<FamUserRoleAssignment | null>();
 export const grantAccessFormRoleName = ref<string | null>();
 
 export const FOREST_CLIENT_INPUT_MAX_LENGTH = 8;
 
-export const setGrantAccessFormData = (formData: any) => {
+export const setGrantAccessFormData = (formData: any, forestClientList: any) => {
```
MCatherine1994, line 11: forestClientList should have a type

frontend/src/store/NotificationState.ts
```diff
@@ -1,6 +1,13 @@
 import { reactive } from "vue"
 
-export const useNotificationMessage = reactive({
+export const successNotificationMessage = reactive({
+    notificationMsg: '',
```
MCatherine1994, line 4: maybe we can have one state for notification message as well?

frontend/src/components/grantaccess/ForestClientCard.vue
NickSaglioni: This was implemented before the scope changed to not include inactive clients to the card. You are right, now it won't be needed anymore.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -139,9 +134,38 @@ async function verifyIdentity(userId: string, domain: string) {
...
+            } else {
+                forestClientInactive.value = true;
+            }
+        });
+    }
+
+    formData.value['forestClientNumber'] = '';
+}
+
+function addForestClientNumber(forestClientNumber: FamForestClient) {
```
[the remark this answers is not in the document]
NickSaglioni: I am doing a refactor on this method, it would change.

frontend/src/components/grantaccess/GrantAccess.vue
NickSaglioni: Fixed. Thanks.

frontend/src/store/GrantAccessDataState.ts
NickSaglioni: Right, it should be an array of FamForestClient. I'll correct it. Thanks.

frontend/src/components/common/Button.vue
```diff
@@ -66,7 +66,11 @@ function onClicked(event: any) {
     </Button>
 </template>
 <style lang="scss">
-
+.nr-button {
+    display: flex;
```
MCatherine1994, line 70: Hi Nick, this line we already removed in Jota's last pr, we don't need it anymore.

MCatherine1994:
While I was typing, I got this error. I guess it complains about the "," before I entered the 2nd forest client number.
Screen Shot 2023-10-19 at 4 12 32 PM

And when I click add client for "00001012, ", it throws the error for invalid client number. 
Screen Shot 2023-10-19 at 4 13 18 PM

MCatherine1994:
Is there a way we can display all the error messages below? In this case two of them are duplicate, one is invalid. 
Screen Shot 2023-10-19 at 4 16 39 PM

NickSaglioni:
> Is there a way we can display all the error messages below? In this case two of them are duplicate, one is invalid.
> Screen Shot 2023-10-19 at 4 16 39 PM

I am trying to think about a solution to this but so far it is getting quite complex

frontend/src/components/common/Button.vue
NickSaglioni: This has been removed. Thanks.

Issue #938, opened 2023-10-02
Restrict access of custom app and data network security groups


basilv:
**Describe the task**
For the fam_app_sg and fam_data_sg security groups further restrict access as follows:
- Remove all the shared services inbound access rules from both groups
- Remove unlimited outbound rule from the fam_data_sg group

**Acceptance Criteria**
- [x] Terraform updated
- [x] Test to confirm no impacts to app functionality
- [x] Update Confluence page https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Network+Architecture
- IMPORTANT - Manual changes needed to deploy to each environment - see comment below.
- IMPORTANT - Unable to remove unlimited outbound rule - it is required (e.g. to talk to AWS Secrets Manager)
**Additional context**

basilv:
May need to change fam_app_sg definition in terraform to explicitly list separate rules, as it appears that the implicitly defined rules being removed aren't actually being deleted in AWS.

basilv:
Switching to explicit rule for fam_app_sg requires first manually deleting existing outbound role.
Also, this still doesn't clean up the inbound rules we want deleted, so that needs to be manually done as well.

basilv:
IMPORTANT: Manual AWS changes needed to successfully deploy:
1. Delete fam_app_sg outbound rule (will be recreated by Terraform on deploy, but tracked in a way that Terraform can manage)
2. Delete fam_app_sg inbound rules (Terraform can't clean these up on its own).

PR #960, opened 2023-10-12
fix: #938 Remove unnecessary flows from nsgs based on security reviews.

basilv:


ConradBoydElliottGustafson (approved):
It looks like this code is doing what it is supposed to be doing, and the Confluence page is updated. Having said that, I'm not sure I understand all this networking stuff well enough to catch any mistakes! Approved with a warning ;-)

Issue #365, opened 2023-01-13
Remove code for unused endpoints


gormless87:
**Describe the task**
Code for unused API endpoints (not part of MVP, not currently secured) has been commented out. This task is to remove the unused endpoint code completely (best practices, avoid Sonar Cloud smell complaints).

**Acceptance Criteria**
- [x] Only the secured MVP endpoints remain
- [x] recreate api-client for frontend

**Additional context**
- Use GitHub to resurrect removed code when the related endpoints come into scope.
- Ticket should be limited to only remove endpoints from "router" code (not CRUD code).

gormless87:
@MCatherine1994, @ianliuwk1019 -Can you confirm whether or not this issue is still relevant?

ianliuwk1019:
There was a problem during recreation of the **api-client** (under `**client-code-gen**`). It is an unexpected tricky issue resulting from some libraries **not having well supported openapi(swagger) version (spec version 3.1.0)**.
The main error due to the openapi spec V3.1.0 with the generator produced below error during api-client generation (although client code is generated):
```
[main] WARN  o.o.codegen.DefaultCodegen - Parameter name not defined properly. Default to UNKNOWN_PARAMETER_NAME
[main] ERROR o.o.codegen.DefaultCodegen - Not handling class PathParameter {
    class Parameter {
        name: application_id
        in: null
        description: null
        required: null
        deprecated: null
        allowEmptyValue: null
        style: simple
        explode: false
        allowReserved: null
        schema: null
        examples: null
        example: null
        content: null
        $ref: null
    }
    in: path
    required: true
} as Body Parameter at the moment
```
The generated api-client will contains this `UNKNOWN_PARAMETER_NAME`.

Secondly, in the past, due to some problem, we needed to upgrade backend FastAPI to `fastapi==0.100.0`. With this upgrade, FastAPI produced openapi(swagger) with default spec version **V3.1.0**; and this V3.1.0 is not well-supported yet from the openapi generator. Downgrade FastAPI might work, but we can't do that. I was able to find a temporary solution based on this reference (see the answer from FastAPI author Tiangolo): https://github.com/tiangolo/fastapi/discussions/9789#discussioncomment-6355603

[...]

PR #968, opened 2023-10-12
fix: #365 Remove Unused Endponts.

ianliuwk1019:
- Backend: Remove unused endpoints for router_user and router_role.
- Frontend: remove unused FAMUsersApi and FAMRolesApi from ApiServiceFactory.
- Bump openapitools/openapi-generator-cli to v7.0.1
- Backend: FastAPI temporary adjustment for swagger generation with app.openapi_version = "3.0.3" and document openapi spec issue causing generated code not working.
- "gen" folder, is generated.
- Manually fixed openapi.json regarding to swagger spec type incorrect on "some" endpoints but currently has no proper solution after investigation.


client-code-gen/gen/api/famapplications-api.ts
```diff
@@ -71,11 +72,11 @@ export const FAMApplicationsApiAxiosParamCreator = function (configuration?: Con
         /**
          * gets the roles associated with an application  :param application_id: application id :param db: database session, defaults to Depends(database.get_db)
          * @summary Get Fam Application Roles
-         * @param {number} applicationId 
+         * @param {any} applicationId
```
MCatherine1994, line 75: Hi Ian, why we change the type for applciationId from number to any?

MCatherine1994:
The rest looks good to me, thanks so much!!

client-code-gen/gen/api/famapplications-api.ts
ianliuwk1019: Hi Catherine, the code is generated based on backend swagger spec, it is strange why it is changed. Let me try to figure it out.

ianliuwk1019: Hi Catherine, thanks for spotting this. Indeed, it is a strange case that some does get generated correctly but few don't from backend FastAPI.
I took some time to downgrade and upgrade but none can solve this issue. For now, as discussed, I manually added the missing schema type into problem endpoints and add section for note to the README.
Please have a look again. Thank you!!

MCatherine1994 (approved):

Issue #978, opened 2023-10-18
Babysitting - Renovatebot and Dependabot


ianliuwk1019:
To deal with Renovate PRs and Dependabot warning currently sitting on our FAM repo in the sprint:
Dependabot Alert:
- [x] [Babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code](https://github.com/bcgov/nr-forests-access-management/security/dependabot/122)
- [ ] ~~[Vulnerable OpenSSL included in cryptography wheels](https://github.com/bcgov/nr-forests-access-management/security/dependabot/120)~~ This is same as Renovate 923 below.

Renovate PRs:
- [x] [chore(deps): update dependency cryptography to v41.0.4 [security]](https://github.com/bcgov/nr-forests-access-management/pull/923)
- [x] [chore(deps): update dependency sqlalchemy to v2](https://github.com/bcgov/nr-forests-access-management/pull/894)
- [x] [chore(deps): update dependency pydantic to v2](https://github.com/bcgov/nr-forests-access-management/pull/893)

ianliuwk1019:
The update from Dependabot "@babel/traverse@7.23.2" is on the .lock file.
This has been done and merged.
However, FAM does not directly use this package, it's mainly from aws-amplify and one other package.
Next time the Dependabot issue may come up again when we update these two packages. (aws-amplify is particularly slow on these package upgrade stuff).

ianliuwk1019:
For Renovate update on sqlalchemy to v2 and pydantic to v2, they are required to change code base. Renovate PR will be closed, instead, I will create another PR to address these two in (branch: chore/978-pydantic-sqlalchemy-upgrade)

PR #894, opened 2023-09-17
chore(deps): update dependency sqlalchemy to v2

bcgov-devops:
This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| [SQLAlchemy](https://www.sqlalchemy.org) ([changelog](https://docs.sqlalchemy.org/en/latest/changelog/)) | major | `==1.4.41` -> `==2.0.22` |

---

### Configuration

📅 **Schedule**: Branch creation - At any time (no schedule defined), Automerge - At any time (no schedule defined).

🚦 **Automerge**: Disabled by config. Please merge this manually once you are satisfied.

♻ **Rebasing**: Whenever PR is behind base branch, or you tick the rebase/retry checkbox.

🔕 **Ignore**: Close this PR and you won't be reminded about this update again.

---

 - [ ] If you want to rebase/retry this PR, check this box

---

This PR has been generated by [Renovate Bot](https://togithub.com/renovatebot/renovate).

PR #893, opened 2023-09-17
chore(deps): update dependency pydantic to v2

bcgov-devops:
This PR contains the following updates:

| Package | Update | Change |
|---|---|---|
| [pydantic](https://togithub.com/pydantic/pydantic) ([changelog](https://docs.pydantic.dev/latest/changelog/)) | major | `==1.10.4` -> `==2.4.2` |

---

### Release Notes

<details>
<summary>pydantic/pydantic (pydantic)</summary>

### [`v2.4.2`](https://togithub.com/pydantic/pydantic/blob/HEAD/HISTORY.md#v242-2023-09-27)

[Compare Source](https://togithub.com/pydantic/pydantic/compare/v2.4.1...v2.4.2)

[GitHub release](https://togithub.com/pydantic/pydantic/releases/tag/v2.4.2)

##### Fixes

-   Fix bug with JSON schema for sequence of discriminated union by [@&#8203;dmontagu](https://togithub.com/dmontagu) in [#&#8203;7647](https://togithub.com/pydantic/pydantic/pull/7647)
-   Fix schema references in discriminated unions by [@&#8203;adriangb](https://togithub.com/adriangb) in [#&#8203;7646](https://togithub.com/pydantic/pydantic/pull/7646)
-   Fix json schema generation for recursive models by [@&#8203;adriangb](https://togithub.com/adriangb) in [#&#8203;7653](https://togithub.com/pydantic/pydantic/pull/7653)
-   Fix `models_json_schema` for generic models by [@&#8203;adriangb](https://togithub.com/adriangb) in [#&#8203;7654](https://togithub.com/pydantic/pydantic/pull/7654)
-   Fix xfailed test for generic model signatures by [@&#8203;adriangb](https://togithub.com/adriangb) in [#&#8203;7658](https://togithub.com/pydantic/pydantic/pull/7658)

##### New Contributors

-   [@&#8203;austinorr](https://togithub.com/austinorr) made their first contribution in [#&#8203;7657](https://togithub.com/pydantic/pydantic/pull/7657)
-   [@&#8203;peterHoburg](https://togithub.com/peterHoburg) made their first contribution in [#&#8203;7670](https://togithub.com/pydantic/pydantic/pull/7670)

### [`v2.4.1`](https://togithub.com/pydantic/pydantic/blob/HEAD/HISTORY.md#v241-2023-09-26)

[Compare Source](https://togithub.com/pydantic/pydantic/compare/v2.4.0...v2.4.1)

[...]

PR #982, opened 2023-10-19
chore: #978 pydantic sqlalchemy upgrade

ianliuwk1019:
The is for ticket _#978.
Both renovate PRs: ref #893 and ref #894 need some code changes and cannot be merged just by bump the versions.
- Upgrade package pydantic to v2.4.2.
- Upgrade package SQLAlchemy to v2.0.22.
- Code changes accordingly.
- Fix tests accordingly.


server/backend/api/app/crud/crud_application.py
```diff
@@ -125,7 +125,6 @@ def get_application_role_assignments(
     crossref = (
         db.query(models.FamUserRoleXref)
         .join(models.FamRole)
-        .options(joinedload('role.parent_role'))
```
MCatherine1994, line 128: Hi Ian, just curious what is this options doing previously?

ianliuwk1019: Hi Catherine, this is an very old code and the way the joinedload was used is invalid anymore for SQLalchemy v2.
I think it is just to get the join for the parent role. I am not entirely sure if it is possible to do a `joinedload` on this case but when I review this code due to the change need from v2, I see that previous code is redundant, so I removed this. The data returned does include the parent_role (if there is) before and after removing this load. This can be verified from the swagger.

MCatherine1994 (approved):

server/backend/api/app/crud/crud_application.py
MCatherine1994: Thanks Ian!

Issue #952, opened 2023-10-05
Eliminate summary page 


OlgaLiber2:
**Describe the task**
Eliminate summary page

**Acceptance Criteria**
- [x] make other adjustments accordingly

**Additional context**
- Add any other context about the task here.
- Or here
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNGdOQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--3cd3ad2e89536ae4bd3227c923706c644bd6bb17/Capture-access%20summary%20page.PNG)

Issue #844, opened 2023-08-30
Frontend: Adjust Grant Access User page


NickSaglioni:
**Describe the task**
Adjust existing components for the Grant Access User page (inputs, buttons, paddings, etc) to the improved design.

**Acceptance Criteria**
- [ ] Components are stylized and work as per the design system

**Additional context**
Figma Design: https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=1773-108689&mode=design&t=YKuzKiz3MHGyGKLU-0

gormless87:
@MCatherine1994 - looks OK to me.

PR #991, opened 2023-10-24
fix: #952 eliminate summary page

NickSaglioni:
- Removing Summary Page and all references to it including components;
- Removing grantAccessDataState because it was created just to be able to move back from summary page to grant access page keeping the form data for edition;
- Code clean up and optimizations;


frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -216,17 +209,88 @@ function areVerificationsPassed() {
...
+                `${
+                    formData.value.userId
+                } was not added with Client IDs: ${errorList.join(', ')}`
+            );
+        }
+
+        router.push('/dashboard');
+    } else {
+        const data = toRequestPayload(formData.value, null);
+        await userRoleAssignmentApi.createUserRoleAssignment(data);
```
MCatherine1994, line 285: maybe we should catch the error for this request as well? But I would suggest confirm with @OlgaLiber2 first, how we would like to handle the error case when the role already assigned to the user. For now, we handle them differently for adding a role with forest client number, and without. One uses warning message, one uses toast message. Maybe cut another ticket to confirm with Olga and make them consistent. 

Screen Shot 2023-10-24 at 3 24 45 PM
Screen Shot 2023-10-24 at 3 26 27 PM

MCatherine1994:
Thanks Nick! Overall looks good to me. Just not sure if these buttons color are same as in the design, but maybe you can check and adjust all the style in the ticket #844
Screen Shot 2023-10-24 at 3 30 08 PM

NickSaglioni:
> Thanks Nick! Overall looks good to me. Just not sure if these buttons color are same as in the design, but maybe you can check and adjust all the style in the ticket #844
> Screen Shot 2023-10-24 at 3 30 08 PM

Yes, I did not adjust the styling because it is going to be done in ticket #844. I just handled the buttons now because the text needed to be changed for this task, but the styling I was going to handle in ticket #844.

frontend/src/components/grantaccess/GrantAccess.vue
NickSaglioni: I makes sense to be the same, I'll adjust it.

MCatherine1994 (approved):

MCatherine1994:
Hi Nick, thanks! This looks good to me know, approved. Could you please create another ticket for this error case? When there is an error from backend, we display the error message in both notification and toast. 


Screen Shot 2023-10-25 at 11 21 37 AM

NickSaglioni:
> Hi Nick, thanks! This looks good to me, approved. Could you please create another ticket for this error case? When there is an error from backend, we display the error message in both notification and toast.
> 
> Screen Shot 2023-10-25 at 11 21 37 AM

Thanks Catherine! I'll open a ticket for that.

MCatherine1994:
> > Hi Nick, thanks! This looks good to me, approved. Could you please create another ticket for this error case? When there is an error from backend, we display the error message in both notification and toast.
> > Screen Shot 2023-10-25 at 11 21 37 AM
> 
> Thanks Catherine! I'll open a ticket for that.

Hi Nick, just a reminder, let's create this new ticket before the next sprint.

Issue #990, opened 2023-10-24
Update terraform to generate dev urls for Cognito client dev environment


MCatherine1994:
**Describe the task**
Write a script to generate 50 dev urls in terraform, for each Cognito dev client configuration of the downstream applications. 

**Acceptance Criteria**
- [x] Added 50 dev urls for both redirect sign in and sign out, follow their deployment url pattern
- [x] Added to dev client for forest client, spar, silva and fom. 

**Additional context**
- Script can get from last spike ticket #951

PR #996, opened 2023-10-25
feat: #990 added 50 dev urls for each cognito client

MCatherine1994:
refs: #990


infrastructure/server/oidc_clients_spar.tf
```diff
@@ -43,10 +43,10 @@ resource "aws_cognito_user_pool_client" "test_spar_oidc_client" {
   callback_urls                                 = [
     "http://localhost:3000/",
     "https://oidcdebugggersecure-c6af30-dev.apps.gold.devops.gov.bc.ca/",
-    "https://nr-spar-webapp-test-frontend.apps.silver.devops.gov.bc.ca/"
+    "https://nr-spar-test-frontend.apps.silver.devops.gov.bc.ca/"
```
ianliuwk1019, line 46: Hi Catherine, I didn't see this in ticket requirement, wondering, does SPAR request this recent change to remove `webapp` from domain?

MCatherine1994:
> Looks good!
> One question, do you want to generalize the `range(50)` counts into a terraform variable for `50` that can be used everywhere for future clients?

Hi Ian, I was thinking that perviously, but not sure if the client could want different range in the future....

infrastructure/server/oidc_clients_spar.tf
MCatherine1994: yeah, their deployment url for test and prod changed.

ianliuwk1019:
> > Looks good!
> > One question, do you want to generalize the `range(50)` counts into a terraform variable for `50` that can be used everywhere for future clients?
> 
> Hi Ian, I was thinking that perviously, but not sure if the client could want different range in the future....

I think most likely not, but it is a variable that has default value 50. And with that the other part of this issue (github workflow change with Derek), I assume 50 is already sufficiently the upper bound. If client has specific needs, it can certainly use other number.

ianliuwk1019 (approved):

Issue #999, opened 2023-10-26
Fix grant access page role not render problem 


MCatherine1994:
**Describe the task**
Found that the dropdown for role section doesn't list all the roles when first reach to the grant access page. It's because we are using a local variable for the role options, and when we change its value in the onMounted method, the page won't get re-rendered. 
![image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/6604f9b7-049f-4713-a0c1-7ddaa63aba39)
![image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/54ab0357-ae68-4edf-a12b-9549f58c6dca)

Previously we change the formData value in the same onMounted method, so the page will be re-rendered because of the value change of the formData. But we removed the whole if-else section after removing the summary page, and the value change of the applicationRoleOptions can't trigger the page re-render.
![image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/fabb8ee8-ab4b-49f7-bfbf-cb72e32f9750)

**Acceptance Criteria**
- [ ] Fix the role dropdown render problem

**Additional context**
- Add any other context about the task here.
- Or here

PR #1000, opened 2023-10-26
fix: #999 fix the role not render problem on the grant access page

MCatherine1994:
refs: #999


ianliuwk1019 (approved):
Thanks for catching this Catherine!

ianliuwk1019 (approved):

Issue #843, opened 2023-08-30
Frontend: Create Progress Bar (stepping) component


NickSaglioni:
**Describe the task**
Create Progress Bar (stepping) component according to the design system;

**Acceptance Criteria**
- [ ] Progress Bar component exists and is displayed in the `Add new user` page
- [ ] Progress Bar component is stylized as per the design system
- [ ] Progress Bar component updates itself and indicates which step the user currently is

**Additional context**
Figma Design: https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=1773-108689&mode=design&t=YKuzKiz3MHGyGKLU-0

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcXhzIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--a573bcd75c785d71e882b3b326dedb14a2b9498e/image.png)

MCatherine1994:
user id changes: 
- cleanup the user role step (step 2)
- cleanup the org info step (step 3)

verifyIdentity (verify idir user):
- mark step 1 as complete

when we do role selection:
- mark step 2 as complete 
- check if need to show the step 3 based on role type

enter forest client input:
- change step3 icon

verify forest client:
- if there is a successful one, mark step3 as completed

J0taFerreira:
The Developers spoke with the PO, and the conclusion was that at this time there is no need to implement the progress bar component.

PR #1006, opened 2023-10-27
feat: #843 stepping progress bar component

J0taFerreira:
- step primevue component implemented 

https://app.zenhub.com/workspaces/fsa-fingerprint-61f04f08304d3e001a4f2578/issues/gh/bcgov/nr-forests-access-management/843

note: some styles on the primevue step component is in common repo


MCatherine1994:
Just wondering in the case that we put the user id and then select the role, but then remove the user id,  shall we remove the checkmark at the 2nd step in the progress bar? Screen Shot 2023-10-30 at 1 12 16 PM

MCatherine1994:
> > Just wondering in the case that we put the user id and then select the role, but then remove the user id, the user role field will be cleaned up as well automatically, shall we remove the checkmark at the 2nd step in the progress bar? Screen Shot 2023-10-30 at 1 12 16 PM
> 
> I think that @aliramoss might be the best person to answer that.

Hi Jota, please ignore this comment, and see my latest comment, thanks!

aliramoss:
I think that this is right, in the scenario where the user assigned an user id, then a role, then he realize that the user id is wrong, then he removes and add again, maybe the role was right. Then we might leave the field filled, and since the field is already filled, it's right to have the gray checkmark.

MCatherine1994:
I have a design question to @OlgaLiber2 and @aliramoss, it's just my personal feeling when first access to our grant access page. 

- I feel normally we have the progress bar to walk user through the steps which are not on the same page. But here all of our form fields are on the same page, I was a little bit confused at the beginning and was trying to find a "next" button to go to the next step. My confusion was solved when I started typing and I saw the progress indicator, so just a little bit concern if other people will have the same confusion as me. 

Screen Shot 2023-11-01 at 5 12 29 PM

- I'm also wondering if we really need a progress bar here, cause our form is quite easy and the information is quite straightforward, not sure if the progress bar will increase the complexity for the user?  Especially in the error case, maybe we could think of a better way to handle it in the progress bar. 
    In this case as shown in the screenshot, the user failed to add some forest client numbers but also already added a valid one, he's able to submit the application, the "submit application" button is enabled. But the progress bar shows the step 3 failed, which makes me feel the form is not finished, and based on the progress bar indicator, he shouldn't be able to submit? It might be confused. 
    Maybe we don't need to display any error message in the progress bar, and as long as there is a valid forest client number, we can mark it as complete? 

Screen Shot 2023-11-01 at 5 23 43 PM

- My last question is that, we'll disable the dropdown for role selection when the user id field is empty? It confused my a little bit, cause I was trying to select a role but it's disabled, is that possible the case that the user wants to see the role options before enter a user id? And if we disable the role selection, should the forest client number field also be disabled or got cleaned up? 

Screen Shot 2023-11-01 at 5 35 10 PM

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -76,6 +77,36 @@ const forestClientApi = apiServiceFactory.getForestClientApi();
...
+
+const orgInfo = ref({
+    label: 'Organization Information',
+    active: false,
+    isVisible: false,
+    icon: IconSteps.queue,
+    errorMessage: ''
+});
+
+stepItems.value = [
```
MCatherine1994, line 105: Hi Jota, we don't update state variable like this. Let's create a method in the store/stepState.ts file to handle the value change, something like the following and then import it to use:
```
export const setStepItems= (newStepItemsValue) => {
    stepItems.value = newStepItemsValue;
}
```
And if you're trying to initialize the stepItem, you should do in the onMounted method, or when you want to assign the value, it shouldn't be here.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -76,6 +77,36 @@ const forestClientApi = apiServiceFactory.getForestClientApi();
...
+    addUser: {
+        stepNumber: 2,
+        label: 'Add User Roles',
+        active: false,
+        icon: IconSteps.queue,
+        errorMessage: ''
+    },
+});
+
+const orgInfo = ref({
```
MCatherine1994, line 97: let's rename this to orgInfoStep

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Hello! Sure, will do!

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: sure, I'll change it.

MCatherine1994:
Hi Jota, so in the case we enter a userId and select the fom_submitter role, but later on we changed mind and cleanup the userID field. The organization information section got cleaned up, but the progress bar still shows 3 steps. Let's cleanup the 3red as well in the case we make change to the userId field. Thanks!!

Screen Shot 2023-11-08 at 3 41 43 PM

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -90,7 +114,9 @@ const isIdirDomainSelected = () => {
 
 function userDomainChange() {
     resetVerifiedUserIdentity();
```
MCatherine1994, line 116: um..... so we called this "resetVerifiedUserIdentity" here, but the userIdChange method also call the "resetVerifiedUserIdentity" method, so we don't need to call it here anymore, let's clean this up.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -76,7 +77,30 @@ const forestClientApi = apiServiceFactory.getForestClientApi();
...
 
+
+const userInfoStep = ref({
+    stepNumber: 1,
+    label: 'User Information',
+    active: false,
+    icon: IconSteps.queue,
+})
+
+const addUserStep = ref({
```
MCatherine1994, line 88: let's rename this, it's the addUserRoleStep

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -135,6 +176,13 @@ async function verifyIdentity(userId: string, domain: string) {
     verifiedUserIdentity.value = (
         await idirBceidProxyApi.idirSearch(userId)
     ).data;
+
+    if(verifiedUserIdentity.value.found) {
+        userInfoStep.value.active = true;
+        userInfoStep.value.icon = IconSteps.checkmark;
+        addUserStep.value.active = addUserStep.value.icon === IconSteps.checkmark ? true : false;
```
MCatherine1994, line 183: when verify the userId, it there a need to update the status of the 2rd or 3rd step?

frontend/src/components/common/PageTitle.vue
```diff
@@ -25,10 +42,49 @@ const route = useRoute();
     />
     <h1 class="title">{{ props.title }}</h1>
     <h2 class="subtitle">{{ props.subtitle }}</h2>
+    <Steps
+        v-if="route.meta.hasSteps && stepItems"
+        :readonly="false"
+        :model="stepItems" aria-label="Form Steps"
```
MCatherine1994, line 48: aria-label should be in a separate line

frontend/src/store/stepState.ts
```diff
@@ -0,0 +1,18 @@
+import { ref } from 'vue';
+import  type { IconSteps } from '@/enum/IconEnum';
+
+export interface IStepInfo {
+    label: string;
+    active: boolean;
+    icon: IconSteps;
+};
+
+export const stepItems = ref<IStepInfo[]>();
```
MCatherine1994, line 10: um... I think we might have a better way to implement this stepItems. 

So

1. We should call this grantAccessStepItems, because later on we could have other stepItems for other components, we need to give a name  that can explain itself
2. We should define its default value here, so 
```
const userInfoStep = ...
const userRoleStep = ...
const orgInfoStep = ...

export const grantAccessStepItems = ref<IStepInfo[]>(
 [...userInfoStep],
 [...userRoleStep]
);
```
3. Create an updateStepItems method: 
```
export const setStepItems = (index, changedStepItemsValue: Object<key: value>) => {
     stepItems.value[index] = {...stepItems.value[index], changedStepItemsValue }
 }
```
The above method will still keep the original data for that step, but change whatever needs to change. So in the grantAccess Component, whenever we need to change the grantAccessStepItems, for example, change the userRoleStep, you will call the method in a way like below. The index for the 2nd step is 1 (array index starts from 0). 
```
setStepItems(1, {active: true, icon: checkmark})
```
The label will remain the same, but active and icon will be changed. 
4. Rename the addNewStep method to grantAccessAddNewStep
5. Since we are now using the grantAccessStepItems just for the grantAccess Component. In the PageTitle component, stepItems should be passed as a parameter, because not all components have it 
Screen Shot 2023-11-08 at 4 11 29 PM

frontend/src/components/common/PageTitle.vue
```diff
@@ -11,11 +16,23 @@ const props = defineProps({
     subtitle: {
         type: String,
         required: true,
-    },
+    }
 });
 
 const route = useRoute();
 
+const isActive = (item: MenuItem) => {
```
MCatherine1994, line 24: we don't need to make an extra method for this

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -90,7 +114,9 @@ const isIdirDomainSelected = () => {
 
 function userDomainChange() {
     resetVerifiedUserIdentity();
+    userIdChange();
     formData.value.userId = '';
+    userInfoStep.value.icon = IconSteps.incomplete;
```
MCatherine1994, line 119: and this is why I recommend that we shouldn't use local step variables. Because you're sharing the stepState variable with PageTitle component, but you're changing the local step variable, not the stepState variable. Because the stepState is pointing to the reference of your local step variables, it still works, but it's really confused and hard to maintain for long term.

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: That's true, thanks for noticing!

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: that was from the older code, I'll remove.

frontend/src/store/stepState.ts
J0taFerreira: sure, we can chat about it.

frontend/src/components/common/PageTitle.vue
J0taFerreira: you're correct, it's not needed anymore. Thanks!

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: can we have a chat about it?

MCatherine1994: yeah, for sure, let's chat about it. Cause I feel probably even there is no need to define it as a global state variable, if we'll pass it as a parameter to the PageTitle component, we can just define that stepState locally in grant access component.

MCatherine1994:
Discussed within the team, we feel our grant access form is simple enough for the user to understand. We don't need to add the progress bar at this point, as it might cause more complexity to the user. Close this pr.

Issue #1017, opened 2023-11-02
Fix auth function code error which causes bcsc login failure 


MCatherine1994:
**Describe the task**
Forest client can't login using bc service card credential, and by testing using debugger tool and the cloud watch log, found a python key error in the auth function when try to write the audit log. But we still need to verify if this is the real problem, or the return data from the bcsc login changed. 

<img width="1243" alt="Screen Shot 2023-11-02 at 1 20 30 PM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/011fb33b-0eb0-4bf5-ba97-5eecab24a069">

<img width="1427" alt="Screen Shot 2023-11-02 at 1 23 02 PM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/199092aa-84b2-453f-baa8-69caa1c3e486">

**Acceptance Criteria**

- [x] Verify the cause of the problem
- [x] If it's the python key error, fix it

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
Added some finding for further investigation. According to BCSC mapping table, there is no "**custom:idp_username**" .
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMitnQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e99ae85ad00684eb58a3692ef4f43c05cfcab02e/image.png)

MCatherine1994:
Found a weird pipeline issue that is not related to this ticket, but happened when try to do dev deployment this afternoon. Just write this down in case we see the similar issue in the future. The issue can be found [here](https://github.com/bcgov/nr-forests-access-management/actions/runs/6749956121/job/18351916145). 

<img width="1130" alt="Screen Shot 2023-11-03 at 4 29 39 PM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/6a4c7463-f581-421e-8e51-0fa873b9c4a7">

Basically it complains the state lock file in the Dynamodb. Ian and I compared the state items in our AWS dev environment and AWS test environment. We found that somehow dev has an extra entry which we don't know why and how it got created. The issue was solved by remove that extra entry manually and rerun the pipeline.
[...]

Issue #1018, opened 2023-11-04
Add test for auth function for bceid and bcsc user


MCatherine1994:
**Describe the task**

- [x] Get the auth function event information for BCSC user from ticket #1017
- [x] Add the bcsc user information to the audit log, if we can't show idp_username, should we show display_name?
- [x] Add test for auth function for bceid user

  Test data for bceid user: 
"""
    { 'eventTime': '2023-11-03 22:47:07', 'logger': 'root', 'level': 'INFO', 'eventMessage': event: {'version': '1', 'triggerSource': 'TokenGeneration_HostedAuth', 'region': 'ca-central-1', 'userPoolId': 'ca-central-1_p8X8GdjKW', 'userName': 'test-bceidbusiness_81069f39b35b4861bcd010582b63b112@bceidbusiness', 'callerContext': {'awsSdkVersion': 'aws-sdk-unknown-unknown', 'clientId': '27psq1bvt0imca09bcs3g3u7vj'}, 'request': {'userAttributes': {'custom:idp_user_id': '81069F39B35B4861BCD010582B63B112', 'sub': 'bdc3cf92-f0c7-4e17-8a1f-359991136919', 'custom:idp_username': 'LOAD-2-TEST', 'custom:keycloak_username': '81069f39b35b4861bcd010582b63b112@bceidbusiness', 'cognito:user_status': 'EXTERNAL_PROVIDER', 'identities': '[{"userId":"81069f39b35b4861bcd010582b63b112@bceidbusiness","providerName":"TEST-BCEIDBUSINESS","providerType":"OIDC","issuer":null,"primary":true,"dateCreated":1699051596416}]', 'email_verified': 'false', 'custom:idp_name': 'bceidbusiness', 'custom:idp_business_id': '2DA5FA0A3E9F4A76825342BCCB861A15', 'custom:idp_display_name': 'NRS Load Test-2', 'custom:idp_business_name': 'NRS Load Test-2', 'email': 'nrpp_test@nrpp.compratech.com'}, 'groupConfiguration': {'groupsToOverride': ['ca-central-1_p8X8GdjKW_TEST-BCEIDBUSINESS'], 'iamRolesToOverride': [], 'preferredRole': None}}, 'response': {'claimsOverrideDetails': None}} }
"""

- [x] Add test for auth function for bcsc user, use the test data in task 1

    Test data for bcsc user:
"""
[...]

PR #1019, opened 2023-11-06
fix: #1017 audit bcsc

MCatherine1994:
- fix the bcsc login issue caused by the audit log #1017
- will add the unit test later

### DIFF
--- a/server/auth_function/lambda_function.py
+++ b/server/auth_function/lambda_function.py
@@ -83,9 +83,9 @@ def lambda_handler(event: event_type.Event, context: Any) -> event_type.Event:
         audit_event_log["requestingUser"]["userType"] = USER_TYPE_CODE_DICT[
             event["request"]["userAttributes"]["custom:idp_name"]
         ]
-        audit_event_log["requestingUser"]["idpUserName"] = event["request"][
-            "userAttributes"
-        ]["custom:idp_username"]
+        # audit_event_log["requestingUser"]["idpUserName"] = event["request"][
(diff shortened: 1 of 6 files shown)


server/auth_function/lambda_function.py
```diff
@@ -82,9 +83,13 @@ def lambda_handler(event: event_type.Event, context: Any) -> event_type.Event:
...
-        audit_event_log["requestingUser"]["idpUserName"] = event["request"][
-            "userAttributes"
-        ]["custom:idp_username"]
+        if (
+            audit_event_log["requestingUser"]["userType"] == USER_TYPE_IDIR
+            or audit_event_log["requestingUser"]["userType"] == USER_TYPE_BCEID_BUSINESS
+        ):
+            audit_event_log["requestingUser"]["idpUserName"] = event["request"][
+                "userAttributes"
+            ]["custom:idp_username"]
```
basilv, line 92: This should fix the issue, but doesn't actually handle logging of the BCSC identity as part of the audit record. So if you want to deploy this fix quickly, a subsequent ticket should be raised to address logging of BCSC identity.

basilv (approved):

server/auth_function/lambda_function.py
MCatherine1994: yeah, I think bcsc doesn't have the idp_username, but we might could show the display name if that makes sense. I'll verify what we can get and include in the follow up ticket #1018.

Issue #1020, opened 2023-11-06
Fix the frontend issue when assign the role


MCatherine1994:
**Describe the task**
In the case the user selected the "fom_submitter" role and entered a forest client number, but changed his mind, and change the role to "fom_reviewer". The forest client card is not got removed, and when click submit application, it's still trying to assign the fom_submitter role with that forest client number.

<img width="887" alt="Screen Shot 2023-11-06 at 11 04 42 AM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/6cbeb291-e86a-473f-bb7d-3251509b9f38">
<img width="1469" alt="Screen Shot 2023-11-06 at 11 04 49 AM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/c6713a83-21ad-4364-af34-b66a405b7e5c">

**Acceptance Criteria**
- [x] Fix the role assignment issue when change role 

**Additional context**
- Add any other context about the task here.
- Or here

gormless87:
@MCatherine1994 - looks good to me - problem solved :)

PR #1023, opened 2023-11-06
fix: #1020 fix grant access page role change issue

MCatherine1994:
refs: #1020

### DIFF
--- a/frontend/src/components/grantaccess/GrantAccess.vue
+++ b/frontend/src/components/grantaccess/GrantAccess.vue
@@ -37,7 +37,7 @@ const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend w
 const defaultFormData = {
     domain: domainOptions.IDIR,
     userId: '',
-    forestClientNumberList: [],
+    forestClientNumber: '',
     role_id: null as number | null,
 };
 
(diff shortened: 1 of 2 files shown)


ianliuwk1019 (dismissed):

ianliuwk1019 (approved):

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -114,15 +114,12 @@ function resetVerifiedUserIdentity() {
...
+    forestClientData.value = [];
     forestClientNumberVerifyErrors.value = [];
 }
 
-function resetForm() {
-    formData.value = defaultFormData;
-}
-
 function cancelForm() {
-    resetForm();
```
MCatherine1994, line 125: we don't use resetForm method anywhere else, removed it and add the reset below.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -211,16 +209,16 @@ function areVerificationsPassed() {
 
 function toRequestPayload(
     formData: any,
-    forestClientNumber: FamForestClient | null
```
MCatherine1994, line 214: it's confused to call this parameter forestClientNumber when it's not just the number, but also has the status and other information, renamed it

PR #1025, opened 2023-11-07
fix: #1020 fix forest client number add error

MCatherine1994:
refs: #1020

### DIFF
--- a/frontend/src/components/grantaccess/ForestClientCard.vue
+++ b/frontend/src/components/grantaccess/ForestClientCard.vue
@@ -114,15 +114,6 @@ p * {
     align-self: baseline !important;
 }
 
-p {
-    display: flex;
-    flex-direction: column;
-}
-
-p * {
-    align-self: baseline !important;
-}
-
 .content-wrapper {
     flex-direction: row !important;
     width: 100%;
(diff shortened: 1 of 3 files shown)


frontend/src/components/grantaccess/ForestClientCard.vue
```diff
@@ -114,15 +114,6 @@ p * {
     align-self: baseline !important;
 }
 
-p {
```
MCatherine1994, line 117: this part is duplicate, it has the same code right above.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -457,7 +463,7 @@ function removeForestClientFromList(index: number) {
                             style="width: 100% !important"
                             v-bind="field.value"
                             @update:modelValue="handleChange"
-                            @change="resetForestClientNumberData()"
+                            @change="removeForestClientSection()"
```
MCatherine1994, line 466: we need to remove the whole forest client section when the role changed, include the forest client number input field and the forest client card

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -494,7 +500,7 @@ function removeForestClientFromList(index: number) {
                                     :validateOnChange="true"
                                     v-bind="field"
                                     class="w-100 custom-input"
-                                    @input="resetForestClientNumberData()"
+                                    @input="cleanupForestClientNumberInput()"
```
MCatherine1994, line 503: we only need to cleanup the forest client number input field when input changes, but will keep the forest client card

ianliuwk1019 (approved):

Issue #1007, opened 2023-10-30
Add the new fam_app_admin table 


MCatherine1994:
**Describe the task**

- [x] Add the new fam_app_admin table in PostgreSQL
- [x] Add a new database user in PostgreSQL. For now just give it SELECT privileges on fam_app_admin.
- [ ] ~Try to use SqlACodeGen to see if we can use that tool to generate the python data model~
- [x] Manually map new table in python data model.
- [x] Add the table and attribute description, audit columns (create/update date time, create/update user as other tables)

**Additional context**
- Update the existing model to include this new table, in server/backend folder
- Reference for the last spike ticket for SqlACodeGen tool #474
- Data model design in confluence https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Admin+Segregated+Schema+Design?

ianliuwk1019:
Checked on `sqlacodegen` package for model.py generation, its current release `sqlacodegen==3.0.0rc3` does support `SQLAlchemy` version 2.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeHl5QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c041584e9dbc43f32cd13ee8f038b1889f7c67d6/image.png)
<br/>
At locally running db, executed command to reverse-generate a model.py from database:
 `sqlacodegen --generator declarative postgresql+psycopg2://postgres:postgres@localhost:5432/fam --schemas app_fam`. It executed pretty fast and it also created quite detail than previous hand-tweak model.py including all the **comments**, **constraints**, **indexes** and **relationships**!!
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBeW15QVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5c55c8db0922cb417a266eaa1082ed0c54d205b3/image.png)
<br/>
[...]

PR #1029, opened 2023-11-08
feat: #1007 new fam app admin model

ianliuwk1019:
* Add V32 flyway script for new "fam_application_admin" table.
  * Add new proxy db user in V32 flyway script and grant privilege on "fam_application_admin", "fam_application" and "fam_user" tables.
  * Add ORM mapping to the existing model.py for new table and relationship.
  * Add new environment place_holder values for flyway migration for Docker compose and for terraform.


docker-base-services.yml
```diff
@@ -38,6 +38,8 @@ services:
       - FLYWAY_BASELINE_ON_MIGRATE=true
       - FLYWAY_PLACEHOLDERS_api_db_username=fam_proxy_api
       - FLYWAY_PLACEHOLDERS_api_db_password=test
+      - FLYWAY_PLACEHOLDERS_admin_api_db_user=fam_proxy_admin_api
+      - FLYWAY_PLACEHOLDERS_admin_api_db_password=ojkr97y663kvqcwlm8hyvf8dalxd
```
ianliuwk1019, line 42: Both values are current intended for **temporary** only. I assume without giving a value the build will fail, so just gave them some value. The subsequent ticket when the user/password are setup on secret manager than these can be adjusted.

infrastructure/server/flyway.tf
```diff
@@ -231,6 +231,8 @@ data "aws_lambda_invocation" "invoke_flyway_migration" {
           "api_db_password" : "md5${md5(join("", [local.flyway_db_creds.password, local.flyway_db_creds.username]))}",
           "auth_lambda_db_user" : "${local.flyway_db_auth_creds.username}",
           "auth_lambda_db_password" : "md5${md5(join("", [local.flyway_db_auth_creds.password, local.flyway_db_auth_creds.username]))}",
+          "admin_api_db_username" : "fam_proxy_admin_api",
+          "admin_api_db_password" : "ojkr97y663kvqcwlm8hyvf8dalxd",
```
ianliuwk1019, line 235: Both values are current intended for **temporary** only. I assume without giving a value the build will fail, so just gave them some value. The subsequent ticket when the user/password are setup on secret manager than these can be adjusted.

docker-base-services.yml
basilv: This doesn't sound right. The creation of the secret in S3 I'd expect to be part of this effort? Although you are right that the api_db_password value is just a placeholder, but I suggest using the value 'test' rather than something that looks like a real password.

infrastructure/server/flyway.tf
basilv: In particular, everyone in the world knows this password now.

docker-base-services.yml
ianliuwk1019: The password is replaced with clear dummy value.

infrastructure/server/flyway.tf
ianliuwk1019: The password is replaced with clear dummy value.

docker-base-services.yml
```diff
@@ -38,6 +38,8 @@ services:
       - FLYWAY_BASELINE_ON_MIGRATE=true
       - FLYWAY_PLACEHOLDERS_api_db_username=fam_proxy_api
       - FLYWAY_PLACEHOLDERS_api_db_password=test
+      - FLYWAY_PLACEHOLDERS_admin_mgmt_api_db_user=fam_proxy_admin_mgmt_api
```
MCatherine1994, line 41: Hi Ian, just wondering do we need to use mgmt? or we can just use management?

ianliuwk1019: Hi Catherine. Thanks. I simply chose the standard abbreviation for management "mgmt" for short and recognizable naming as a preference. I don't actually mind using "management". Do you like seeing it as "admin_management_api_db_user" for this variable instead? I can change it.

MCatherine1994 (dismissed):

docker-base-services.yml
```diff
@@ -38,6 +38,8 @@ services:
       - FLYWAY_BASELINE_ON_MIGRATE=true
       - FLYWAY_PLACEHOLDERS_api_db_username=fam_proxy_api
       - FLYWAY_PLACEHOLDERS_api_db_password=test
+      - FLYWAY_PLACEHOLDERS_admin_management_api_db_user=fam_proxy_admin_management_api
```
MCatherine1994, line 41: oh btw, do we call it "FLYWAY_PLACEHOLDERS_admin_management_api_db_user" or "FLYWAY_PLACEHOLDERS_admin_management_api_db_username" ? I saw in the flyway.tf file, it's called "admin_management_api_db_username"

MCatherine1994:
also wondering do we need to add the new db user to this file? not sure what this is for.... 
Screen Shot 2023-11-09 at 4 00 08 PM

docker-base-services.yml
ianliuwk1019: Good catch, I didn't notice it. I will fix it for just "_user" not name.

ianliuwk1019:
Hi @basilv I have updated the password to be some temporary dummy value. Catherine review and approves it. Will merge this one for Catherine to continue on her task; password will be then changed.

> also wondering do we need to add the new db user to this file? not sure what this is for.... Screen Shot 2023-11-09 at 4 00 08 PM

I don't know what this file is for, never hear from others before, and I feel we don't need to change it as I don't even see lambda api here. I feel it is ok. When testing the terraform in task 2, we can see if our change is working.

docker-base-services.yml
ianliuwk1019: This is fixed. Thank you!

Issue #1015, opened 2023-11-02
[Frontend] - Notification issues when the text is too long


NickSaglioni:
**Describe the bug**
When the notification message is too big, the styling gets broken.

**Screenshots**
1.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNE9hQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--307019b31bfae7b7a292c1425be37e52ac0600b3/image.png)

2.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFNhQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--ac6696625c871b9bb83b070508398bd65636a7ef/image.png)

PR #1034, opened 2023-11-13
fix: #1015 Notification issues when the text is too long

J0taFerreira:
- added breakpoint for the ids shown
- limit the max of visible show ids to 5


frontend/src/store/screenSizeState.ts
```diff
@@ -0,0 +1,20 @@
+import { sideNavState } from "@/store/SideNavState";
+import { setNotificationMsg, newMsgList, newFormData, newSeverity } from "./NotificationState";
+
+export let screenSize = window.innerWidth;
```
MCatherine1994, line 4: Hi Jota, if you export this variable and share it with other components, it has to be a reactive variable, not a local changeable variable.  Cause whenever you change its value you want all other components receive the new value and re-render the page.

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+import { isDesktop, screenSize } from "@/store/screenSizeState";
+import type { Severity } from "@/enum/SeverityEnum";
 
 const defaultNotification = {
     success: "",
     warn: "",
     error: "",
 };
 
+export const newMsgList = ref();
```
MCatherine1994, line 17: and we can't use one reactive variable to support multiple notification message, in the case we have both successful and warning message, the auto adjustment on the notification message will fail
Screen Shot 2023-11-14 at 11 04 54 AM

J0taFerreira: Hello! I'll fix it. Thanks!

frontend/src/store/screenSizeState.ts
J0taFerreira: Sure!

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
     It is structured this way so it doesn't stack multiple messages with the same severity (by design).
     It is intended to be used in conjunction with the NotificationStack component.
 */
-import type { Severity } from "@/enum/SeverityEnum";
+
 import { ref } from "vue"
+import { isDesktop, screenSize } from "@/store/screenSizeState";
+import type { Severity } from "@/enum/SeverityEnum";
 
 const defaultNotification = {
```
MCatherine1994, line 11: So instead of creating t here new variables (newMsgList, newFormData, newSeverity), we can update the structure of our notification variable.

```
const defaultNotification = {
    success: {msg: "", forestClientNumberList: ""},
    warn: {msg: "", forestClientNumberList: ""},
    error: {msg: "", forestClientNumberList: ""},
}
```

And when you push notification, we store both message and the forest client number list:
```
export const pushNotification = (severity: Severity, message: string, forestClientNumberList: Array<string>) => {
     notifications.value[severity].msg = message;
     notifications.value[severity].forestClientNumberList = forestClientNumberList
 }
```

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+    const hiddenIdList = msgList.splice(idPerMsg());
+
+    const isPlural = msgList.length === 1 && hiddenIdList.length === 0 ? 'ID' : 'IDs'
+    const msgByType = {
+        success: `was successfully added with Client ${isPlural}:`,
+        warn: `already exists with Client ${isPlural}:`,
+        error: `was not added with Client ${isPlural}:`
+    };
+
+    const setNotificationMsg = `
```
MCatherine1994, line 72: we better not name a variable the same as a method, please rename it

frontend/src/store/NotificationState.ts
MCatherine1994: we can remove these three new variables

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+    } else if(screenSize <= 1290 && isDesktop() || screenSize <= 960) {
+        return 4;
+
+    } else {
+        return 5;
+
+    }
+};
+
+export const setNotificationMsg = (msgList: string[], formData: any, severity: Severity) => {
```
MCatherine1994, line 57: we don't use formData parameter at all, and please rename the msgList, I feel it's actually the forest client number list?

frontend/src/store/screenSizeState.ts
```diff
@@ -0,0 +1,20 @@
...
+
+window.addEventListener("resize", (event) => {
+    screenSize = window.innerWidth;
+
+    //update the message when the screensize change
+    if(newMsgList.value) {
+        setNotificationMsg(newMsgList.value, newFormData.value, newSeverity.value);
```
MCatherine1994, line 11: and here, you can create a for loop:

Object.keys(notifications.value).forEach((severity) => {
    if(notifications.value[severity].msg &&  notifications.value[severity].forestClientNumberList > 1) {
        setNotificationMsg(notifications.value[severity].forestClientNumberList, severity);
    }
})

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -28,7 +28,8 @@ import {
 
 import { IconSize } from '@/enum/IconEnum';
 import { Severity } from '@/enum/SeverityEnum';
-import { pushNotification } from '@/store/NotificationState';
+import { pushNotification, setNotificationMsg } from '@/store/NotificationState';
```
MCatherine1994, line 31: if we don't call pushNotification in this file, let's remove the import

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+    const msgByType = {
+        success: `was successfully added with Client ${isPlural}:`,
+        warn: `already exists with Client ${isPlural}:`,
+        error: `was not added with Client ${isPlural}:`
+    };
+
+    const setNotificationMsg = `
+        ${formData} ${msgByType[severity]} ${fullIdList.join(', ')}
+        ${isPlural === 'IDs' ?
+        'and ' + hiddenIdList.length + ' more...'
```
MCatherine1994, line 75: when the hiddenIdList is 0, there is no need to show "and 0 more..."

frontend/src/store/screenSizeState.ts
ianliuwk1019: Use getter if you need screenSize value.

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+        return 3;
+
+    } else if(screenSize <= 1290 && isDesktop() || screenSize <= 960) {
+        return 4;
+
+    } else {
+        return 5;
+
+    }
+};
```
ianliuwk1019, line 55: The implementation requirement isn't clear here.
Did you discuss with business (or @OlgaLiber2 / @MCatherine1994 ) in the ticket that: hiding id(s) for the user is **acceptable**? 
Is it one of the acceptance criteria in the ticket? Please add this if its been discussed.
(We don't just make code works).
We need to ask, for example:
* how many of the ids can the field take, is there such requirement?
* if there is no limit, can we fit the ids in the rendering container conditionally (css style change)?
* if there is no limit, for mobile device, what else can we do to handle it better before we really chop the ids.
Please discuss these kind of questions before implementing it or before opening a PR if you already see a problem.

ianliuwk1019 (changes requested):
Hi @Jota,
When implementing some logic, it is better to know if this logic should be in common place or it should be in specific place.
We don't want to mix common file/function/logic with business specific. It will confuse developers and resulting into hard to maintain code. Time consuming to refactor. Sometimes it is not easy but please keep that in mind.
For example in this case, the purpose of `NotificationState` is rather confusing me. I am not getting a sense it is for general purpose, but rather incorporate specific business concept embedded in there already (like using **formData**, **forestClientId**).
Also rather having `screenSizeState` state to change some notification message, it may be better to let screenSizeState alone and notify (since `ref` is reactive) whichever state or component needs to change to do its jobs.

frontend/src/store/NotificationState.ts
J0taFerreira: I'll use your suggestion

frontend/src/store/NotificationState.ts
J0taFerreira: I'll teste if it works without the formData

J0taFerreira: the formData is used to get the user id but I'll double check

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: sure, I'll remove it

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -257,30 +259,13 @@ async function handleSubmit() {
...
         if (successList.length > 0) {
-            pushNotification(
-                Severity.success,
-                `${
-                    formData.value.userId
-                } was successfully added with Client IDs: ${successList.join(
-                    ', '
-                )}`
-            );
+            setNotificationMsg(successList, formData.value.userId, Severity.success);
```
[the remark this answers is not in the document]
J0taFerreira: it takes the userId, I'll refactor to be more clear

frontend/src/store/NotificationState.ts
```diff
@@ -3,27 +3,78 @@
...
+
 import { ref } from "vue"
+import { isDesktop, screenSize } from "@/store/screenSizeState";
+import type { Severity } from "@/enum/SeverityEnum";
 
 const defaultNotification = {
     success: "",
     warn: "",
     error: "",
 };
```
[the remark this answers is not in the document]
J0taFerreira: it handle a single message

frontend/src/store/NotificationState.ts
J0taFerreira: hiding the ids that don't fit is on the design

frontend/src/store/NotificationState.ts
MCatherine1994: user id you mean the forest client number? but the msgList are for the forest client number as well?

frontend/src/store/NotificationState.ts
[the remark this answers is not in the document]
MCatherine1994: yeah it handles single message for now, I think in the future if we need to handle multiple success message we can rethink about it, but we can focus on the the issue in the ticket for now.

frontend/src/store/NotificationState.ts
J0taFerreira: the formData is the user name and the msgList it the list with the client ids

MCatherine1994: then let's just pass the user name, and name the parameter that we can understand what it's for

frontend/src/store/NotificationState.ts
```diff
@@ -3,15 +3,27 @@
...
 import type { Severity } from "@/enum/SeverityEnum";
-import { ref } from "vue"
 
 const defaultNotification = {
     success: "",
     warn: "",
     error: "",
 };
 
+export const seeAll = reactive({
```
MCatherine1994, line 15: Hi Jota, I got what you're trying to do here and I think it's a good idea to store the full message. But I wonder if we could combine the two variables "notifications" and "seeAll" together? like following? otherwise I feel it's a little bit hard to follow and understand, we should always try to keep the logic simple and reusable. Please let me know if you have any questions, thanks!!
```
const defaultNotifications = {
    success: {msg: "", fullMsg: "", showFullMsg: false},
    warn:  {msg: "", fullMsg: "", showFullMsg: false},
    error:  {msg: "", fullMsg: "", showFullMsg: false},
};

export const notifications = ref(
    JSON.parse(JSON.stringify(defaultNotification))
);

export const showFullNotificationMsg = (severity: Severity) => {
    notifications.value[severity].msg = notifications.value[severity].fullMsg
    notifications.value[severity].showFullMsg = true
}
```

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -242,44 +243,13 @@ const composeAndPushNotificationMessages = (
...
-                warningIdList[0] === ''
-                    ? `already exists with the role ${
-                          getSelectedRole()?.role_name
-                      }`
-                    : `already exists with Client IDs: ${warningIdList.join(
-                          ', '
-                      )}`
-            }`
-        );
+        setNotificationMsg(warningIdList, username, Severity.warning );
```
MCatherine1994, line 249: let's remove the space between "warning" and ")"

MCatherine1994:
the new style make the notification box really wide, is it just me? or this is expected maybe?
Screen Shot 2023-11-22 at 12 42 49 PM

frontend/src/store/NotificationState.ts
J0taFerreira: Hello, Catherine! I can do that

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: sure, will do!

MCatherine1994 (approved):

Issue #1010, opened 2023-10-30
Add an admin-management API Gateway


MCatherine1994:
**Describe the task**
 
- [x] Add an admin-management API Gateway. Map all methods to the new lambda function.
- [x] Add api-gateway url to Cognito callbackUrl.
- [x] Prepared terraform "output" for new api url.

**Additional context**
-- ~separate task to add API gateway log content to CloudWatch.~ (This is already added in #975)

MCatherine1994:
We can do the estimate after done the spike ticket, so we have a better understanding about what we need to do.

ianliuwk1019:
FYI...
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM01EQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b862d9c94f7bd378661ccb88739c774a753a331c/image.png)
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBM1FEQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--52b2e13b07646ed97b25f9c53767f1cd58424136/image.png)

ianliuwk1019:
The newly created api gateway has these two Security Hub medium/low warnings.
```
"APIGateway.3 API Gateway REST API stages should have AWS X-Ray tracing enabled"
"APIGateway.1 API Gateway REST and WebSocket API execution logging should be enabled"
```
These already has been documented in ticket #975 so there is no need to add enable logging ticket.

PR #1050, opened 2023-11-22
feat: #1010 admin api gateway

ianliuwk1019:
* A new API `admin_management_api_gateway` Gateway.
* Associate new API gateway with new `fam_admin_management_api` lambda.
* Add api-gateway url to Cognito callbackUrl.
* Prepared terraform "output" for new api url.
image

success + 200 + 13 ?!! :
image

### DIFF
--- a/infrastructure/server/admin_management_api_gateway.tf
+++ b/infrastructure/server/admin_management_api_gateway.tf
@@ -68,6 +68,8 @@ resource "aws_api_gateway_deployment" "admin_management_api_gateway_deployment" 
       aws_api_gateway_method.admin_management_api_gateway_method_proxy.id,
       aws_api_gateway_integration.admin_management_api_gateway_integration_proxy.id
     ]))
+    # https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/api_gateway_deployment mentioned
+    # using whole resouce with filesha1() that may be consider in the future.
   }
 
   # TODO: remove this after success deployment.
--- a/infrastructure/server/admin_management_api_gateway.tf
+++ b/infrastructure/server/admin_management_api_gateway.tf
@@ -4,6 +4,11 @@ resource "aws_api_gateway_rest_api" "admin_management_api_gateway_rest_api" {
   endpoint_configuration {
     types = ["REGIONAL"]
   }
+
+  tags = {
+    Name = "fam_admin_api_gateway"
+    managed-by = "terraform"
+  }
 }
 
 resource "aws_api_gateway_resource" "admin_management_api_gateway_resource_proxy" {
@@ -63,6 +68,10 @@ resource "aws_api_gateway_stage" "admin_management_api_gateway_stage" {
   deployment_id = aws_api_gateway_deployment.admin_management_api_gateway_deployment.id
   rest_api_id   = aws_api_gateway_rest_api.admin_management_api_gateway_rest_api.id
   stage_name    = "v1"
+
+  tags = {
+    managed-by = "terraform"
+  }
 }
 
 resource "aws_lambda_permission" "admin_management_api_lambda_permission" {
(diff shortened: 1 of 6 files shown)


MCatherine1994:
Hi Ian, I think we need to add the Api Gateway doc url to Cognito callback urls, something like below, so we can login to the swagger console and try our endpoint

```
callback_urls                        = "${concat(var.fam_callback_urls, ["${aws_api_gateway_deployment.fam_api_gateway_deployment.invoke_url}/docs/oauth2-redirect"])}"
```

ianliuwk1019:
> Hi Ian, I think we need to add the Api Gateway doc url to Cognito callback urls, something like below, so we can login to the swagger console and try our endpoint
> 
> ```
> callback_urls                        = "${concat(var.fam_callback_urls, ["${aws_api_gateway_deployment.fam_api_gateway_deployment.invoke_url}/docs/oauth2-redirect"])}"
> ```
> 
> And maybe we can remove the manual created one in tools space, thanks!!

That's right, I will add this and update the acceptance criteria for this ticket.
Hm... just realized there are more to it now for this api gateway..... I will create a new ticket for some other work related to this new api gateway, but the url will be added to Cognito with this ticket.
Thanks Catherine!!

MCatherine1994:
Hi Ian, I tried the smoker test router in our new Api Gateway, but I got the internal server error, I checked the CloudWatch logs and found it complains the key error... 
Screen Shot 2023-11-23 at 1 52 54 PM

ianliuwk1019:
> Hi Ian, I tried the smoke test router in our new Api Gateway, but I got the internal server error, I checked the CloudWatch logs and found it complains the key error... Screen Shot 2023-11-23 at 1 52 54 PM

Hi Catherine, I tried again with swagger, it does not have issue with 200 success from the smoke_test.
Were you test in different way?
Also, if the api-gateway reaches the lambda, then api-gateway works since its job is only a proxy passing all requests to the lambda.
Please let me know if I can try it out on your way.

infrastructure/server/admin_management_api_gateway.tf
```diff
@@ -0,0 +1,103 @@
...
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_api_gateway_stage" "admin_management_api_gateway_stage" {
+  deployment_id = aws_api_gateway_deployment.admin_management_api_gateway_deployment.id
+  rest_api_id   = aws_api_gateway_rest_api.admin_management_api_gateway_rest_api.id
+  stage_name    = "v1"
```
basilv, line 70: I always found this 'v1' stage name weird. I'm not sure what the AWS docs recommend. I have a vague recollection of talking to Conrad about this but I don't remember the details.

basilv (commented):
Left one comment, more of a question. Otherwise looks good.

infrastructure/server/admin_management_api_gateway.tf
ianliuwk1019: Thanks Basil! I was also not certain where to see the usefulness of this "stage_name" (**v1**) field come in but didn't ask this question. Simple I just reuse what it was from original first api-gateway.
I guess this has something to do with **pipeline** (dev/test/prod?) deployment but we don't really need this, since we have different AWS space for deployments; however, it is a required field.
The other guess is if we have some lambda api changes (second version **v2**) that needed to be tested on before making it official, it's easy to create different "stage_name" (v2?) for experimental and only allow for restrict testing before switching from v1 to v2 I guess.
Anyways, I just coded it the same as first api-gateway. If you feel we can make this name more meaningful I will change that name to your suggestion.
By the way, this "stage" is now as a **Terraform resource** now, rather than "**stage_name**" in api_gateway_deployment's attribute (as Terraform suggested), so that's different than the first api-gateway; and it works.

basilv: I looked into the AWS docs. To quote "a stage is a named reference to a deployment, which is a snapshot of the API. You use a Stage to manage and optimize a particular deployment". 

I wasn't able to come up with any better name, so might as well leave it v1, at least that's consistent with the other gateway.

basilv (approved):

Issue #1044, opened 2023-11-17
Frontend: Better Error Message for Self Granting User Error.


ianliuwk1019:
**Describe the bug**
When a user is granting "itself" at FAM, backend disallows for self granting and returns 403 error.
Frontend does not seem to catch this specific error, instead the error notification is showing a general message.

Additional than fixing this ticket, it might be also good to bring this ticket #820 to finalize error response behavior from the backend.

**To Reproduce**
1. Use access FAM and go to Grant Access page.
2. Assign itself any role and submit.
3. System displays error notification: see below screenshot.

**Expected behavior**
- [ ] A clear error message showing the user why the role cannot be added.

**Screenshots**
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNnJvQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--9faffa5784ac6666f62eb462aa7c473b7e910b22/image.png)

**Additional context**
- Please ask Olga/Pete for better wording on this error message.
- There are general 403 errors that could happen from system, but for this case, luckily it has error "code" (see screenshot), frontend can probably rely on this code for determine what error (message) it needs to show.

PR #1053, opened 2023-11-27
fix: #1044 Better Error Message for Self Granting User Error

J0taFerreira:
- Change the self granting message error message to be more descriptive.


ianliuwk1019 (changes requested):
Has this case been tested, when the user tries to remove self a role?
image
(The behavior is different than adding self a role.)
image

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -132,7 +132,7 @@ async def enforce_self_grant_guard(
                 status_code=HTTPStatus.FORBIDDEN,
                 detail={
                     "code": ERROR_SELF_GRANT_PROHIBITED,
-                    "description": "Altering permission privilege to self is not allowed",
+                    "description": "Granting roles to self is not allowed.",
```
ianliuwk1019, line 135: Hi Jota, you probably missed, but as already mentioned in the ticket, currently you can fix this in frontend.
We will need a good approach for both frontend/backend to show consistent messages and will refactor that time for backend if needed. (please discard both `server/admin_management/api/app/routers/router_guards.py` and `server/backend/api/app/routers/router_guards.py` change)

J0taFerreira: hello, Ian! I'll change the fix to be in frontend

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -257,9 +257,16 @@ const composeAndPushNotificationMessages = (
             getSelectedRole()?.role_name
         );
     }
-    if (errorIdList.length > 0) {
+
+    if(errorIdList.errorMsg) {
+        setNotificationMsg(
```
MCatherine1994, line 262: Hi Jota, since we handle all the notification message in the "setGrantAccessNotificationMsg" method in the "NotificationState.ts" file, maybe let's pass this error message as a parameter, and still set the notification over there.

ianliuwk1019 (changes requested):
Hi Jota, is the "remove" case fixed? Or are we leaving it for different behavior?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -271,7 +278,7 @@ const composeAndPushNotificationMessages = (
 const handleSubmit = async () => {
     const successForestClientIdList: string[] = [];
     const warningForestClientIdList: string[] = [];
-    const errorForestClientIdList: string[] = [];
+    const errorForestClientIdList = {errorMsg: '', forestClienNumber: [] as string[]};
```
ianliuwk1019, line 281: Strange to see the "**error**..." variable "type" is different than the "success/warning" ones.
So this variable was specific to "**idList**" only, now you change its definition, the variable should reflect that and not leaving it hanging for other developer to get confused.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -238,7 +238,7 @@ function toRequestPayload(
 const composeAndPushNotificationMessages = (
     successIdList: string[],
     warningIdList: string[],
-    errorIdList: string[]
+    errorIdList: {errorMsg: string, forestClienNumber: string[]}
```
ianliuwk1019, line 241: Same comment as below, for the variable naming when you change its definition.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -257,9 +257,16 @@ const composeAndPushNotificationMessages = (
...
     }
-    if (errorIdList.length > 0) {
+
+    if(errorIdList.errorMsg) {
+        setNotificationMsg(
+        Severity.error,
+        `An error has occured. ${errorIdList.errorMsg}`
+        );
+    };
+    if (errorIdList.forestClienNumber.length > 0) {
```
ianliuwk1019, line 267: The above 2 "if" statements (line 261 and line 267) look suspicious at the intention of the `const errorForestClientIdList = {errorMsg: '', forestClienNumber: [] as string[]};` usage.

It simply states like this:
```
if(errorIdList.errorMsg) {
   // call function A to handle message
}
if (errorIdList.forestClienNumber.length > 0) {
   //  call function B to handle message - with ids.
}
```
So developer may wonder are these two properties "**errorMsg**" and "**forestClienNumber**" mutually exclusive in ` {errorMsg: '', forestClienNumber: [] as string[]}`?
Is there a third condtion when both **errorMsg** and **forestClienNumber** exist? And how the code is going to react with that? From **above**, does it intend to produce "**two**" messages?
Or if it cannot happen the same time, why not just use "**if(){ // do this... } ... else{ // do that... }**" to be clear?

More over, as mentioned above why the "**error**" variable definition/structure is different than "**success/warning**" when the name of "**Notification**" function sounds generic enough? I feel adding/displaying notifications are getting confusing and might need some thought later.

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Hello, I'll do that!

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -257,9 +257,16 @@ const composeAndPushNotificationMessages = (
...
         );
     }
-    if (errorIdList.length > 0) {
+
+    if(errorIdList.errorMsg) {
+        setNotificationMsg(
+        Severity.error,
+        `An error has occured. ${errorIdList.errorMsg}`
+        );
+    };
```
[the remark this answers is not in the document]
J0taFerreira: Hello, Ian! Sure, thanks!

J0taFerreira:
> Hi Jota, is the "remove" case fixed? Or are we leaving it for different behavior?

the remove case will use the backend message but it will now be displayed as a notification message instead of toast. Catherine and I talked yesterday and I'll be doing this on this very pr

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: I'll see what can be done to make it better, thanks!

MCatherine1994:
Hi Jota, thanks this pr looks good now. Just need to check the SonarCloud code smell, I think some of them can be solved.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -240,20 +240,21 @@ const handleSubmit = async () => {
...
+                    errorNotification.msg = 'Granting roles to self is not allowed.';
                 } else {
-                    errorForestClientIdList.push(
+                    errorNotification.errorForestClientIdList.push(
                         item?.forest_client_number
-                            ? item.forest_client_number
-                            : ''
-                    );
+                            ? item?.forest_client_number
+                            : '');
```
MCatherine1994, line 249: so here it can be `item?.forest_client_number || ''`, maybe that can help resolve the code smell

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -302,16 +303,28 @@ const composeAndPushNotificationMessages = (
...
+    }
+
+    if (errorMsg.errorForestClientIdList.length > 0) {
+        setGrantAccessNotificationMsg(
+            errorMsg.errorForestClientIdList,
             username,
             Severity.error,
             getSelectedRole()?.role_name
         );
     }
```
ianliuwk1019, line 324: So if the `composeAndPushNotificationMessages()` method receives this `errorMsg` 
```
{
msg: "Yo, why don't you give me permissions?",
errorForestClientIdList: [
   "00000000",
   "99999999"
   ]
}
```
How does this "**composeAndPushNotificationMessages**()" method pops out the notifications?

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: Hello, Catherine! What code smell is this related to? Thanks for the help!

frontend/src/store/NotificationState.ts
```diff
@@ -42,9 +42,12 @@ export const setGrantAccessNotificationMsg = (
     forestClientNumberList: string[],
     userId: any,
     severity: Severity,
-    role = ''
+    role = '',
+    specificMsg = ''
 ) => {
     let notificationFullMsg = '';
+    let notificationMsg = '';
```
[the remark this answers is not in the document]
J0taFerreira: Just fixed, thanks!

frontend/src/store/NotificationState.ts
```diff
@@ -42,9 +42,12 @@ export const setGrantAccessNotificationMsg = (
     forestClientNumberList: string[],
     userId: any,
```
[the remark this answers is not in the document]
J0taFerreira: Hello, Ian! You are correct! Thanks!

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -240,20 +240,21 @@ const handleSubmit = async () => {
                             ? item?.forest_client_number
                             : ''
                     );
+                } else if (error.response.data.detail.code === "self_grant_prohibited") {
+                    errorNotification.msg = 'Granting roles to self is not allowed.';
                 } else {
-                    errorForestClientIdList.push(
+                    errorNotification.errorForestClientIdList.push(
```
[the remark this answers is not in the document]
J0taFerreira: typescript complain if it's without the "?" . Regarding the repetition, you mean that I should put the ternary  `item?.forest_client_number ?  item?.forest_client_number  : ''` on a variable ?

MCatherine1994 (dismissed):

frontend/src/components/grantaccess/GrantAccess.vue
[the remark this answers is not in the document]
ianliuwk1019: I am not sure what you meant for typescript complain; but I see this condition repetition 3 times, so why not make it easier to assign it to a variable, it is just to determine the id right?

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: Please be careful validate the code line by line.
This is what I can produce and not the message you expect to see.
image

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/GrantAccess.vue
J0taFerreira: How did you produce that?

ianliuwk1019: @J0taFerreira why not look at my original comment and this comment thread to understand how I arrive this instead? I already explain this at the beginning if you look carefully.
I used exact testing object from what I asked you before on how it would produce the message when the object is passed to your function.
We can talk after scrum if you still is not sure. @MCatherine1994

frontend/src/components/grantaccess/GrantAccess.vue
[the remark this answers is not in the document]
MCatherine1994: Hi Jota, let's create a variable and use this variable to replace "item?.forest_client_number ? item.forest_client_number : ''" where we use it at 3 locations, and test in local again. 

```
const itemForestClientNumber = item?.forest_client_number ? item.forest_client_number : ''
```

Screen Shot 2023-12-04 at 2 36 57 PM

You might want to reformat the file, I checked again and I think the format is not correct. The properties pass to a component should be in the same line, this doesn't look correct. Please check with Nick how to set the formatter if you need help! thanks!
Screen Shot 2023-12-04 at 2 41 08 PM

ianliuwk1019 (approved):

Issue #1022, opened 2023-11-06
Refactor: change all functions inside frontend component to be consistent


MCatherine1994:
**Describe the task**
We use two ways to define functions in our frontend. Let's change it to the same way to keep consistency. Let's change to use:
```
const functionName = () => {}
```

**Acceptance Criteria**
- [x] check all frontend component files, and change the function definition if needs 
- [x] refactor the grant access component, group the method based on steps, and add some comment. for example:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMnFwQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--e2d1adbe90750d12c9986cf7aa79e245e6151b58/Screen%20Shot%202023-11-06%20at%202.11.46%20PM.png)

- [x] refactor the names in grant access component, for example, the default form data should have "forestClientNumbers" not "forestClientNumber", cause we now support multiple numbers

**Additional context**
- Add any other context about the task here.
- Or here

PR #1060, opened 2023-11-28
refactor:  #1022 refactor frontend code to be consistent

MCatherine1994:
- refactor the frontend component code, to define function in a consistent way
- adjust the import statement, always put external import at the top
- auto format the component files
- cleanup unused code
- group the methods in grant access component based on form field

### DIFF
--- a/frontend/src/components/grantaccess/GrantAccess.vue
+++ b/frontend/src/components/grantaccess/GrantAccess.vue
@@ -3,20 +3,21 @@ import router from '@/router';
 import { ErrorMessage, Field, Form as VeeForm } from 'vee-validate';
 import { onMounted, ref } from 'vue';
 import { number, object, string } from 'yup';
-
-import Button from '@/components/common/Button.vue';
-import ForestClientCard from '@/components/grantaccess/ForestClientCard.vue';
-import UserIdentityCard from '@/components/grantaccess/UserIdentityCard.vue';
 import Dropdown from 'primevue/dropdown';
 import InputText from 'primevue/inputtext';
 import RadioButton from 'primevue/radiobutton';
 
+import Button from '@/components/common/Button.vue';
+import ForestClientCard from '@/components/grantaccess/ForestClientCard.vue';
+import UserIdentityCard from '@/components/grantaccess/UserIdentityCard.vue';
 import { ApiServiceFactory } from '@/services/ApiServiceFactory';
+import { IconSize } from '@/enum/IconEnum';
+import { Severity } from '@/enum/SeverityEnum';
+import { setGrantAccessNotificationMsg } from '@/store/NotificationState';
 import {
     selectedApplication,
     selectedApplicationDisplayText,
 } from '@/store/ApplicationState';
-
 import LoadingState from '@/store/LoadingState';
 import {
     FamForestClientStatusType,
@@ -26,32 +27,23 @@ import {
     type FamUserRoleAssignmentCreate,
 } from 'fam-app-acsctl-api';
 
-import { IconSize } from '@/enum/IconEnum';
-import { Severity } from '@/enum/SeverityEnum';
-import { setGrantAccessNotificationMsg } from '@/store/NotificationState';
-
 const FOREST_CLIENT_INPUT_MAX_LENGTH = 8;
-
 const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend when backend has the endpoint.
 
 const defaultFormData = {
     domain: domainOptions.IDIR,
     userId: '',
-    forestClientNumber: '', // form field input, for multiple forest client numbers
+    forestClientNumbers: '', // form field input, for multiple forest client numbers
     role_id: null as number | null,
 };
-
 const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
-
-const forestClientNumberVerifyErrors = ref([] as Array<string>);
-
 const formValidationSchema = object({
     userId: string()
         .required('User ID is required')
         .min(2, 'User ID must be at least 2 characters')
         .nullable(),
     role_id: number().required('Please select a value'),
-    forestClientNumber: string()
+    forestClientNumbers: string()
         .when('role_id', {
             is: (_role_id: number) => isAbstractRoleSelected(),
             then: () =>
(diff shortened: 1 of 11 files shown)


frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -26,32 +27,23 @@ import {
...
-
 const FOREST_CLIENT_INPUT_MAX_LENGTH = 8;
-
 const domainOptions = { IDIR: 'I', BCEID: 'B' }; // TODO, load it from backend when backend has the endpoint.
 
 const defaultFormData = {
     domain: domainOptions.IDIR,
     userId: '',
-    forestClientNumber: '', // form field input, for multiple forest client numbers
+    forestClientNumbers: '', // form field input, for multiple forest client numbers
```
MCatherine1994, line 36: rename this to forestClientNumbers as we're now supporting multiple numbers

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -85,15 +79,34 @@ onMounted(async () => {
     ).data;
 });
 
+/* ------------------ User information method ------------------------- */
```
MCatherine1994, line 82: group the methods in this file based on functionality

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -189,19 +181,30 @@ async function verifyForestClientNumber(forestClientNumber: string) {
     }
 
     //The input is updated with only the numbers that have errored out. The array is converted to a string values comma separated
-    formData.value['forestClientNumber'] = forestNumbers.toString();
-}
+    formData.value['forestClientNumbers'] = forestNumbers.toString();
+};
 
-function isForestClientNumberNotAdded(forestClientNumber: string) {
+const ifForestClientNumberNotAdded = (forestClientNumber: string) => {
```
MCatherine1994, line 187: rename this method from "isForestClientNumberNotAdded" to "ifForestClientNumberNotAdded", so the logic when call this method is easier to understand

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -267,53 +312,6 @@ const composeAndPushNotificationMessages = (
...
-        successForestClientIdList,
-        warningForestClientIdList,
-        errorForestClientIdList
-    );
-
-    router.push('/dashboard');
-};
-
-function removeForestClientFromList(index: number) {
-    forestClientNumberVerifyErrors.value = [];
```
MCatherine1994, line 314: removed this line, this method is now at line 193. Because when we remove a forest client number from the information card, there is no need to remove the validation error under the input field. They are two difference places.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -404,7 +402,7 @@ function removeForestClientFromList(index: number) {
                                 :name="'verifyIdir'"
                                 :label="'Verify'"
                                 @click="
-                                    verifyIdentity(
+                                    verifyUserId(
```
MCatherine1994, line 405: rename this method from verifyIdentity to verifyUserId, feel that's more clear about which form field we're trying to verify

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -68,10 +66,6 @@ const selectApplication = async (e: DropdownChangeEvent) => {
     }
 };
 
-const selectApplicationOptions = computed(() => {
```
MCatherine1994, line 71: I feel there is no need to have this computed variable, can use the applicationsUserAdministers directly

frontend/src/components/managePermissions/ManagePermissionsTitle.vue
```diff
@@ -1,13 +1,4 @@
-<script setup lang="ts">
-
-const props = defineProps({
```
MCatherine1994, line 3: this is not in use, removed

J0taFerreira (approved):

NickSaglioni (approved):
Seems to be good. Thank you.

frontend/src/components/header/Header.vue
```diff
@@ -43,18 +40,18 @@ const props = defineProps({
             </Button>
             <span class="header-title">
                 {{ props.title }}
-                <strong class="subtitle">{{ props.subtitle }} {{ environmentLabel }}</strong>
+                <strong class="subtitle"
+                    >{{ props.subtitle }} {{ environmentLabel }}</strong
+                >
```
ianliuwk1019, line 45: Wondering if it is the VS code breaking the tag like this with strange looking "**</strong**" and 
"**>**" at the next line?

frontend/src/components/managePermissions/ManagePermissions.vue
ianliuwk1019: Sounds good. Could you also remove the unused **"computed" import** on line 2?
I am a bit curious that in original code at line 64 with the same variable `applicationsUserAdministers` used in if condition, **does it work** that way in javascript for Vue's `ref`? Or should it be `if (applicationsUserAdministers.value) {`? Maybe it does not even need this "if" condition... Would you mind to verify this condition?

frontend/src/components/grantaccess/GrantAccess.vue
ianliuwk1019: The best practice for naming **boolean** variables (or methods) is using "**is**" "**has**" "**should**" "**can**" with **positive** wording. I would strongly favor to use them, although sometimes it is hard to find a good naming. 
For example, `if ( `**if**ForestClientNumberNotAdded()` )` would be really strange when other developers see that.

ianliuwk1019 (commented):
Thanks Catherine. Just some minor comments.

frontend/src/components/header/Header.vue
MCatherine1994: I think this is done by auto formatting. I tried again, it auto formats this way.

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: sure I'll revert it back, sometimes it's hard to give a name :   )

frontend/src/components/managePermissions/ManagePermissions.vue
MCatherine1994: yeah, this is good catch, and it should check if it has a value and if the length is > 0, which means it should have at least one role, I'll change it
```
if (
        applicationsUserAdministers.value &&
        applicationsUserAdministers.value.length > 0
    )
```

ianliuwk1019 (approved):
Thank you Catherine!!

PR #1065, opened 2023-11-30
feat: #1018 add auth test for bceid and bcsc login

MCatherine1994:
- add auth test for bceid and bcsc login
- add display name to the audit log for bcsc login, cause bcsc login doesn't return idp username


server/auth_function/readme.md
```diff
@@ -35,6 +35,7 @@ sudo dockerd
 cd server/auth_function
 pytest
 ```
+Note: We need to use master db user(postgres) to run the tests.
```
MCatherine1994, line 38: currently the auth function needs access to fam_user and fam_role table. Later one once we read admin role, we need access to admin tables. So only master db user can have access to all those tables.

server/auth_function/test/conftest.py
```diff
@@ -96,46 +94,8 @@ def cognito_context():
     yield context
 
 
-@pytest.fixture(scope="session")
-def test_user_properties(cognito_event):
```
MCatherine1994, line 100: Remove these fixture methods, as they rely on other fixture methods that is parametrized. Refactor them into the util file as regular function, so we can pass parameter in.

server/auth_function/test/lamda_function_test.py
```diff
@@ -5,20 +5,34 @@
...
+@pytest.mark.parametrize(
+    "cognito_event",
+    [
+        "login_event.json",
+        "login_event_bceid.json",
+        "login_event_bcsc.json",
+    ],
+    indirect=["cognito_event"],
+)
+def test_create_user_if_not_found(db_pg_transaction, cognito_event, cognito_context):
```
MCatherine1994, line 33: parametrize the cognito_event fixture method, so by given different json file, it will return different data to run the test

server/auth_function/test/conftest.py
```diff
@@ -80,8 +78,8 @@ def db_pg_transaction(db_pg_connection, monkeypatch):
 
 # Load a sample cognito test event if necessary
 @pytest.fixture(scope="session")
-def cognito_event():
-    event_file_path = os.path.join(os.path.dirname(__file__), "login_event.json")
+def cognito_event(request):
```
ianliuwk1019, line 81: Just curious, I assume "request" isn't the http request but the pytest internal object?

server/auth_function/test/util.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+    # For Insert statement, pass parameters as .execute()'s second arguments so they get proper sanitization.
+    cursor.execute(
+        raw_query,
+        (
+            test_user_properties.get("idp_username"),
+            test_user_properties.get("idp_type_code"),
+            TEST_ROLE_NAME,
+        ),
+    )
```
ianliuwk1019, line 135: Just curious, were these functions previous from `conftest.py` not easy to use (despite you refactored a little on code)? 
All of these functions migrated into utils.py seems very much like dependencies before running tests, which is one thing pytest's "**@pytest.fixture**" is good at and convenient to use. And fixture can contain other fixtures, the function in test does not need to call all of them(the setup/dependencies).

server/auth_function/test/lamda_function_test.py
```diff
@@ -5,20 +5,34 @@
...
-):
 
+@pytest.mark.parametrize(
+    "cognito_event",
+    [
+        "login_event.json",
+        "login_event_bceid.json",
+        "login_event_bcsc.json",
+    ],
+    indirect=["cognito_event"],
```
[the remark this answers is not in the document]
MCatherine1994: So I have to set indirect in order to pass parameter to fixture methods. Oh but I can just set it to be true, cause we only one parameter for cognito_event here. 

Using the indirect=True parameter when parametrizing a test allows to parametrize a test with a fixture receiving the values before passing them to a test.

server/auth_function/test/conftest.py
MCatherine1994: this is how we can read the parameter passed to fixture method

server/auth_function/test/util.py
MCatherine1994: oh yeah, initially I need to pass some parameter into the cognito_event fixture, but you're right, let me try again, maybe it's now working.

MCatherine1994: Hi Ian, you're right, even we pass parameters the fixture contains another fixture still works as expected, I switched back, thanks so much for noticing this!!

server/auth_function/lambda_function.py
```diff
@@ -90,6 +89,10 @@ def lambda_handler(event: event_type.Event, context: Any) -> event_type.Event:
             audit_event_log["requestingUser"]["idpUserName"] = event["request"][
                 "userAttributes"
             ]["custom:idp_username"]
+        else:
+            audit_event_log["requestingUser"]["idpDisplayName"] = event["request"][
+                "userAttributes"
+            ]["custom:idp_display_name"]
```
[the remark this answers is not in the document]
ianliuwk1019: I see, I didn't notice this idp_user_id is in log already. Thanks for point that out.
Sure, then that seems to make sense with display_name; I assume it might be empty but does not matter.

server/auth_function/test/lamda_function_test.py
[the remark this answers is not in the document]
ianliuwk1019: "True/False" is the default with one parameter. I see how it works now.

ianliuwk1019 (approved):

server/auth_function/lambda_function.py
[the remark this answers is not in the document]
MCatherine1994: I just saw your comment haha, our test bcsc account has display name that's why I feel it might be good to show it, it's a readable name.

Issue #1052, opened 2023-11-24
Frontend and Pipeline: New Configuration for AXIOS for The Second API (admin_management api).


ianliuwk1019:
**Describe the task**
Current FAM has 1 (one) backend api from frontend to communicate to.
With the 2nd FAM backend api (admin_management api) created, frontend will also need to connect to this new api.

Frontend uses package `axios` for issuing http requests and handling some middleware configuration (such as exception handling and header token injection). This is `axios` is used as **an** **instance** for our all Vue components communications to the backend. 
The 2nd backend api will also needs to be hooked up and configured with **the instance** so all requests still work (communications to api-1 and api-2) and other behaviors in between are still the same (exception handling and headers). 

**Acceptance Criteria**

Pipeline:
- ~[ ]  Add one more smoke testing for the new admin lambda at the end of the Terraform Apply (`reusable_terraform_server.yml`).~ This is moved to #1056

Frontend local:
- [x] Add `fam_admin_management_api_base_url` (Terraform apply already creates this), into frontend `env.json`.

Frontend configuration:
- [x] Unclear as of the time of writing the task, but should research during implementation that if we can conditionally call `api-1` and `api-2` from the same `axios` instance by setting its instance's `baseURL` (api-1 and api-2 has different url path) (see frontend `HttpCommon.ts`).
- [x] If first criteria isn't easy or not possible, then create a second instance to connect to api-2, however, need to dig into what's the good way for components to easily submit different requests without IF condition to handle issuing requests for `axios` to different api. (Probably adding another layer before `ApiServiceFactory.ts` for components to call). Refactoring is probably required.
- [x] When above 1 or 2 criteria is done, include previous generated 2nd api-client (api-2) to frontend (`ApiServiceFactory.ts`).
- [x] Test the whole frontend system to work, including error handling behavior (can ignore 2-api since no frontend UI yet).

**Additional context**
-

ianliuwk1019:
We need two different `baseURL` for (now) two different API calls.
After experimenting overriding `baseURL` from the initial `axios` instance (after instance is `create()` and baseURL is set), it somehow does not work as expected from current npm `axios` documentation. However, due to the current version (`0.26.0`) of axios FAM is using is much lower than latest version (`1.6.2`), things might be good on latest version.
[...]

PR #1067, opened 2023-12-05
feat: #1052 incorporate axios with admin api

ianliuwk1019:
- Add `fam_admin_management_api_base_url` value into `env.json` for local admin api server url.
- Remove `baseURL`, from `defaultAxiosConfig` so that baseURL can be conditionally added later
	for different api(s).
- Add new `AdminMgmtApiBaseUrl` to `EnvironmentSettings.ts` and refactor.
- Define a SERVICE_KEY for ApiServiceFactory for dependency injection.
- Group services into `appAccessControlApiService` and `adminManagementApiService` and provide
	getters for these two groups of services.
- Add generic createInstance to instantiate Axios api(s) in ApiServiceFactory.
- At main.ts, use Vue `provide` to provide services.
- Inject `apiService` to `ManagePermission.vue` and `GrantAccess.vue`.


MCatherine1994:
this looks good me from the testing perspective, I tried a way to mock the inject and the tests pass

Screen Shot 2023-12-06 at 12 18 30 PM
Screen Shot 2023-12-06 at 12 18 44 PM

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -19,16 +19,17 @@ import {
...
-const userRoleAssignmentApi = apiServiceFactory.getUserRoleAssignmentApi();
+const apiService = requireInjection(ApiServiceFactory.SERVICE_KEY); // Inject service
 const userRoleAssignments = shallowRef<FamApplicationUserRoleAssignmentGet[]>();
 
 onMounted(async () => {
     // Reload list each time we navigate to this page to avoid forcing user to refresh if their access changes.
     applicationsUserAdministers.value = (
-        await applicationsApi.getApplications()
+        await apiService
+            .getAppAccessControlApiService()
```
MCatherine1994, line 31: I wonder if we still want to create a variable for these apis first? instead of every time call the whole chain "apiService.getAppAccessControlApiService()....."? maybe still can have:

const applicationsApi = apiService.getAppAccessControlApiService().applicationsApi
const userRoleAssignmentApi = apiService.getAppAccessControlApiService().userRoleAssignmentApi
etc.?

ianliuwk1019: Yes, I have noticed that as well, but didn't put much thought of it yet.
For now for long chain might be ok within the component itself (like you did) and can find an idea later.
However, I have been thinking since long time ago to get rid of this `onMounted` if necessary, so some magic might be good to happen on the routing level.
Or..., maybe create separate key(s) for the two `applicationsApi` and `forestClientsApi` as like what I did for the `apiService`. Let me try...

ianliuwk1019: Hi @MCatherine1994,
I have now provide two different services("appAccessControlApiService" and "adminManagementApiService") for injection available for convenience.
It will be up to the component to decide what to call from the service within the component.
I assume that would not break the test mocking.

MCatherine1994 (approved):

Issue #888, opened 2023-09-13
Refactor auth lambda to read FAM roles from new table


ConradBoydElliottGustafson:
Auth lambda currently reads from the fam_user_role_xref table. For FAM privileges, we want it to also read from fam_app_admin. Keep the authorization mechanism tied to the JWT so we don't have to refactor the whole API and frontend.

- [x] Add logic in the auth lambda. When the application being logged into is FAM itself, instead of reading from fam_user_role_xref, read from fam_app_admin. If the user has a record for the FAM application, add "FAM_ADMIN" to the JWT. If it is a different application, add "{app_name}_ADMIN". 
- [x] Create a Flyway script to migrate existing FAM role data from fam_user_role_xref, to fam_app_admin. 
- [ ] Remove the FAM roles from fam_user_role_xref, this will be done in another ticket #1066

MCatherine1994:
Do this last.

MCatherine1994:
The 1st and 2nd task is not blocked anymore. But the last cleanup task need to wait until frontend implementation is done, so frontend will read data from the new table, and then it will be safe to remove old roles.

MCatherine1994:
Tested in AWS dev, application admins have been added successfully. But the login doesn't work anymore, cause we forgot to add the read permission for auth lambda db user to the new fam_application_admin table, I opened a new pr for the fix https://github.com/bcgov/nr-forests-access-management/pull/1074

<img width="1269" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0JIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7466667dd982f70c8a1d1e411221b06a3c77d5a9/AWS_dev_1207.png" alt="AWS_dev_1207.png" />

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN0ZIQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--5fa7d665abb86973d3a3574adf96887434c295b6/AWS_dev%29app_admin.png" alt="AWS_dev)app_admin.png" />

PR #1068, opened 2023-12-05
feat: #888 transfer to admin table

MCatherine1994:
refs: #888
- add flyway script to migrate user with fam role to fam application admin table
- update auth function to add fam application admin to role list when user login through fam itself
- update the auth function test to verify the admin role can be added as expected
- update the backend and admin management router guard, replace the hardcoded admin role from "{app_name}_ACCESS_ADMIN" to "{app_name}_ADMIN", cause once we merge this pull request, auth function will append admin roles as "{app_name}_ADMIN"


server/auth_function/lambda_function.py
```diff
@@ -207,6 +207,45 @@ def handle_event(db_connection, event) -> event_type.Event:
     for record in cursor:
         role_list.append(record[0])
 
+    # check if login through FAM
```
MCatherine1994, line 210: I wonder if we want to check if the user login through FAM itself, which way is better?

- use the cognito_client_id and query in the database to see if the application_name is "FAM"
- hardcode the cognito_client_id for FAM (dev, test, prod), and just check if the given cognito_client_id is FAM's cognito_client_id

We check if user login through FAM is because we only need to query the fam_application_admin table when user login through FAM itself.

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
```diff
@@ -0,0 +1,18 @@
+CREATE UNIQUE INDEX fam_usr_app_admin_uk ON app_fam.fam_application_admin(user_id, application_id);
```
MCatherine1994, line 1: think we need to add the unique index constrains, so user admin record won't have any duplicate

server/admin_management/tests/constants.py
```diff
@@ -21,9 +21,10 @@
 TEST_APPLICATION_NAME_FAM = "FAM"
 
 # -------------------- test application admin data ------------------ #
+TEST_APPLICATION_ADMIN_ID = 3
```
MCatherine1994, line 24: change the testing application id to 3 (I think it's FOM_TEST), cause most of our default users are already admin of application 1 (FAM) and 2 (FOM_DEV)

ianliuwk1019: I think you used it for `application_id`, is it ok to name it `TEST_APPLICATION_ADMIN_APPLICATION_ID `?

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
```diff
@@ -0,0 +1,18 @@
...
+-- first select all fam roles
+-- and then select role_id and application_id this role is admin of, based on if the role name contains the application_name
+-- -- for example, role name: FAM_ACCESS_ADMIN contains application name: FAM
+-- --              role name: FOM_DEV_ACCESS_ADMIN contains application name: FOM_DEV
+-- select the user_id and application_id the user is admin of, insert into fam_application_admin
+INSERT INTO app_fam.fam_application_admin (user_id, application_id, create_user, create_date)
+SELECT user_role_xref.user_id, application.application_id, CURRENT_USER, CURRENT_DATE
+FROM app_fam.fam_role role
+JOIN app_fam.fam_application application
+    ON role.role_name LIKE '%' || application.application_name || '%'
```
ianliuwk1019, line 13: Why do we need this join condition? Can it be just `role.application_id = application.application_id` ( on which `role.application_id = 1`)?

server/auth_function/lambda_function.py
```diff
@@ -207,6 +207,45 @@ def handle_event(db_connection, event) -> event_type.Event:
...
+        cognito_client_id=sql.Literal(cognito_client_id),
+    )
+    cursor.execute(sql_query_application)
+    # if login through FAM, check fam app admin and add to role list
+    for record in cursor:
+        if record[0] == "FAM":
+            query_fam_app_admin = """
+            SELECT application.application_name
+            FROM app_fam.fam_application_admin app_admin
+                INNER JOIN app_fam.fam_application application ON
```
ianliuwk1019, line 229: I guess, `JOIN` already imply `INNER JOIN` by default.

server/auth_function/lambda_function.py
ianliuwk1019: I feel fine from your implementation. 
However, later for delegated admin, I guess `auth_function` at this condition (when login to FAM) needs to be adjusted further and query new table again for adding extra roles into JWT and we might need to refactor it into better block of code/function?

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
MCatherine1994: thanks Ian, I didn't notice the naming convention, will update

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
ianliuwk1019: Ah, I see, it requires some brain twisting!

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
```diff
@@ -0,0 +1,15 @@
...
+JOIN app_fam.fam_user_role_xref user_role_xref
+    ON role.role_id = user_role_xref.role_id
+WHERE role.application_id=1;
```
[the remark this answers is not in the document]
ianliuwk1019: (since you mentioned) do you feel it is safer to use sub-query to replace this `application_id=1` for all environment `application_id = (sub-query ... where application_name = 'FAM')`? 

Also, after migration, maybe we like to have second script to remove this user from FAM_ADMIN (`('GATITEBI')`) record and add one more person ('CMENG'), probably need to add into fam_user if not exists?
Not sure what @basilv think.

ianliuwk1019:
Hope I am wrong, or maybe I am just confused myself, but would still like to ask this question:

Although if this PR is good, should this PR be pending for merge until **either** "**frontend**" or somehow "**backend api**" starts directing "role assignment" for "**xyz_ACCESS_ADMIN**" role to the new "**fam_application_admin**" table first?
If this PR flyway script is promoted to production, but our frontend/api isn't ready to make a switch yet, then will later the records for "**xyz_ACCESS_ADMIN**" in `fam_user_role_xref` be **out of sync** than current state (if, for example, Basil is being added to "**SPAR_PROD**_ACCESS_ADMIN" in current `fam_user_role_xref` table then he will not exists in `fam_application_admin` table.
@MCatherine1994 @basilv

basilv:
Agree somewhat with Ian that we can't release this functionality until the rest of it is done. We could merge this and deploy to dev, understanding that we will no longer be able to add admins in dev until the rest of the associated changes are done. (we can add admins, but will be in the wrong table and this won't be used to determine authentication).

server/flyway/sql/V34__migrate_fam_role_assignment_to_admin_table.sql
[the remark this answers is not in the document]
basilv: Ian, there's strictly speaking no need for a migration, we should have already removed Gbola's access and added Catherine's access as fam admin in the various environments, which will then be mirrored in the new table.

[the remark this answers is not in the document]
ianliuwk1019: Ah, yes @basilv , I was too focused into sql and forgot the easy way. Thanks for mentioning that :)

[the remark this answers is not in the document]
ianliuwk1019: I was only wondering maybe use sub-query would be safer. But no need, since you already make sure the id.

ianliuwk1019 (approved):

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -39,7 +39,7 @@
 
 
 def authorize_by_fam_admin(claims: dict = Depends(validate_token)):
-    required_role = "FAM_ACCESS_ADMIN"
+    required_role = "FAM_ADMIN"
```
MCatherine1994, line 42: I need to update hardcoded required role in router guard for both backend, and admin_management, because once we merge this pull request, all FAM roles will be added as "{app_name}_ADMIN" instead of "{app_name}_ACCESS_ADMIN"

MCatherine1994:
Verified in AWS tools space

we have these persons have FAM roles:
AWS_tools_1207

migrate them into the new fam_application_admin table:
AWS_tools_app_admin

ianliuwk1019 (approved):

PR #1074, opened 2023-12-09
fix: #888 add missing access for auth lambda db user

MCatherine1994:
add missing access for auth lambda db user #888

image


MCatherine1994:
adding read access for the auth lambda db user to fam_application_admin table should solve our login issue. I don't have any access in fam-tools, maybe @ianliuwk1019 you don't mind try to login to fam-tools and see if it's success?

ianliuwk1019:
> adding read access for the auth lambda db user to fam_application_admin table should solve our login issue. I don't have any access in fam-tools, maybe @ianliuwk1019 if you don't mind try to login to fam-tools and see if it's working? Thanks so much!!

Thanks @MCatherine1994 . The fix looks good.
Tools space frontend does not work properly yet, error occurred, but I can still login and verify the token at least in Tools space.
I also connect my local to tools space Cognito client and it looks correct to me.
image

ianliuwk1019 (approved):

PR #1075, opened 2023-12-09
fix: #888 fix auth function when read application admins

MCatherine1994:


server/auth_function/lambda_function.py
```diff
@@ -228,14 +228,11 @@ def handle_event(db_connection, event) -> event_type.Event:
             FROM app_fam.fam_application_admin app_admin
                 INNER JOIN app_fam.fam_application application ON
                     app_admin.application_id = application.application_id
-                JOIN app_fam.fam_application_client client ON
-                    app_admin.application_id = client.application_id
```
MCatherine1994, line 232: Just realized we already checked if fam_application_client.cognito_client_id = {cognito_client_id} above, and here it's just the case when user login to FAM itself. So when we read from the fam_application_admin table, there is no need to match the application id with fam_application_client table anymore, otherwise the only application_name got returned is "FAM", but it should return all the rest application name where the user is admin of. 

Found the problem when I login to FAM-dev, the token should have roles
[
"FAM_ACCESS_ADMIN", 
"FOM_DEV_ACCESS_ADMIN",  # this first two are got from role table, because we haven't clean it up
"FAM_ADMIN", 
"FOM_DEV_ADMIN"  # the last two are from fam_application_admin table
]

However I didn't got "FOM_DEV_ADMIN", and then I realized the problem is from here.

ianliuwk1019: I see. Thanks for finding that.
This line probably also no need anymore (the parameter): `cognito_client_id=sql.Literal(cognito_client_id),` on line 240.

ianliuwk1019 (approved):

server/auth_function/lambda_function.py
ianliuwk1019: This is token from tools space for me:
image
Looks correct.

Issue #1076, opened 2023-12-10
Improve frontend component with fetching before navigation.


ianliuwk1019:
**Describe the task**
Improve frontend component with fetching before navigation happens.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

PR #1077, opened 2023-12-10
fix: #1076 fetch before navigation

ianliuwk1019:
- [x] Use singleton for ApiServiceFactory and export services instead, then remove usage of Provide/Inject.
- [x] Organize some data fetching code into fetchData.ts.
- [x] Add router beforeEnter() to handle dataFetching before navigating to components.
- [x] Minor naming refactoring.


frontend/src/store/UserAccessRoleState.ts
```diff
@@ -0,0 +1,14 @@
+import type { FamApplicationUserRoleAssignmentGet } from 'fam-app-acsctl-api';
+import { shallowRef } from 'vue';
+
+export const userRoleAssignments = shallowRef<
+    FamApplicationUserRoleAssignmentGet[]
+>([]);
+
+// --- Setter
+
+export const setUserRoleAssignments = (
```
MCatherine1994, line 10: Hi Ian, I feel normally we set state variables when we want to share the value between different components. But in our case, only the ManagePermission component needs this value, is that better if we pass it as prop as well from router? and still use local variable to handle?

ianliuwk1019: Thanks Catherine, initially I feel `userRoleAssignments` should be a global state, due to the fact that "**ManagePermission**" seems to be our top level component (dashboard) already. I may have overlook that time.
 
When I look again, the "**userRoleAssignments**" are shared between "**ManagePermission**" and child component "**UserDataTable**" component.

I wasn't aware this "**userRoleAssignments**" was passed to "**UserDataTable**" component as a **props**. Now I feel even "UserDataTable" does not need this props, it should reflect right away what's changed in the state to show the table. 

Question, do you think in the future "**UserDataTable**" is going to be **used by different tabs**, so the props is needed (I kind of think it is going to be stay only specific for showing and sharing "**userRoleAssignment**") ?

ianliuwk1019: Hi @MCatherine1994 I have switched this to be as props. Hope I didn't miss anything.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -14,78 +14,44 @@ import {
...
-        await getAppUserRoleAssignment();
-    }
+import {
+    deletAndRefreshUserRoleAssignments,
+    fetchUserRoleAssignments,
+} from '@/services/fetchData';
+
+const props = defineProps({
+    userRoleAssignments: {
+        // options fetched from route.
```
MCatherine1994, line 28: shall we update this comment?

frontend/src/services/fetchData.ts
```diff
@@ -0,0 +1,70 @@
+import { AppActlApiService } from '@/services/ApiServiceFactory';
+import { setApplicationsUserAdministers } from '@/store/ApplicationState';
+import type { FamApplicationUserRoleAssignmentGet } from 'fam-app-acsctl-api';
+
+// --- Fetching data (from backend)
```
MCatherine1994, line 5: Hi Ian, this file looks good to me, but I wonder if we could have another name? This is like a central place to make service calls including get and delete. I know in Redux these are called actions. Maybe something like appActlApiActions?  I'm really bad at naming ....

frontend/src/components/managePermissions/ManagePermissions.vue
ianliuwk1019: This is not necessary, I removed it.

frontend/src/services/fetchData.ts
ianliuwk1019: I don't use Redux before and don't know what actions does. I think it is for state management.
I don't prefer terms stick to a particular technology stacks unless the terms really speaks to itself, like "store/state".
If other people use different stack then it isn't clear what "actions" do to them and may confuse others.
The fetchData is created just for convenience on taking care of the calls (really, it is just to fetching the data from using the api(s) ) many components could face and does not limit to just one API (unless we like to). I basically just like to fetch data from database (and some minor/general processing I see fit after data is fetched), does not involve state if it does not need to.

MCatherine1994 (approved):

Issue #1069, opened 2023-12-06
Refactor frontend grant access form section into reusable component


MCatherine1994:
**Describe the task**
Refactor the frontend grant access form sections into reusable components so we can reusable them when grant application admin and delegated admin

- [x] create a new folder under grantaccess folder called "form"
- [x] separate the user domain radio selection into a new component
- [x] separate the username and verification into a new component, add validation 
- [x] separate the adding forest client number and verification into a new component, add validation
- [x] update the current GrantAccess.vue component to use the reusable form component, make sure it works as expected

**Additional context**
- Add any other context about the task here.
- Or here

MCatherine1994:
created a new ticket #1092 to keep improve the logic when handle submissions and push notifications.

PR #1080, opened 2023-12-11
refactor: #1069 refactor the grant access form

MCatherine1994:
- separate user domain selection, user name input and verification error, forest client number input and verification error
- simplify styles and remove unnecessary class names


MCatherine1994:
We can add the validation rule into each form field component, but since some of the validation rule depends on other field condition, I still keep the validation on the form level, didn't separate them into single component.

frontend/src/components/grantaccess/form/UserDomainSelect.vue
```diff
@@ -0,0 +1,46 @@
...
+import { UserType } from 'fam-app-acsctl-api';
+
+const domainOptions = { IDIR: UserType.I, BCEID: UserType.B };
+
+const props = defineProps({
+    domain: { type: String, default: UserType.I },
+});
+
+const emit = defineEmits(['change']);
+const computedDomain = computed({
```
MCatherine1994, line 13: we use a [writable computed variable](https://vuejs.org/guide/essentials/computed#writable-computed) here, so whenever we read its value, it's get from props.domain. And whenever we change it's value, it will call the set() method and call our emit method "change". 

Emit a method basically means calling a method from parent. So whenever emit has been called, it will call the method we got from the parent component GrantAccess.vue, which is "userDomainChange()", and it will update the domain in the formData. And once formData.value.domain got changed, it will pass the domain as props again to the child UserDomainSelect component and rerender it.  

```
const userDomainChange = (selectedDomain: string) => {
    formData.value.domain = selectedDomain;
    formData.value.userId = '';
};

<UserDomainSelect
  :domain="formData.value.domain"
  @change="userDomainChange"
/>
```

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -0,0 +1,194 @@
...
+const props = defineProps({
+    userId: { type: String, required: true },
+    roleId: String,
+    fieldId: { type: String, default: 'forestClientNumbers' }, // field id used in the form validation
+});
+
+const emit = defineEmits([
+    'setVerifiedForestClients',
+    'removeVerifiedForestClients',
+    'resetVerifiedForestClients',
```
MCatherine1994, line 29: these three emit functions are used to set the "verifiedForestClients" data in the formData in parent component

ianliuwk1019:
Hi @MCatherine1994 , I am half way but I think I probably don't even need to look into rest of it (of course I will, later on). I  already understand your idea when I look at first file, GrantAccess.vue page. Looks very good to me, basically that's also what I expect and would do. Thanks so much taking the efforts for making them reusable and maintainable.

Could you also look into this case, that I run into problem (**I haven't pinpoint where you might miss, or is it a current bug in current system**). When I switch from "**Submitter**" role (with client number enter) to "**Reviewer**" (no need for client id). I see the post request seems to be wrong, it adds the user as "Submitter". In this case, Basil should have been added as "**Reviewer**" without clientID.
image

Thank you Catherine.

Also, the way parent/child component works in your refactoring is a two-way data-binding. Later, we probably like to **upgrade Vue to 3.3+**, I think we can even refactor more by taking advantage on a new feature for what you are doing in this PR.

MCatherine1994:
Hi Ian, think I found where the problem is. That is because when user change to "FOM_REVIEWER" role, the forest client number section will not show anymore, that new component will not be rendered, so the watch effect on role id change will won't run in this case. So we have to remove the verifiedForestClients in the GrantAccess component when role dropdown onChange.  And let ForestClientInput.vue watch for abstract role id change.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -443,30 +257,28 @@ const composeAndPushNotificationMessages = (
...
                             optionValue="role_id"
                             :modelValue="field.value"
                             placeholder="Choose an option"
-                            class="w-100 custom-input"
+                            class="w-100 custom-height"
                             style="width: 100% !important"
                             v-bind="field.value"
                             @update:modelValue="handleChange"
-                            @change="removeForestClientSection()"
+                            @change="resetVerifiedForestClients"
```
MCatherine1994, line 275: we need this line to remove the verified forest client number from the formData when role changed

MCatherine1994:
Hi @NickSaglioni, please review this pull request when you have time, and check the example for forest client tests, feel free to let me know if you have any question, thanks so much!!

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -220,32 +128,28 @@ const handleSubmit = async () => {
...
-            : '';
-        const data = toRequestPayload(formData.value, item);
-
+        const forestClientNumber = formData.value.verifiedForestClients.pop();
+        const data = toRequestPayload(formData.value, forestClientNumber);
         await AppActlApiService.userRoleAssignmentApi
             .createUserRoleAssignment(data)
             .then(() => {
-                successForestClientIdList.push(itemForestClientNumber);
+                successForestClientIdList.push(forestClientNumber || '');
```
ianliuwk1019, line 136: Hi @MCatherine1994 , not sure if this is from previous design issue with subsequent function using the "...IdList", but I am wondering if `forestClientNumber` is "undefined" why does it needs to be pushed into the list with empty string?
Is it because the empty string is being used for output combined message?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -220,32 +128,28 @@ const handleSubmit = async () => {
         errorForestClientIdList: [] as string[],
     };
     do {
-        const item = forestClientData.value.pop();
-        const itemForestClientNumber = item?.forest_client_number
-            ? item.forest_client_number
-            : '';
-        const data = toRequestPayload(formData.value, item);
-
+        const forestClientNumber = formData.value.verifiedForestClients.pop();
```
ianliuwk1019, line 131: Hi @MCatherine1994 , I probably never try this before, but would this also works if we really need an empty string for the const, so we don't need to repeat it?
`formData.value.verifiedForestClients.pop() || ''`
Thanks.

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: I was a little bit confused about the logic here as well yesterday. But I think currently the setGrantAccessNotificationMsg method can detect the empty sting in the ForestClientIdList and return the correct message, it has been treated as a special case. I want a refactoring later to make this logic better. But for this ticket, I just follow what we have right now so it won't have too many logic changes.

frontend/src/components/grantaccess/GrantAccess.vue
MCatherine1994: well, at this step it needs to be undefined so when we pass to the toRequestPayload method, it will be ignored. The logic here is a little bit confused. Later on we need some refactoring as I mentioned above. For this ticket will just follow what we have now.

MCatherine1994: I was confused a lot yesterday, and got the error when use your suggestion. And then I realized there are some logic that we need to rework on.

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        ).mockImplementation(() => {
+            return Promise.resolve(testSuccessSearchResponse);
+        });
+
+        wrapper = mount(ForestClientInput, {
+            props: {
+                userId: 'testUser',
+            },
+        });
+    });
```
MCatherine1994, line 44: when we add more tests, maybe there is no need to mount the component in beforeEach, in the case we'll mock different response to the forest client search method.

ianliuwk1019 (approved):
Hi @MCatherine1994 ,
This is approved.
One question I noticed: when the **userDomain changes**, the userId is cleared but the **Organization** isn't being cleared, which is fine I guess. But when I start **typing userId** (due to userDomain change) to verify user, then the **Organization** suddenly gets cleared. Do you feel the Organization can also be cleared(reset) when userDomain is changed?

MCatherine1994:
> Hi @MCatherine1994 ,
> This is approved.
> One question I noticed: when the **userDomain changes**, the userId is cleared but the **Organization** isn't being cleared, which is fine I guess. But when I start **typing userId** (due to userDomain change) to verify user, then the **Organization** suddenly gets cleared. Do you feel the Organization can also be cleared(reset) when userDomain is changed?

Hi Ian, I tested in the current dev, and it's the same behaviour. But I think if change username will cleanup the organization, then change domain should do the same as well. I think we should have a separate review session with Olga to check these depends.

Issue #1087, opened 2023-12-12
Modify the data returned for the get application admin api


MCatherine1994:
**Describe the task**
The get_application_admin_by_application_id api should return the data that is readable for frontend to display in the table. Follow the similar structure as user_role_assignment. 
<img width="1370" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNFZUQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--c14058aad552e506ee07c8bfebfed80247ad6528/image.png" alt="image.png" />

**Acceptance Criteria**
- [x] Return user information as shown in the above screenshot
- [x] Return application information, application name and environment
- [x] Return only the information we need

**Additional context**
- The return should looks like below, with more detail information
 {
    "application_admin_id": 5,
    "application_id": 1,
    "user_id": 5,
    "user": {},
    "application": {}
  }

MCatherine1994:
<img width="1188" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNHB1QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--49936f25a7a44187c951dbc1650764044bcacb5e/Screen%20Shot%202023-12-18%20at%202.26.08%20PM.png" alt="Screen Shot 2023-12-18 at 2.26.08 PM.png" />

PR #1099, opened 2023-12-18
fix: #1087 update return of get app admin api to include user and application information 

MCatherine1994:
update return of get app admin api to include user and application information #1087
Screen Shot 2023-12-18 at 3 17 47 PM

fyi, from Pydantic v2, the "fields" option in model config has been removed, https://docs.pydantic.dev/dev/migration/#changes-to-config. For this ticket, will just add the schema with the attributes we need.


server/admin_management/api/app/services/user_service.py
```diff
@@ -29,13 +29,12 @@ def find_or_create(self, user_type_code: str, user_name: str, requester: str):
 
         fam_user = self.get_user_by_domain_and_name(user_type_code, user_name)
         if not fam_user:
-            request_user = schemas.FamUser(
-                **{
```
ianliuwk1019, line 33: Sorry, I forgot what the "**" do in Python, do you remember and why do we remove it here?

MCatherine1994: I think it's a way to unpack a dictionary so can pass its key-value pairs to the schema. Because I removed the FamUser schema, and create a new one called FamUserCreate with just what we need for creating a user, and no need to do this anymore.

ianliuwk1019: Although I don't feel it is due to your change on a new FamUserCreate class but it is interesting to see it simply works and probably it already worked like that before, the "**" wasn't needed maybe.... good to try later...

ianliuwk1019 (approved):

Issue #1097, opened 2023-12-15
[Frontend] - Routing to protected routes without being authenticated


NickSaglioni:
**Describe the bug**
The route guard is not blocking when to user tries to route to a protected routed without being authenticated. It opens up the page and in some cases like trying to open `/grant` for instance, it gives a lot of error messages.

**To Reproduce**
1. Logout from FAM
2. Route the browser to a protected route like `/grant` for instance
3. The page will be loaded with error messages in the screen

**Expected behavior**
The route guard should not allow unauthenticated users to route to protected routes and should redirect to the login page

**Screenshots**
<img width="1168" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNHRsQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--d2da47c59dcf9c1aa086f909bd7e03689fe0d778/Screenshot%202023-12-15%20at%2008.51.28.png" alt="Screenshot 2023-12-15 at 08.51.28.png" />

ianliuwk1019:
This should be a feature to implemented. We never had router guards before for routes.
And this ticket should be a direct ticket for implementation after previous spike ticket Jota had: #742.

NickSaglioni:
The router guard per se could be considered a feature but because we don't have it implemented right now, it leads to lots of error messages displayed to the user  when navigating to certain routes.

ianliuwk1019:
Hi @MCatherine1994 , I have created a [branch](fix/1097-add-route-security-ian) for this matters, and incorporate your change on toast service refactoring with my simplification changes. We can talk together after you come back next week. Hope you have a great holiday.

PR #1100, opened 2023-12-19
fix: #1097 frontend router error for user not login

MCatherine1994:
Fix the router error when user doesn't login, but try to reach '/grant' page. Previously we can still see the grant page, with a lot of toast error messages which is not correct.

Update the error handling in frontend, we show these errors in toast message:

- if user does not login, and tries to access any protected pages, always redirect back to landing page, and show the error message like "you're not login"
- if user login without any role
    - ‘/dashboard’, show empty table and the error message something is like “don’t have the necessary authorization for the requested action”
    - ‘/grant’, because no application can be selected to grant, redirect to the ‘/dashboard’, the error in dashboard page will show


frontend/src/services/fetchData.ts
```diff
@@ -59,7 +61,16 @@ export const deletAndRefreshUserRoleAssignments = async (
 export const fetchApplicationRoles = async (
     applicationId: number | undefined
 ) => {
-    if (!applicationId) return;
```
MCatherine1994, line 62: So the original error is cased by this situation, when no application is selected, it just return, and not make any api call, so we can't get the 401 unauthorized error from backend. By logic, when there is no application selected, we should redirect the user back to the dashboard page to let him make a selection. And when he gets redirect to the dashboard page, dashboard page will try to get applications and then return the 401 unauthorized error and redirect back to landing page.

MCatherine1994:
Another thing I noticed is that, when the user is redirect back to the landing page, there is no toast message anymore to show the error "You are not logged in. Please log in."  It looks like the new ApiServiceFactory and the http instance interceptors somehow is not working as we expected. I'll take a deeper look again, and wondering if Ian @ianliuwk1019 has any idea?

frontend/src/components/NotFound.vue
```diff
@@ -1,8 +1,11 @@
 <script setup lang="ts">
 import router from '@/router';
 import { useRoute } from 'vue-router';
+import { showToastErrorTopRight } from '@/store/ToastState';
 
-console.log(`Route "${useRoute().path}" not found, redirect to home.`);
+const notFoundMsg = `Route "${useRoute().path}" not found, redirect to home.`;
+console.log(notFoundMsg);
+showToastErrorTopRight(notFoundMsg);
```
MCatherine1994, line 8: when user reaches to a non exists page, pop up the toast error message to them, and redirect back to landing page.

frontend/src/main.ts
```diff
@@ -23,4 +25,7 @@ app.use(ConfirmationService);
 app.use(PrimeVue);
 
 app.use(router).mount('#app');
+
+app.config.errorHandler = (err) => onError(err);
```
MCatherine1994, line 29: um.... I'm not so sure if we need this or not, this is like a global error handling, but we have it previously in the Toast component, so I still keep I here.

frontend/src/router/index.ts
```diff
@@ -69,11 +75,19 @@ const routes = [
...
+                await fetchApplications();
+                // if user already selected an application in cache
+                const userRoleAssignments = await fetchUserRoleAssignments(
+                    selectedApplication.value?.application_id
+                );
+                Object.assign(to.meta, {
+                    userRoleAssignments: userRoleAssignments,
+                });
+            } catch (error) {
+                setToastErrorMsg(error);
```
MCatherine1994, line 88: we have to catch the error on api calls in the beforeEnter method, otherwise it will throw the router error, and it can't be handled. And we also can't popup the toast at this step because it's not in any of the router yet, we can only store the error messages and popup whenever enter the router page.

frontend/src/router/index.ts
```diff
@@ -129,13 +153,27 @@ const router = createRouter({
...
+        // will rediret the user back to the landing page, and pop up the error message
+        const { showToastErrorTopRight } = useToastService();
+        showToastErrorTopRight('You are not logged in. Please log in.');
+        router.push('/');
     }
 });
 
+router.afterEach((to, from) => {
+    const { showToastErrorTopRight } = useToastService();
+    if (toastError.value !== '') showToastErrorTopRight(toastError.value);
```
MCatherine1994, line 174: after we enter the router, if we set any error message in the beforeEnter method, it will be pop up.

frontend/src/router/index.ts
MCatherine1994: one design question, so when the user login but has no role, do we show him empty table on the manage permission page, or we direct him back to landing page?

frontend/src/router/index.ts
```diff
@@ -69,11 +75,19 @@ const routes = [
         component: ManagePermissionsView,
         beforeEnter: async (to: any) => {
             // Requires fetching applications the user administers.
-            await fetchApplications();
-            const userRoleAssignments = await fetchUserRoleAssignments(
-                selectedApplication.value?.application_id
-            );
-            Object.assign(to.meta, { userRoleAssignments: userRoleAssignments });
+            try {
```
ianliuwk1019, line 78: Hi @MCatherine1994 ,
We want to avoid using `try-cath` in router transition, it is not scalable for `try-catch` in every route (just same as you already saw in backend `router` for audit logging), and isn't the intention from Vue Router lifecycle hook.
In this case, if we try and catch, then when error happens, it will still land on `next` route. That's exactly what router guards trying to prevent: **only go to new route component when everything is fine or user is permitted, otherwise, not at all and stay in previous route**.

I see you also refactor ToastMessage to make message working when error is happening in routing. Thanks for the good refactoring! Now I kind of understand what's going why toast message wasn't showing previously. I will comment on the ToastState.ts since you have a good find actually!!

However, we can talk about this later. I will take a deeper look on this.

frontend/src/store/ToastState.ts
```diff
@@ -0,0 +1,72 @@
...
+    toastError.value = error;
+};
+export const clearToastError = () => {
+    toastError.value = '';
+};
+
+// the primevue toast can only be injected in setup or functional component,
+// in order to reuse the toast service,
+// we have to define it in the way as below
+// https://stackoverflow.com/questions/72425297/vue-warn-inject-can-only-be-used-inside-setup-or-functional-components
```
ianliuwk1019, line 17: Thanks @MCatherine1994  for finding this out and adding this comment, very useful!!
From the reference, it is probably the same issue I encountered just few weeks ago when using `Provide/Inject` that they only work for Vue components; and same issue happens for the `PrimeVue Toast`.
I see the reference mentions in the future, they want to make Toast able to handle all cases, that time we can simplify more on using toast service.

MCatherine1994:
close this one as Ian did a good refactoring to include the changes in this pr, #1104

Issue #970, opened 2023-10-13
Frontend Test: Add tests for adding forest client number


MCatherine1994:
**Describe the task**
Add integration tests for the adding forest client number functionality in grant access page.

**Acceptance Criteria**
- [x] test the invalid user input, forest client number is invalid (length<8), will not do this one in this ticket, validation will be tested in ticket #717
- [x] test the not found user input, forest client number is not found
- [x] test the inactive user input,  forest client number is inactive
- [x] test the duplicate user input, forest client number is already added
- [x] test the successful case for single forest client number
- [x] test the successful case for multiple forest client numbers
- [x] test adding multiple forest client number with mixed of successful and error case
- [x] test the adding button in on loading while we call the api
- [x] test the delete forest client number case
- [x] test the forest client card in error case
- [x] test the forest client card in successful case
- [x] test when property change, will cleanup the errors
- [x] test information card cleanup when property change 

**Additional context**
- might need to think about how to mock the forest client api

NickSaglioni:
I could not find a way to have the test 'adding button in on loading while we call the api' to work. It needs a change in the `mockImplementation` for the API call to have a non-blocking delay while handling the `LoadingState`. I have tried some approaches:

- setTimeout function doesn't work for testing, the delay cannot be seem;
- Using Vitest fake timers do apply the delay, but I could not verify the correct value of the `LoadingState` after the delay, so the test always fails (is fake timer running in a blocking way?).

@MCatherine1994

PR #1102, opened 2023-12-21
feat: #970 add tests for adding forest client number

NickSaglioni:
- Adding tests for the `ForestClientInput` component;
- Adding tests for the `ForestClientCard` component;
- Refactoring `LoadingState` to have getter and setter for the loading state;


frontend/src/store/LoadingState.ts
```diff
@@ -6,11 +6,17 @@ import { ref } from 'vue';
...
-export default {
-    isLoading,
-};
+// Refactor adding setter and getter so they can be be used in testing as well
+
+export const setLoadingState = (status: boolean) => {
+    loadingState.value = status;
+}
+
+export const isLoading = () => {
```
MCatherine1994, line 20: Hi Nick, I feel normally we call it getLoadingState? so set/get/reset/remove ?

frontend/src/tests/ForestClientCard.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { it, describe, beforeEach, expect, afterEach } from 'vitest';
+import { mount } from '@vue/test-utils';
+import type { VueWrapper } from '@vue/test-utils/dist/vueWrapper';
+import ForestClientCard from '@/components/grantaccess/ForestClientCard.vue';
+import type { FamForestClient } from 'fam-app-acsctl-api';
+
+const testActiveClient: FamForestClient[] = [
+    {
+        client_name: 'TEST',
```
MCatherine1994, line 9: Hi Nick, I think some of this testing data is reusable, like the forest client numbers, shall we create a constant file for testing data?

frontend/src/tests/ForestClientCard.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    afterEach(() => {
+        wrapper.unmount();
+    });
+
+    it('should show inactive for forest client card in error case', async () => {
+        // Error case
+        await wrapper.setProps({ forestClientData: testInactiveClient });
+        expect(wrapper.get('.p-tag-value').element.textContent).toBe('Inactive');
+    });
```
MCatherine1994, line 45: let's add a empty line after each test

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -1,46 +1,66 @@
...
+    });
+
+    afterEach(() => {
+        vi.clearAllMocks();
+        wrapper.unmount();
+        vi.runOnlyPendingTimers();
+        vi.useRealTimers();
     });
 
     it('should render forest client number input field', () => {
```
MCatherine1994, line 66: let's check the verify button is rendered, and initially is disabled

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
         // the forest client input should now be empty
         expect(inputField.value).toBe('');
     });
+
+    it('should raise error for forest client number not found', async () => {
+        const input = wrapper.find('#forestClientInput');
+        const inputField: HTMLInputElement = input.element as HTMLInputElement;
```
MCatherine1994, line 127: I wonder if we could set the input and inputField to the beforeEach, we use them almost in all the test cases

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        await input.setValue(TEST_INVALID_FOREST_CLIENT_NUMBER);
+        expect(inputField.value).toBe(TEST_INVALID_FOREST_CLIENT_NUMBER);
+
+        await wrapper.find("[aria-label='Add Client Numbers']").trigger('click');
+        await flushPromises();
+
+        expect(wrapper.find('#forestClientInputValidationError').element.textContent).toBeTruthy();
```
MCatherine1994, line 134: let's verify the text content as well, add the validation errors to the constant file, like the part "is inactive and cannot be added."

MCatherine1994: and no card is displayed in all the error cases

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        // Try to add the same client number again
+        await input.setValue(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
+        expect(inputField.value).toBe(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
+
+        await verifyButton.trigger('click');
+        await flushPromises();
+
+        expect(wrapper.find('#forestClientInputValidationError').element.textContent).toBeTruthy();
+        //The input field should still display the duplicate input
+        expect(inputField.value).toBe(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
```
MCatherine1994, line 171: and verify the forest client card has only one line

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        // the forest client information card shows and verify the information
+        expect(wrapper.html().includes('Verified Client ID information')).toBe(
+            true
+        );
+        expect(
+            wrapper
+                .find('.client-id-wrapper')
+                .text()
+                .includes(TEST_SUCCESS_FOREST_CLIENT_NUMBER)
+        ).toBe(true);
```
MCatherine1994, line 215: should we check all the success forest client numbers are added in the card?

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+    //     expect(isLoading()).toBe(false);
+
+    //     await wrapper.find("[aria-label='Add Client Numbers']").trigger('click');
+    //     await flushPromises();
+
+    //     expect(isLoading()).toBe(true);
+    //     expect(wrapper.find("[aria-label='Add Client Numbers']").element.textContent).contains('Loading')
+    // });
+
+    it('should test the delete forest client number case', async () => {
```
MCatherine1994, line 306: Hi Nick, when we test the delete functionality, there is no need to verify the setVerifiedForestClients is called, because we already tested it multiple times in the previous test cases, and there is no need to verify the card content. 

I feel it's better to just add one forest client number, and verify the card is displayed and has one row, and when you click delete, the card is removed.

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+    it('should cleanup error when property change', async () => {
+        const input = wrapper.find('#forestClientInput');
+        const inputField: HTMLInputElement = input.element as HTMLInputElement;
+        await input.setValue(TEST_INACTIVE_FOREST_CLIENT_NUMBER);
+        expect(inputField.value).toBe(TEST_INACTIVE_FOREST_CLIENT_NUMBER);
+
+        await wrapper.find("[aria-label='Add Client Numbers']").trigger('click');
+        await flushPromises();
+
+        const validationError = wrapper.get('#forestClientInputValidationError');
```
MCatherine1994, line 374: are we using validationError and validationErrorElement variables?

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        const validationErrorElement: HTMLInputElement = validationError.element as HTMLInputElement;
+        // Check if the error message has been raised
+        expect(wrapper.find('#forestClientInputValidationError').element.textContent).toBeTruthy();
+
+        await input.trigger('click');
```
MCatherine1994, line 379: what we are doing here?

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        await flushPromises();
+        expect(
+            wrapper
+                .find('.client-id-wrapper')
+                .text()
+                .includes(TEST_SUCCESS_FOREST_CLIENT_NUMBER)
+        ).toBe(false);
+    });
+
+    it('should cleanup error when property change', async () => {
```
MCatherine1994, line 365: I think this test case is about when input change, the error will be cleanup

MCatherine1994: And we need to add another test case, when userId or roleId changed, the forest client errors and card will be removed, and the resetVerifiedForestClients emit method should be called

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+
+        await input.trigger('click');
+        await flushPromises();
+
+        await input.setValue(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
+        expect(inputField.value).toBe(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
+        // The error message should be cleared
+        expect(wrapper.find('#forestClientInputValidationError').exists()).toBeFalsy();
+    });
+    it('should cleanup forest client card when property change', async () => {
```
MCatherine1994, line 387: This test doesn't look correct. 

So you should
1. add a success forest client number, and verify it, the forest client card displays
2. change the userId property, verify the forest client card is removed, and resetVerifiedForestClients emit method should be called 
3. repeat step 1, and change the roleId property, verify the forest client card is removed, and resetVerifiedForestClients emit method should be called

MCatherine1994: And we need to add another test case "should cleanup forest client number when property change". 

1. add an invalid or not found forest client number, and verify it, the error displays, the number remains in the input
2. change the userId property, verify the error message removed, the input is empty, and resetVerifiedForestClients emit method should be called
3. repeat step 1, and change the roleId property, verify the error message removed, the input is empty, and resetVerifiedForestClients emit method should be called

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+                ? setVerifiedForestClients[0][0]
+                : undefined
+        ).toEqual(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
+        // The second entry in the forest client card should be equal to the second active input
+        expect(
+            setVerifiedForestClients
+                ? setVerifiedForestClients[1][0]
+                : undefined
+        ).toEqual(TEST_SUCCESS_FOREST_CLIENT_NUMBER_2);
+        // the forest client information card shows and verify the information
```
MCatherine1994, line 263: let's verify both success numbers has been added to the card

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -102,4 +121,286 @@ describe('ForestClientInput', () => {
...
+        expect(inputField.value).toContain(TEST_INACTIVE_FOREST_CLIENT_NUMBER);
+        expect(inputField.value).toContain(TEST_INVALID_FOREST_CLIENT_NUMBER);
+    });
+
+    //TODO - The test case below needs to adjust the mockImplementation for the api call to have a non blocking delay
+    // so the loadingState and the button label could be tested. I have tried some approaches but none seem to work:
+    // - Use of setTimeout doesn't work well in testing (the timout is not applied)
+    // - Use of vi.useFakeTimers() does apply delay, but I am not sure if it is a blocking or non blocking task
+
+    // it('should test the adding button in on loading while we call the api', async () => {
```
MCatherine1994, line 292: for this test case, let's do this. Inside this test case, let's mock the api call to set the loading state to be true, and create a new wrapper. So when you put an input and click on button, the button label should become 'Loading...' and will be disabled. 
```
        vi.spyOn(
            AppActlApiService.forestClientsApi,
            'search'
        ).mockImplementation(async (item) => {
            setLoadingState(true)
            return Promise.resolve(mockAPIResponse(item));
        });
        tempWrapper = mount(ForestClientInput, {
            props: {
                userId: 'testUser',
            },
        });
        await flushPromises();
```

frontend/src/store/LoadingState.ts
NickSaglioni: I felt that `isLoading` would make more sense understanding what we are doing throughout the code, but you are right, it's not the usual pattern. I'll change it to getLoadingState.

frontend/src/tests/ForestClientInput.spec.ts
NickSaglioni: The reason why I didn't verify the text content for the validation errors in the first place is that it is not recommended in the documentation because the text might change in the future, which would lead to test fail.

If you feel that the validation error texting would be stable enough, we could change this test.

Maybe an alternative would be to check just if the text contains "inactive" or "invalid" words, which I think would not change. What do you think? @MCatherine1994 

Reference: https://vee-validate.logaretm.com/v4/guide/testing/
Screenshot 2023-12-22 at 07 51 30

frontend/src/tests/ForestClientInput.spec.ts
NickSaglioni: Ok, makes sense. I'll update it. Thanks.

ianliuwk1019:
Hi @NickSaglioni , could you also add a minor update on frontend readme about what types of tests we currently do, what tools we use, and how to run test locally? Thank you.

NickSaglioni:
> Hi @NickSaglioni , could you also add a minor update on frontend readme about what types of tests we currently do, what tools we use, and how to run test locally? Thank you.

Sure, I'll do that.

frontend/src/store/LoadingState.ts
ianliuwk1019: This is a state with Boolean value and the exported function already retrieves the `.value` from the state for the need. I think it should be ok for calling `isLoading`, so the usage is more readable on the javascript or html condition.

ianliuwk1019:
Hi Nick, I am not sure what this comment means, maybe its left over from the past.
Could you check and cleanup if not making sense.
Thank you.
image

ianliuwk1019:
Hi Nick, a bit confusing on this part, should be simple, could you check?
Is this logic unnecessary? The previous expect already make sure it has a emitted property, right?
image

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -86,20 +118,310 @@ describe('ForestClientInput', () => {
...
+            setVerifiedForestClients
+                ? setVerifiedForestClients[1][0]
+                : undefined
+        ).toEqual(TEST_SUCCESS_FOREST_CLIENT_NUMBER_2);
+        // The third entry in the forest client card should be equal to the third active input
+        expect(
+            setVerifiedForestClients
+                ? setVerifiedForestClients[2][0]
+                : undefined
+        ).toEqual(TEST_SUCCESS_FOREST_CLIENT_NUMBER_3);
```
ianliuwk1019, line 211: Probably you missed my previous comment. Since you already know `setVerifiedForestClients` is not `undefined` in this case, you can do: `expect(setVerifiedForestClients![2][0]).toEqual(TEST_SUCCESS_FOREST_CLIENT_NUMBER_3);`
And, some as other part of the `expect` as well, please scan your file and adjust them.
Thanks.

ianliuwk1019 (changes requested):

frontend/src/tests/ForestClientCard.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    it('should show multiple inactive forest client card in error case', async () => {
+        // Multiple success case
+        await wrapper.setProps({ forestClientData: testMultipleInactiveClient });
+        const clientList = wrapper.findAll('#forest-client-status') as DOMWrapper<Element>[];
+        for (let index in clientList) {
+            expect(clientList[index].element.textContent).toBe(STATUS_DESCRIPTION_INACTIVE);
+        }
+    });
```
ianliuwk1019, line 137: Hi @NickSaglioni , not sure if this case is already done somewhere else (or other test file for similar test case). I think now you can add a new test that "mix success, failure, inactive together in the forestClientData list" and verify if client number vs client status are what you expected.
Thanks Nick!

ianliuwk1019: I think I saw somewhere you had "mix" case.
Resolve this issue. Thanks.

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -75,31 +107,297 @@ describe('ForestClientInput', () => {
...
+
+        const clientIdCardList = wrapper.findAll('#forest-client-id');
+
+        expect(clientIdCardList).toHaveLength(2);
+
+        expect(clientIdCardList[0].text().includes(TEST_SUCCESS_FOREST_CLIENT_NUMBER)).toBe(true);
+        expect(clientIdCardList[1].text().includes(TEST_SUCCESS_FOREST_CLIENT_NUMBER_2)).toBe(true);
+    });
+
+    //TODO - The test case below needs to adjust the mockImplementation for the api call to have a non blocking delay
```
ianliuwk1019, line 275: Hi @NickSaglioni, if this case is very tricky and had already spent lots of time, should we do this in separate ticket to add and remove this commented out section? Thanks.
@MCatherine1994

ianliuwk1019 (changes requested):

frontend/src/tests/ForestClientCard.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+        // Success case
+        await wrapper.setProps({ forestClientData: testActiveClient });
+        expect(wrapper.get('#forest-client-status').element.textContent).toBe(STATUS_DESCRIPTION_ACTIVE);
+
+    });
+
+    it('should show multiple active forest client card in successful case', async () => {
+        // Multiple success case
+        await wrapper.setProps({ forestClientData: testMultipleActiveClient });
+        const clientList = wrapper.findAll('#forest-client-status') as DOMWrapper<Element>[];
```
MCatherine1994, line 124: could we verify the length as well? thanks!

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -75,31 +107,297 @@ describe('ForestClientInput', () => {
...
+
+        await input.setValue(TEST_SUCCESS_FOREST_CLIENT_NUMBER_2);
+        expect(inputField.value).toBe(TEST_SUCCESS_FOREST_CLIENT_NUMBER_2);
+
+        await wrapper.find("[aria-label='Add Client Numbers']").trigger('click');
+        await flushPromises();
+
+        expect(wrapper.findAll('#forest-client-id')).toHaveLength(1);
+
+        await wrapper.setProps({ userId: 'Test2' });
```
MCatherine1994, line 355: why we repeat the test when set props userId? do you want to set roleId here and verify the forest client card is removed?

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -75,31 +107,297 @@ describe('ForestClientInput', () => {
...
+
+        await wrapper.find("[aria-label='Add Client Numbers']").trigger('click');
+        await flushPromises();
+
+        expect(wrapper.find('#forestClientInputValidationError').element.textContent).toBeTruthy();
+
+        expect(inputField.value).toBe(TEST_INACTIVE_FOREST_CLIENT_NUMBER);
+
+        // Changing the userId prop should cleanUp Input
+        await wrapper.setProps({ userId: 'Test2' });
```
MCatherine1994, line 394: here again, I think you want to test userId change and roleId change, thanks!!

frontend/src/tests/ForestClientInput.spec.ts
```diff
@@ -1,46 +1,76 @@
...
             'search'
-        ).mockImplementation(() => {
-            return Promise.resolve(testSuccessSearchResponse);
+        ).mockImplementation(async (item) => {
+            setLoadingState(true);
+            return Promise.resolve(async () => {
+                setTimeout(() => { }, 5000);
+            }).then(() => {
+                return forestClientsApiSearchMock(item)
+            }).finally(() => setLoadingState(false));
```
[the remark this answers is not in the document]
MCatherine1994: there is no need to set loading here and wait the timeout, check my other comment for test the loading state. Let's just `return Promise.resolve(forestClientsApiSearchMock(item));`

frontend/src/tests/ForestClientInput.spec.ts
MCatherine1994: so here, we can mock the api a different way and mount another wrapper just for this test case,

```
 it('should test the adding button in on loading while we call the api', async () => {
        vi.spyOn(
            AppActlApiService.forestClientsApi,
            'search'
        ).mockImplementation(async (item) => {
            // here we suppose while we call backend api, the loading state will be true
            setLoadingState(true);
            return forestClientsApiSearchMock(item);
        });
        const wrapper = mount(ForestClientInput, {
            props: {
                userId: 'testUser',
            },
        });
        const input = wrapper.find('#forestClientInput');
        const inputField = input.element as HTMLInputElement;
        await input.setValue(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
        expect(inputField.value).toBe(TEST_SUCCESS_FOREST_CLIENT_NUMBER);
        await wrapper
            .find("[aria-label='Add Client Numbers']")
            .trigger('click');
        expect(isLoading()).toBe(true);
        expect(
            wrapper.find("[aria-label='Add Client Numbers']").element
                .textContent
        ).contains('Loading');

        // cleanup
        setLoadingState(false);
        vi.clearAllMocks();
        wrapper.unmount();
}
```

MCatherine1994:
Looks pretty good overall, just some small things left, almost there! thanks so much Nick and Ian!!!

MCatherine1994 (approved):

ianliuwk1019 (approved):

PR #1104, opened 2023-12-29
fix: #1097 add route security

ianliuwk1019:
Add route security and handle route error for toast.

- if user does not login, and tries to access any protected pages, always redirect back to landing page, and show the error message like "you're not login"
- if user login without any role
    - ‘/dashboard’, show empty table and the error message something is like “don’t have the necessary authorization for the requested action”
    - ‘/grant’, because no application can be selected to grant, redirect to the ‘/dashboard’, the error in dashboard page will show

### DIFF
--- a/frontend/src/router/routeHandlers.ts
+++ b/frontend/src/router/routeHandlers.ts
@@ -1,0 +1,46 @@
+import { FamRouteError, RouteErrorName } from "@/errors/FamCustomError";
+import { routeItems } from '@/router/routeItem';
+import AuthService from "@/services/AuthService";
+import { isApplicationSelected } from "@/store/ApplicationState";
+import { setRouteToastError as emitRouteToastError } from '@/store/ToastState';
+import type { RouteLocationNormalized } from "vue-router";
+
+// Route `beforeEach` handler.
+// Responsible for route guards and necessary operations before each route.
+export const beforeEachRouteHandler = async (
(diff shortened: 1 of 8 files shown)


MCatherine1994 (approved):

frontend/src/router/routeHandlers.ts
```diff
@@ -0,0 +1,105 @@
...
+    to: RouteLocationNormalized
+) => {
+    populateBreadcrumb([routeItems.dashboard, routeItems.grantUserPermission]);
+
+    const appRolesFetchResult = await asyncWrap(
+        fetchApplicationRoles(selectedApplication.value!.application_id)
+    );
+    if (appRolesFetchResult.error) {
+        emitRouteToastError(appRolesFetchResult.error);
+        return { path: routeItems.dashboard.path };
```
MCatherine1994, line 51: Hi @ianliuwk1019 , I only changed here, let it return to dashboard when there is an error before enter grant access page, thanks!!

ianliuwk1019: Thanks for the change Catherine! This route (in theory) should already come "from" dashboard.
However, I was thinking the same thing, that this can (for now) solve the case that user change browser address.

ianliuwk1019:
Merge this. Thanks Catherine!

Issue #1047, opened 2023-11-21
Create a new URL for access request


OlgaLiber2:
**Describe the task**
Create a new router for the delegate admin application home. Enable the login for IDIR and BCEID. 

**Acceptance Criteria**
- [x] determine the name of the url link
- [x] blank screen
- [x] url work/exists 
- [x] authentication redirect works. Check if the user already login through FAM, then no need to ask for login. If the user didn’t login, ask them to login. And if they don’t have any role, don’t allow them to access to other FAM urls. 

**Additional context**
- login option is idir and bceid
- URL structure something like fam.gov.bc.ca/public/end-user-request

NickSaglioni:
The current Vite configuration doesn't allow for URLs with `public` keyword. For now we will use something like `fam.gov.bc.ca**/end-user-request**`

<img width="872" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjZwQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fbea4a12e990f65f4c7987c15722052b3d8a1efc/Screenshot%202024-01-08%20at%2016.20.24.png" alt="Screenshot 2024-01-08 at 16.20.24.png" />

OlgaLiber2:
@NickSaglioni can you change "/end-user-request" to "/access-request"?

NickSaglioni:
@OlgaLiber2 sure!

NickSaglioni:
I had to do a small refactor in the `AuthCallbackHandler` to enable redirection to previous URL after Login. To achieve that I had to store the previous URL using the localStorage, as regular Vue ref gets cleared after Cognito's redirection. @MCatherine1994

NickSaglioni:
The priority for this task has been changed, so I am moving it back to the backlog so we can return to it when needed. @gormless87 @OlgaLiber2 @MCatherine1994

PR #1117, opened 2024-01-10
feat: #1047 adding new url to request fam access

NickSaglioni:
- Implement an URL link to request FAM access: `/access-request`
- Add a placeholder screen;
- Change authentication workflow to allow for redirection after login;
- Disable access to other FAM urls if the user don't have any roles;

### DIFF
--- a/frontend/src/components/AuthCallbackHandler.vue
+++ b/frontend/src/components/AuthCallbackHandler.vue
@@ -1,13 +1,19 @@
 <script setup lang="ts">
 import router from '@/router';
 import authService from '@/services/AuthService';
+import { getNavigationPath } from '@/store/NavigationState';
 
 /**
  * This component is to deal with Auth callback.
  */
 
 authService.methods.handlePostLogin().then((data) => {
-    router.push('/dashboard');
+    const navigationPath = getNavigationPath();
+    if (navigationPath) {
+        router.push(navigationPath);
+    } else {
+        router.push('/dashboard');
+    }
 });
 </script>
 
--- a/frontend/src/router/routeHandlers.ts
+++ b/frontend/src/router/routeHandlers.ts
@@ -12,6 +12,7 @@ import {
     selectedApplication,
 } from '@/store/ApplicationState';
 import { populateBreadcrumb } from '@/store/BreadcrumbState';
+import { setNavigationPath } from '@/store/NavigationState';
 import { setRouteToastError as emitRouteToastError } from '@/store/ToastState';
 import type { RouteLocationNormalized } from 'vue-router';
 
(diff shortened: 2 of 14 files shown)


MCatherine1994:
I feel there is no need to show the navigation bar on the left. 

Screen Shot 2024-01-11 at 2 20 02 PM

MCatherine1994:
I feel we need some design discussion here:

So let's say we provide a public link to users `https://fam.nrs.gov.bc.ca/access-request`, when the user access to the link, currently:
- if the user already login, display the page
- if not login, redirect to the FAM landing page and ask or login

But maybe it's better to popup the login option directly to them? Instead of redirect them to the landing page for login? Cause I feel most of the time, they don't need to go to our FAM landing page. This is just the design question, I'm fine with both way. @ianliuwk1019 @OlgaLiber2 
Screen Shot 2024-01-11 at 2 19 16 PM

`/dashboard`: 
- require auth: if user does not login, redirect to landing
- require role? if the logged in user has no role, should we redirect the user back to landing or redirect to the `/access-request` form? 

`/grant`:
- require auth: if user does not login, redirect to landing
- require application selected: if no application is selected, redirect back to dashboard
- require role? might do the same handling as the dashboard page

`/access-request`:
- require auth: if user does not login, pop up login options?

frontend/src/components/AuthCallbackHandler.vue
```diff
@@ -1,13 +1,19 @@
...
  */
 
 authService.methods.handlePostLogin().then((data) => {
-    router.push('/dashboard');
+    const navigationPath = getNavigationPath();
+    if (navigationPath) {
+        router.push(navigationPath);
+    } else {
+        router.push('/dashboard');
+    }
```
MCatherine1994, line 16: Just wondering if we `setNavigationPath(to.path)` in the `beforeEachRouteHandler`, then maybe we'll always have the navigationPath? no need to check if else here?

frontend/src/router/routeHandlers.ts
```diff
@@ -77,14 +78,25 @@ export const beforeEachRouteHandler = async (
...
     }
 
+    if (AuthService.getters.isLoggedIn() && to.path != routeItems.accessRequest.path && !AuthService.getters.userHasRoles()) {
+        const routeError = new FamRouteError(
+            RouteErrorName.NO_ROLES_ERROR,
+            "You do not have any FAM roles. Please request access.",
+            { to, from }
+        );
+        emitRouteToastError(routeError);
+        return { path: routeItems.accessRequest.path };
```
MCatherine1994, line 97: I like the idea to check if user has role or not, but might need some discussion about where we direct them to...

MCatherine1994:
thanks so much Nick for trying add more tests!!

MCatherine1994:
> > I feel there is no need to show the navigation bar on the left.
> > Screen Shot 2024-01-11 at 2 20 02 PM
> 
> I was thinking about to ways to hide the sidenav:
> 
> * Adding conditions to the existing layout (it needs both to hide the element and adjust the style for the class) which possibly would make it harder to understand later on;
> * Add a new Layout without the sidenav;
> 
> Which one do you think is the best approach? @MCatherine1994

Let's create a new one, thanks!!

Issue #887, opened 2023-09-13
Implement application admin management UI - manage permission UI


ConradBoydElliottGustafson:
**Describe the task**
We need a cool new screen to manage application admins! Implement the awesome design in #886 for manage permission UI, design in [Figma](https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=3637-204243&mode=design&t=a9nlm8GzUvGPUPz9-0) 
 [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN2tVQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--69aedd31e8a21c66688a9545a85ca89fe405948d/Screen%20Shot%202023-11-28%20at%2011.26.20%20AM.png)

**Acceptance Criteria**
 - [x] In `ManagePermissions.vue` component, when the user select "FAM" from the dropdown, load the application admins using the new admin management api
- [x] Create a new ApplicationAdminTable component, and whenever the user select "FAM" from the dropdown, display the new ApplicationAdminTable component with the application admin data
- [x] The new ApplicationAdminTable component will display the columns as in the Figma design (username, domain, application, environment, role, action). And we don't need to read role from database, just display "admin" for all the users. The application column we just show the application name, and we display the application environment in the environment column.

OlgaLiber2:
@ConradBoydElliottGustafson  I'm assuming a new screen to manage app admins + screen/s for granting access to a new app admin. Correct?

ConradBoydElliottGustafson:
Correct! Currently they are roles in FAM. The point of segregating is to remove those definitions from that table, so that someone can't hack the Lambda and give themselves FAM ADMIN. 

So yes, we need a simple new screen that lets the FAM admin grant app admin to users.

ConradBoydElliottGustafson:
I made a ticket #886 for a UI design task.

MCatherine1994:
We don't do the estimate now until the UI is ready so we'll have a better understanding.

OlgaLiber2:
@OlgaLiber2 update mockup with added columns

gormless87:
Closed 3 weeks ago in Project, but issue still showed it as open

Issue #1114, opened 2024-01-09
Add a sort icon in dashboard table column header


MCatherine1994:
**Describe the task**
Currently if we click on the table column header, it will sort the table based on selected column. We'd like to add a sort icon in the header so it's more obvious to the user. 
An example icon like this, could confirm with Olga or Alison how we want to align the icon:
<img width="1996" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNzJ3QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--cf5ac65ef65e60340f2bad9c1196381ba6315f50/Screen%20Shot%202024-01-09%20at%2012.29.17%20PM.png" alt="Screen Shot 2024-01-09 at 12.29.17 PM.png" />

**Acceptance Criteria**
- [ ] Added the sort icon to table header 
- [ ] Table sorting works 

**Additional context**
- Add any other context about the task here.
- Or here

MCatherine1994:
duplicated with #1130

Issue #1059, opened 2023-11-28
Implement application admin management UI - add application admin UI


MCatherine1994:
**Describe Task**
We need a cool new screen to manage application admins! Implement the awesome design in #886 for adding application admin UI, design in [Figma](https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=3637-204243&mode=design&t=a9nlm8GzUvGPUPz9-0) 
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNzRVQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4e8446c70c153dd1281484dbdb6fd922a4b50228/Screen%20Shot%202023-11-28%20at%2011.33.33%20AM.png)

**Acceptance Criteria**
- [x] Create a new component "GrantApplicationAdmin" under the grant access folder, follow the same style as "Add user permission" page. Follow the code implementation as example in GrantAccess.vue component. 
    - [x] On mounted, get all fam applications
    - [x] username field and application selection field are both required
    - [x] submit button is disabled if not pass username validation or missing any field
    - [x] Whenever click submit, call the new api to create an application admin, (need to add the user type "I" to the api call) and redirect to the manage permission page (under the same tab) with the notifications as what we have now
- [x] Follow the current design and create a view component for "GrantApplicationAdmin" and call it "GrantApplicationAdminView"
- [x] Add the router for the new "GrantApplicationAdmin" component, and setup the breadcrumb for it
- [x] Whenever user click "Add application admin" button on the manage permission page, the user will be direct to the new "Add application admin" page

NickSaglioni:
For the item 'On mounted, get all fam applications' should it be all the FAM applications or just the ones the user has access to? @MCatherine1994 @OlgaLiber2

MCatherine1994:
Hi @NickSaglioni, it should be a all FAM applications. Cause as a FAM admin, I should be able to assign application admin to all applications.

MCatherine1994:
Hi Nick @NickSaglioni, I think we need to fix a few things:

- When I click on "cancel" the add application admin form, in the screenshot you will see there is an "undefined" text, which is suppose to the the application name. I feel maybe we don't have to show the application name, maybe just say "Enter the user information to add a new application admin"

[...]

PR #1118, opened 2024-01-10
feat: #887 Implement application admin management UI - manage permission UI

J0taFerreira:
- show admin data table when FAM application is selected


frontend/src/components/common/DataTableHeader.vue
```diff
@@ -0,0 +1,149 @@
+<script setup lang="ts">
```
MCatherine1994, line 1: Hi Jota, this file is duplicated, you have one here, another one under managePermissions folder. I think this DataTableHeader component and the ConfirmDialogText component can both placed under the managePermissions folder, cause they are only used for that page. 

And under managePermissions folder, we can create a new directory maybe called table, and move all the table related ones under it.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -31,26 +34,46 @@ const props = defineProps({
...
+            ? 'Application admins'
+            : 'Users';
+})
+
 const onApplicationSelected = async (e: DropdownChangeEvent) => {
     setSelectedApplication(e.value ? JSON.stringify(e.value) : null);
-    userRoleAssignments.value = await fetchUserRoleAssignments(
-        selectedApplication.value?.application_id
-    );
+    if(e.value.application_id === 2) {
```
MCatherine1994, line 63: we need to change the logic here, cause we have more than 2 applications. If application_id is 1, then fetch admins, otherwise fetch user role assignments. And I could recommend create constant variable on top of this file for the the FAM application name and application id, and use that constant variable in the file.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -68,6 +91,26 @@ async function deleteUserRoleAssignment(
         );
     }
 }
+
+const deleteAppAdmin = async ( admin: FamAppAdminGet) => {
```
MCatherine1994, line 95: you might want to try format the file again, just left some small pieces like the space here

frontend/src/static/sideNav.json
```diff
@@ -13,6 +13,12 @@
                 "icon": "virtual-column--key",
                 "link": "/grant",
                 "disabled": true
+            },
+            {
+                "name": "Add application admin",
+                "icon": "enterprise",
+                "link": "#",
```
MCatherine1994, line 20: and this link? I can get the grant access page

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -68,6 +91,26 @@ async function deleteUserRoleAssignment(
...
+    try {
+        applicationAdmins.value = await deleteAndRefreshApplicationAdmin(
+            admin.application_admin_id,
+            admin.application_id
```
MCatherine1994, line 99: and I think we shouldn't pass "admin.application_id". So the deleteAndRefreshApplicationAdmin method will call fetchApplicationAdmin with application_id to get the admins again after deletion, and this application id is what we selected from dropdown, which is the id of "selectedApplication". However, admin.application_id is the application this user is admin of, so it could be the id for fom, forest client or spar, and we can't use it. 

So we might don't need to pass the application id, inside the deleteAndRefreshApplicationAdmin, we can just call fetchApplicationAdmin(FAM_APPLICATION_ID), where FAM_APPLICATION_ID should be a new constant variable we created in `frontend/src/store/Constants.ts`

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,19 +143,22 @@ async function deleteUserRoleAssignment(
...
                 }"
             >
-                <TabPanel header="Users">
+                <TabPanel :header="tabHeader">
                     <template #header>
                         <Icon icon="user" :size="IconSize.small" />
                     </template>
                     <UserDataTable
-                        :isApplicationSelected="isApplicationSelected"
+                        v-if="selectedApplicationShortDisplayText == 'FOM_DEV'"
```
MCatherine1994, line 151: here as well, this should be the case when the selected application is not FAM, we have many other apps

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,19 +143,22 @@ async function deleteUserRoleAssignment(
...
                         :loading="isLoading()"
                         :userRoleAssignments="userRoleAssignments || []"
-                        :selectedApplicationDisplayText="
-                            selectedApplicationDisplayText
-                        "
                         @deleteUserRoleAssignment="deleteUserRoleAssignment"
                     />
+                    <ApplicationAdminTable
```
MCatherine1994, line 156: the format here doesn't look correct as well

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,19 +143,22 @@ async function deleteUserRoleAssignment(
...
+                        v-if="selectedApplicationShortDisplayText == 'FAM'"
```
MCatherine1994, line 157: so basically we put v-if here, then the user data table can just use v-else, like 
<UserDataTable v-else .....>

frontend/src/components/managePermissions/DataTableHeader.vue
```diff
@@ -0,0 +1,153 @@
...
+        >
+            <Icon icon="add" :size="IconSize.small" />
+        </Button>
+        <span class="p-input-icon-left">
+            <Icon icon="search" :size="IconSize.small" />
+            <InputText
+                id="dashboardSearch"
+                class="dash-search"
+                placeholder="Search by keyword"
+                v-model="filters['global'].value"
```
MCatherine1994, line 44: the search is not working anymore

frontend/src/components/common/DataTableHeader.vue
J0taFerreira: that's a good idea! I'll remove the duplicate and make a folder for the table ralated

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: I'll do that!

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: that one is on me, I will  fix it

frontend/src/components/managePermissions/ApplicationAdminTable.vue
```diff
@@ -0,0 +1,164 @@
...
+                :userName="confirmDeleteData.adminName"
+                :role="confirmDeleteData.role"
+            />
+        </template>
+    </ConfirmDialog>
+    <div class="data-table-container">
+        <div class="custom-data-table">
+            <DataTableHeader
+                btnLabel="Add application admin"
+                btnRoute="#"
```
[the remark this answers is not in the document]
J0taFerreira: I left with the "#" until the component on task #1059 is completed, but I can add the route

frontend/src/static/sideNav.json
J0taFerreira: sure, I'll do

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: ok, I'll make that refactor

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: got it, I'll make the change

frontend/src/components/managePermissions/DataTableHeader.vue
J0taFerreira: I'll take a look on what is going on

frontend/src/components/managePermissions/ApplicationAdminTable.vue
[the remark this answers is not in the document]
MCatherine1994: oh yes, you're right, sorry I forgot we haven't implement that feature, thanks!!

frontend/src/static/sideNav.json
MCatherine1994: sorry please ignore my comment here, you're right, it's not implemented yet

ianliuwk1019:
Hi Jota, the style seems not centered:
image

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
     type ISideNavItem,
 } from '@/components/common/SideNav.vue';
 import sideNavData from '@/static/sideNav.json';
-import { isApplicationSelected } from '@/store/ApplicationState';
+import { isApplicationSelected,selectedApplicationShortDisplayText } from '@/store/ApplicationState';
 
 const navigationData = ref<[ISideNavData]>(sideNavData as any);
 
+const initialLabel = () => {
+    if(selectedApplicationShortDisplayText.value === 'FAM') {
```
ianliuwk1019, line 14: Perhaps this application 'FAM' now should be a constant in frontend to be used on other places.

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
...
     disableSideNavOption('Add user permission', !isApplicationSelected.value);
 });
 
 watch(isApplicationSelected, (value) => {
-    disableSideNavOption('Add user permission', !value);
+    initialLabel();
+});
+
+watch(selectedApplicationShortDisplayText, () => {
+    initialLabel();
```
ianliuwk1019, line 32: Why are two `watch`s on watching almost the same thing? `isApplicationSelected` and `selectedApplicationShortDisplayText`. 
Doesn't `isApplicationSelected` suite the need (since `selectedApplicationShortDisplayText` also depends on that as well)?

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
     type ISideNavItem,
 } from '@/components/common/SideNav.vue';
 import sideNavData from '@/static/sideNav.json';
-import { isApplicationSelected } from '@/store/ApplicationState';
+import { isApplicationSelected,selectedApplicationShortDisplayText } from '@/store/ApplicationState';
 
 const navigationData = ref<[ISideNavData]>(sideNavData as any);
 
+const initialLabel = () => {
```
ianliuwk1019, line 13: Could you put a "comment" on why it is called "initial"?
Is it only for initial page load?
It is not clear to me since you are watching for something change and executing this function again.
If it is not only "initial", may be better adjust the function naming.

ianliuwk1019 (changes requested):

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +68,47 @@ export const fetchApplicationRoles = async (
...
 };
+
+/// Admin
+export const fetchApplicationAdmin = async (
+    applicationAdminId: number | undefined
+): Promise<FamAppAdminGet[]> => {
+    if (!applicationAdminId) return [];
+    const applicationAdmin = (
+        await AdminMgmtApiService.applicationAdminApi.getApplicationAdminByApplicationId (
+            applicationAdminId
```
ianliuwk1019, line 79: Hi @J0taFerreira , several things here:
The context you are creating the function for is important: the **name**, the input **arguments** and the **return**.

* I believe the function you created is to return more than 1. So, the function name sometimes might be better reflecting this fact (by appending 's' or other good plural name choices; otherwise developer would think (by looking at the name) the function just return 1.

* The argument you gave for your function seems indicating a wrong thing for internal function `getApplicationAdminByApplicationId`. Why would you pass `applicationAdminId` to internal function's `applicationId`?

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +68,47 @@ export const fetchApplicationRoles = async (
     ).data;
     return applicationRoles;
 };
+
+/// Admin
```
ianliuwk1019, line 72: Is this unfinished comment? Not sure...

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +68,47 @@ export const fetchApplicationRoles = async (
...
+};
+
+
+/**
+ * This will call api to delete applicationAdminId record from backend and fetch again
+ * to refresh the state.
+ * @param applicationAdminId id to delete fam_user_role_assignment record.
+ * @param applicationId id to fetch and refresh records with the applicationId.
+ */
+
```
ianliuwk1019, line 103: Please remove the gap here.

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +68,47 @@ export const fetchApplicationRoles = async (
...
+export const deleteAndRefreshApplicationAdmin = async (
+    applicationAdminId: number,
+    applicationId: number
+): Promise<FamAppAdminGet[]> => {
+    console.log('delete admin')
```
ianliuwk1019, line 108: Please remove console log.

ianliuwk1019 (changes requested):

frontend/src/static/sideNav.json
```diff
@@ -13,6 +13,12 @@
...
+                "disabled": true
             }
         ]
     }
```
ianliuwk1019, line 24: This `sideNav.json` is a bit strange, isn't it similar to the `routeItem.ts`?
If they are different purpose, at least why we code the:
*  "**link**" with value duplicated and same as "**path**" in `routeItems`?
* And "**name**" with value duplicated and same as "**label**" in `routeItems`?

ianliuwk1019 (changes requested):

frontend/src/store/ApplicationState.ts
```diff
@@ -49,3 +50,20 @@ export const selectedApplicationDisplayText = computed(() => {
         return '';
     }
 });
+
+export const filters = ref({
```
ianliuwk1019, line 54: What's this for, why is it in `ApplicationState.ts`? Or could you add comment for the purpose?

frontend/src/router/routeHandlers.ts
```diff
@@ -29,15 +30,27 @@ import type { RouteLocationNormalized } from 'vue-router';
...
 const beforeEnterDashboardRoute = async (to: RouteLocationNormalized) => {
+    let applicationAdmins;
+    let userRolesFetchResult;
     // Requires fetching applications the user administers.
     await asyncWrap(fetchApplications());
-    const userRolesFetchResult = await asyncWrap(
-        fetchUserRoleAssignments(selectedApplication.value?.application_id)
-    );
-    Object.assign(to.meta, { userRoleAssignments: userRolesFetchResult.data });
+    if(selectedApplication.value?.application_id === 1) {
```
ianliuwk1019, line 37: If I code this as **"1"** or **"2"**, you would not know what **"1"** or **"2"** are for if you just see it.

However, even if I knew, I would not use **"1"** or **"2"** in the logic, since they are meaningless and they are randomly assigned (a sequence id from database), they are not really reliable and controlled by us. In some cases or in different environments, your "**2**" maybe be "**5**" (although in our case, it is not the case), still it is dangerous to use. I would use some "code" (we don't have) or "application name" (at least not likely it will be changed, but you could put a comment for developer to be aware).

And I would only trust using application name with "1" = "FAM" (exceptional case), I would not trust "2" is "FOM_DEV". But, all with above saying, please use "FAM" not "1".

frontend/src/router/routeHandlers.ts
```diff
@@ -29,15 +30,27 @@ import type { RouteLocationNormalized } from 'vue-router';
...
+        applicationAdmins = await asyncWrap(
+            fetchApplicationAdmin(selectedApplication.value?.application_id)
+        )
+    } else if(selectedApplication.value?.application_id === 2) {
```
ianliuwk1019, line 41: The `else if` (for the case of `2`) is also quite wrong as well.
We will have many applications in FAM, what about other "3" "4" "5" applications etc?
Please rewrite for a better logic.

ianliuwk1019 (changes requested):

frontend/src/router/routeHandlers.ts
MCatherine1994, line 41: and here, just else, for all the rest apps, we fetch user role assignments

frontend/src/components/managePermissions/DataTableHeader.vue
MCatherine1994: I wonder if this is caused by the filters... you have a global filter in the store, but also have a separate admin filter in the ApplicationAdminTable. Maybe it's better the pass the filter as a property.

MCatherine1994:
I noticed you added the sort icon to the table header, that's really good, we actually have a ticket to do that #1114. But I just wonder if you want to do it in this pr, or leave it for that ticket? Cause you might also need to adjust the styles for that icon and I'm fine to leave the sort icon work to that new ticket. thanks!!

Screen Shot 2024-01-11 at 1 31 00 PM

frontend/src/components/managePermissions/DataTableHeader.vue
J0taFerreira: I already found the problem, I'm just testing to see if I can reuse without problem

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: Catherine suggested that, already put it into a costant

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: I'll test to see if it works.

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: I can do that

frontend/src/services/fetchData.ts
J0taFerreira: Hello! I followed the the famapplication-admin-api file. 

Screenshot from 2024-01-12 14-38-51

J0taFerreira: My mistake, I'll fix it.

frontend/src/static/sideNav.json
J0taFerreira: this is to auto populate the side nav, you just add a new here and the nav will display

frontend/src/store/ApplicationState.ts
J0taFerreira: this should have been removed.

frontend/src/router/routeHandlers.ts
J0taFerreira: thanks for the comment, makes sense.

frontend/src/components/managePermissions/ApplicationAdminTable.vue
```diff
@@ -0,0 +1,164 @@
...
+                    header="Domain"
+                    sortable
+                ></Column>
+                <Column
+                    field="application.application_name"
+                    header="Application"
+                    sortable
+                ></Column>
+                <Column
+                    field="application.app_environment"
```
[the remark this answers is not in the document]
J0taFerreira: the env comes as null from app_environment, it should I hard code it?

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: it does need both.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -1,27 +1,32 @@
 <script setup lang="ts">
-import { onUnmounted, shallowRef, type PropType } from 'vue';
+import { onUnmounted, shallowRef, computed, type PropType } from 'vue';
 import Dropdown, { type DropdownChangeEvent } from 'primevue/dropdown';
 import TabView from 'primevue/tabview';
 import TabPanel from 'primevue/tabpanel';
 import ManagePermissionsTitle from '@/components/managePermissions/ManagePermissionsTitle.vue';
-import UserDataTable from '@/components/managePermissions/UserDataTable.vue';
+import UserDataTable from '@/components/managePermissions/table/UserDataTable.vue';
+import ApplicationAdminTable from './table/ApplicationAdminTable.vue';
```
MCatherine1994, line 8: Hi Jota, could we make the import in a consistent way? thanks!!

frontend/src/components/managePermissions/table/DataTableHeader.vue
```diff
@@ -0,0 +1,169 @@
...
+        required: true
+    },
+    filter: {
+        type: String,
+        required: true
+    }
+});
+const emit = defineEmits(['change']);
+
+const proxyValue = computed({
```
MCatherine1994, line 24: normally we call computed variable in this case "computedFilter"

frontend/src/components/managePermissions/table/UserDataTable.vue
```diff
@@ -33,19 +33,10 @@ const props = defineProps({
...
-    isApplicationSelected: {
-        type: Boolean,
-        required: true,
-        default: false,
-    },
 });
 
-const filters = ref({
-    global: { value: null, matchMode: FilterMatchMode.CONTAINS },
+const userAssignmentsfilters = ref({
```
MCatherine1994, line 38: userAssignmentsFilters?

MCatherine1994: and you might need to add the user type in the filter, cause below you're using it
Screen Shot 2024-01-15 at 5 56 18 PM

frontend/src/components/managePermissions/table/DataTableHeader.vue
```diff
@@ -0,0 +1,169 @@
...
+    width: 16rem;
+    z-index: 2;
+    border-radius: 0rem;
+}
+
+.dash-search {
+    border-radius: 0 0 0 0;
+}
+
+.btn-icon {
```
MCatherine1994, line 100: this class I'm not seeing in use in this file, also the .btn-icon:disabled

frontend/src/components/managePermissions/table/DataTableHeader.vue
```diff
@@ -0,0 +1,169 @@
...
+
+@media (max-width: 390px) {
+    .data-table-container {
+        margin: 0;
+        padding: 0;
+    }
+}
+
+@media (min-width: 768px) {
+    .data-table-container {
```
MCatherine1994, line 141: the data-table-container classname is not in use in this file, these styles should be in the data table and admin table component file

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +69,41 @@ export const fetchApplicationRoles = async (
     ).data;
     return applicationRoles;
 };
+
+export const fetchApplicationAdmins = async (
+    applicationId: number | undefined
+): Promise<FamAppAdminGet[]> => {
+    if (!applicationId) return [];
+    const applicationAdmin = (
+        await AdminMgmtApiService.applicationAdminApi.getApplicationAdminByApplicationId (
```
MCatherine1994, line 78: So fetch application admins should return all application admins, not only FAM. When user select FAM from the dropdown, it will list the admins for all apps, cause FAM admin can manage the admin for all applications. @ianliuwk1019 I feel we need to add a new backend api call to return admins for all applications.

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
...
 } from '@/components/common/SideNav.vue';
 import sideNavData from '@/static/sideNav.json';
-import { isApplicationSelected } from '@/store/ApplicationState';
+import { isApplicationSelected,selectedApplicationShortDisplayText } from '@/store/ApplicationState';
 
 const navigationData = ref<[ISideNavData]>(sideNavData as any);
 
+// Show and hide the correct sideNav btn based on the application
+const sideNavAddPermissionOption = () => {
+    console.log('test')
```
MCatherine1994, line 15: let's remove the console.log

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
...
+    if(selectedApplicationShortDisplayText.value === 'FAM') {
```
MCatherine1994, line 16: let's switch to use the constant variable

frontend/src/services/fetchData.ts
ianliuwk1019: Correct @MCatherine1994 ; the new function `fetchApplicationAdmins` shouldn't switch context to call `getApplicationAdminByApplicationId` from API during implementation.
We can either change this `getApplicationAdminByApplicationId` (since for now don't seem to have a need for this function) on backend or create a new api at backend. I can have a quick adjustment at the backend (in separate PR) for incorporating the missing function tomorrow if needed urgently, but we probably need to discuss with Olga. This PR cannot be released alone without the change.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -31,26 +36,40 @@ const props = defineProps({
...
 });
 
 const onApplicationSelected = async (e: DropdownChangeEvent) => {
     setSelectedApplication(e.value ? JSON.stringify(e.value) : null);
-    userRoleAssignments.value = await fetchUserRoleAssignments(
-        selectedApplication.value?.application_id
-    );
+    if (e.value.application_id === FAM_APPLICATION_ID) {
+        applicationAdmins.value = await fetchApplicationAdmins(
+            FAM_APPLICATION_ID
```
ianliuwk1019, line 61: As @MCatherine1994 mentioned in the comment in `fetchData` on function `fetchApplicationAdmins()`, passing this `FAM_APPLICATION_ID` is wrong but will need to wait for new backend api to be ready.

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: Hello, Catherine! Just did, thanks!

frontend/src/components/managePermissions/table/UserDataTable.vue
J0taFerreira: added the missing filter, is the naming userRoleAssignmentsFilters ok?

frontend/src/components/managePermissions/table/DataTableHeader.vue
J0taFerreira: just removed it

frontend/src/components/managePermissions/table/DataTableHeader.vue
J0taFerreira: I moved it to the layout file to apply the style to both tables

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: did it, It is now using the constant

frontend/src/services/fetchData.ts
MCatherine1994: yeah we have to do the backend api first. Let's chat with Olga, this ticket needs that api.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,20 +138,34 @@ async function deleteUserRoleAssignment(
...
             >
-                <TabPanel header="Users">
+                <TabPanel
+                    header="Application admins"
+                    v-if="selectedApplicationShortDisplayText === FAM_APPLICATION_NAME"
+                >
+                    <template #header>
+                        <Icon icon="enterprise" :size="IconSize.small" />
+                    </template>
+                    <ApplicationAdminTable
```
NickSaglioni, line 148: May I suggest to have the TabPanel as a custom component that receives as props the header, the icon and the table? That way I think it could be reused for as many tabs as needed for this page with minimal code repetition.

NickSaglioni (changes requested):
Thanks Jota. Just a couple of comments.

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -0,0 +1,171 @@
+<script setup lang="ts">
+import { reactive, ref } from 'vue';
+import type { PropType } from 'vue';
```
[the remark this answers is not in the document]
J0taFerreira: thanks, just did!

frontend/src/components/managePermissions/table/UserDataTable.vue
```diff
@@ -33,24 +33,19 @@ const props = defineProps({
...
-    isApplicationSelected: {
-        type: Boolean,
-        required: true,
-        default: false,
-    },
 });
 
-const filters = ref({
-    global: { value: null, matchMode: FilterMatchMode.CONTAINS },
+const userRoleAssignmentsfilters = ref({
```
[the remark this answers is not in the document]
J0taFerreira: nice catch, thanks!

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,20 +138,34 @@ async function deleteUserRoleAssignment(
...
+                        :loading="isLoading()"
+                        :applicationAdmins="applicationAdmins || []"
+                        @deleteAppAdmin="deleteAppAdmin"
```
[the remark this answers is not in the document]
J0taFerreira: I tried your suggestion using slots but I think primevue has some limitations, I don't think it would work.

frontend/src/services/fetchData.ts
```diff
@@ -66,3 +69,39 @@ export const fetchApplicationRoles = async (
     ).data;
     return applicationRoles;
 };
+
+export const fetchApplicationAdmins = async (
+    applicationId: number | undefined
```
ianliuwk1019, line 74: Probably just `applicationId: number` should be fine, no need for `undefined` I think.

frontend/src/static/sideNav.json
ianliuwk1019: Maybe you didn't understand what I meant.
Probably better to unify them, especially, if the **path**s or **link**s can be out of sync.
Maybe you and @NickSaglioni can think about it later, I will ignore it for now.
I will just post screenshot of what I saw.
image

frontend/src/services/fetchData.ts
ianliuwk1019: Resolved this, code is changed for new endpoint used.

frontend/src/services/fetchData.ts
J0taFerreira: Thanks for the comment, Ian! Just removed any unnecessary code related.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,20 +136,34 @@ async function deleteUserRoleAssignment(
                     },
                 }"
             >
-                <TabPanel header="Users">
+                <TabPanel
+                    header="Application admins"
+                    v-if="selectedApplicationShortDisplayText === FAM_APPLICATION_NAME"
```
MCatherine1994, line 141: I wonder if we use selectedApplicationShortDisplayText, or we check the id `selectedApplication.value.application_id == FAM_APPLICATION_ID`, cause if we use the application to check in the above, we better keep the same way. Maybe we could change this variable to be selectedApplicationId

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -0,0 +1,169 @@
...
+                :userName="confirmDeleteData.adminName"
+                :role="confirmDeleteData.role"
+            />
+        </template>
+    </ConfirmDialog>
+    <div class="data-table-container">
+        <div class="custom-data-table">
+            <DataTableHeader
+                btnLabel="Add application admin"
+                :btnRoute="routeItems.grantUserPermission.path"
```
MCatherine1994, line 94: just a reminder to Nick, this has change to the new router for granting application admin

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
...
 onMounted(() => {
-    disableSideNavOption('Add user permission', !isApplicationSelected.value);
+    if(isApplicationSelected.value) {
+        sideNavAddPermissionOption()
+    }
 });
 
-watch(isApplicationSelected, (value) => {
-    disableSideNavOption('Add user permission', !value);
+watch(selectedApplicationShortDisplayText, () => {
```
MCatherine1994, line 31: and maybe in this file, we also check selectedApplicationId, so it's consistent.

frontend/src/layouts/ProtectedLayout.vue
```diff
@@ -6,16 +6,30 @@ import SideNav, {
...
+    } else {
+        disableSideNavOption('Add application admin', true);
+        disableSideNavOption('Add user permission', false);
+    }
+};
+
 onMounted(() => {
-    disableSideNavOption('Add user permission', !isApplicationSelected.value);
+    if(isApplicationSelected.value) {
+        sideNavAddPermissionOption()
```
MCatherine1994, line 27: I wonder maybe we can call it setSideNavAddPermissionOption to be more like a method name? thanks!!

frontend/src/router/routeHandlers.ts
```diff
@@ -29,15 +31,25 @@ import type { RouteLocationNormalized } from 'vue-router';
...
 const beforeEnterDashboardRoute = async (to: RouteLocationNormalized) => {
+    let applicationAdmins;
+    let userRolesFetchResult;
     // Requires fetching applications the user administers.
     await asyncWrap(fetchApplications());
-    const userRolesFetchResult = await asyncWrap(
-        fetchUserRoleAssignments(selectedApplication.value?.application_id)
-    );
-    Object.assign(to.meta, { userRoleAssignments: userRolesFetchResult.data });
+    if(selectedApplication.value?.application_id === FAM_APPLICATION_ID) {
```
MCatherine1994, line 38: yeah, like if we created a selectedApplicationId variable in the state, we can use it in all places I mentioned above, instead of the short name, thanks!!

frontend/src/components/managePermissions/ApplicationAdminTable.vue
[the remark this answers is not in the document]
MCatherine1994: think only our team members can be FAM admin, feel it's okay to leave it blank for now.

frontend/src/components/managePermissions/ManagePermissions.vue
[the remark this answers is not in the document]
MCatherine1994: might be okay for now

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: I'll do the change!

frontend/src/layouts/ProtectedLayout.vue
J0taFerreira: ok, I'll change the name

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -100,20 +136,34 @@ async function deleteUserRoleAssignment(
                     },
                 }"
             >
-                <TabPanel header="Users">
+                <TabPanel
+                    header="Application admins"
+                    v-if="selectedApplication?.application_id === FAM_APPLICATION_ID"
```
MCatherine1994, line 141: I wonder if here should be "selectedApplicationId"

J0taFerreira: It did not need the .value but I forgot to change that to the selectedApplicationId state

MCatherine1994 (approved):

ianliuwk1019 (approved):

NickSaglioni (approved):

Issue #921, opened 2023-09-21
Add first name and last name to user card when verify IDIR user


MCatherine1994:
**Describe the task**
Display first name and last name to user card when verify IDIR user when grant user access.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcFRHIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--ba4a6fef8d62923c8731ef0fb3df2913072753c1/Screen%20Shot%202023-09-21%20at%202.32.08%20PM.png)

**Acceptance Criteria**
- [x] backend get first name and last name information from our IDIM proxy project, and return to frontend
- [x] backend from_api_json method update json_dict["userId"] to json_dict.get("userId") to avoid key error in not found case 
- [x] frontend receive first name and last name information from back and display in the user card
- [x] verify if still need to show display name, if not, remove the display name and create a ticket in IDIM-PROXY project to remove display name in the return data and clean up return data in the not found case to only return "found: false"

This ticket is blocked by https://app.zenhub.com/workspaces/fsa-fingerprint-61f04f08304d3e001a4f2578/issues/gh/bcgov/nr-fam-idim-lookup-proxy/40

**Additional Context**
* This ticket may encounter issue faced at #365 (see the ticket comment on temporary solution for api-client gen).
* A special note is added in `client-code-gen` README to deal with current bug from swagger spec. Please also take a look of the note when generating api-client code.

MCatherine1994:
Cleanup ticket created in the nr-fam-idim-lookup-proxy project: https://app.zenhub.com/workspaces/fsa-fingerprint-61f04f08304d3e001a4f2578/issues/gh/bcgov/nr-fam-idim-lookup-proxy/64

and the cleanup need to wait until we deploy the changes in FAM prod

OlgaLiber2:
@MCatherine1994  looks good

PR #1119, opened 2024-01-10
fix: #921 fix user identity card style on mobile

MCatherine1994:
So currently this fix just to display the same content on mobile screen, but for long user name, it still doesn't look good. But in Figma we already have the design for card on mobile screen (will show less information), later on we can create a separate ticket to work on the mobile screen for both user card and forest client card.


frontend/src/assets/styles/card.scss
```diff
@@ -0,0 +1,50 @@
+@use 'sass:map';
+@use '@bcgov-nr/nr-theme/design-tokens/light-theme.scss' as lightTheme;
+
+.p-card .p-card-header {
```
MCatherine1994, line 4: the style for card header and content is in the common repo previously, but I feel this is just for our FAM, so moved it to our repo.

ianliuwk1019 (approved):

Issue #1085, opened 2023-12-11
Add the new fam_access_control_privilege table


MCatherine1994:
**Describe the task**
Add the new fam_access_control_privilege table in PostgreSQL

**Acceptance criteria**
- [x] The new table is created
- [x] Give our fam_admin_management_api db user SELECT, UPDATE, DELETE, INSERT privileges on the new fam_access_control_privilege table
- [x] Update the existing model to include this new table, for both admin_management and backend 
- [x] Add the table and attribute description, audit columns (create/update date time, create/update user as other tables)

**Additional context**
Data model design in confluence https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Admin+Segregated+Schema+Design

**Blocking**
[Add delegated admin GET endpoints](https://app.zenhub.com/workspaces/fsa-fingerprint-61f04f08304d3e001a4f2578/issues/gh/bcgov/nr-forests-access-management/1086)
[Add delegated admin DELETE endpoint](https://app.zenhub.com/workspaces/fsa-fingerprint-61f04f08304d3e001a4f2578/issues/gh/bcgov/nr-forests-access-management/1090)

ianliuwk1019:
There seems to have no need for this new table "fam_access_control_privilege" to be in regular access server/backend api, so it is not added into `model.py` for regular access api and does not create a grant for the regular access api db user.

PR #1120, opened 2024-01-10
feat: #1085 new access control privilege table

ianliuwk1019:
* New table for access control privilege: flyway script V36.
* Necessary grants for db users.
* Python models update.


server/admin_management/api/app/models/model.py
```diff
@@ -166,6 +168,75 @@ class FamApplicationAdmin(Base):
...
+        String(60),
+        comment="The user or proxy account that created or last updated the record.",
+    )
+    update_date = Column(
+        TIMESTAMP(timezone=True, precision=6),
+        onupdate=datetime.datetime.utcnow,
+        comment="The date and time the record was created or last updated.",
+    )
+    role = relationship("FamRole", back_populates="fam_access_control_privilege")
+    user = relationship("FamUser", back_populates="fam_access_control_privilege")
```
MCatherine1994, line 237: Hi Ian, it's just my for my learning, I wonder when we need lazy="join"?

MCatherine1994 (approved):

MCatherine1994:
look good, just need to merge main to this branch, thanks so much!!

server/admin_management/api/app/models/model.py
ianliuwk1019: Hi Catherine, that's good question though!
On high level, you might already know, it means "lazy loading" from ORM objects.
When there is a relationship (parent/child) between two ORM model classes, this "lazy" gives choices for when to load the relationship (child) object(s) when you have a select statement. And for loading relationship, there are 3 types: Eager, Lazy, or None (not loading at all). "Lazy" mode of loading means the ORM engine will only load the relationship when you actually access the child objects. 

However, in the beginning when learning to code in Python with SQLAlchemy, honestly I was aware but didn't pay much attention to it because SQLAlchemy made it quite confusing to me for some reason (than Java JPA). But since you asked, I looked again to its docs, still confusing because SQLAlchemy makes this "lazy" on many types. 
I think, your question on lazy="join" to me is essentially an "Eager" as mentioned one of 
the 3 types above (although it has lazy/join wording, to me it means lazy="false").

In this case on your commented area, the relationships to "user" and "role" have no "lazy" in the mapping, so I believe it is by default Lazy="true". In the past, very often this (lazy="true") is good for performance and probably what we want; Eager loading will be a little slower. However, I am not really sure in Python, where the ORM session boundary is...so that it can still manage the loading strategy...
I would say, leave it as default, then when we have CRUD code, pay attention to the query. If it does not return the child relationship we want, then probably we need to turn it to Eager mode (easiest way) or know where the lazy loading boundary is (normally it ends within the function being called, but I am not sure at Python, maybe more to learn in SQLAlchemy in detail).

MCatherine1994: Thanks Ian!! yes I was checking the documentation, and the lazy has a few options, but the explanation is not very clear. I like the idea to check if when we use it. Cause I saw the application admin model uses the lazy joint. thanks so much!!

ianliuwk1019: Thanks Catherine, I don't remember in application admin model why it has that... maybe was inherited from other model, we can tweak when we know more later.

Issue #972, opened 2023-10-16
Frontend: Upgrade Axios version.


ianliuwk1019:
FAM frontend is using package "axios" for http requests. Current axios version is "**0.26.1**". This seems to be too low as of now the latest axios is "**1.5.1**". We probably ignored axios upgrade from Renovate in the past.
Find a good time to deal with axios upgrade.

**Additional Context**
- This might be a version upgrade that will break frontend code and requires coding fixing and testing. 
- At the same time, the "**client-code-gen/gen**" also has "axios" version that need to be consistent with frontend.

**Acceptance Criteria**
- [x] Upgrade frontend "axios" version.
- [x] Upgrade "client-code-gen/gen" axios version and verify again code-gen works.
- [x] Frontend build/test pass.
- [x] Verify Frontend works correctly.

ianliuwk1019:
Couldn't find an `option` to specify `axios` version when using tools `openapitools/openapi-generator` with `typescript-axios`.
Since no such option available, I decided to individually bump version up manually for `axios` **after** api-client code is generatred. 

Increase our FAM `frontend`, `client-code-gen\gen\admin-management-api`, `client-code-gen\gen\app-access-control-api` for `axios`  from version `0.26.1` to `^1.6.5` and update readme for how-to.

The major version update does impact frontend and requires some code changes.

PR #1125, opened 2024-01-16
fix: #972 axios version upgrade

ianliuwk1019:
* After investigation, for now there is no option to specify **"axios" version** during api-client code generation using openapitools/openapi-generator with typescript-axios. So we upgrade `axios` version manually on "**frontend**", "**client-code-gen\gen\admin-management-api**" and "**client-code-gen\gen\app-access-control-api**".
* Bump "axios" version up from version 0.26.1 to **1.6.5**.
* Fixed affected code due to "axios" major version upgrade (However, the type for "**AxiosRequestConfig**" needs to be changed to **InternalAxiosRequestConfig**. This is a bit strange to me by the name but is needed based on library).
* Update client-code-gen readme.


frontend/package.json
```diff
@@ -18,7 +18,7 @@
         "@bcgov-nr/nr-theme": "^1.8.2",
         "@carbon/icons-vue": "^10.72.1",
         "aws-amplify": "^5.3.6",
-        "axios": "0.26.1",
+        "axios": "^1.6.5",
```
MCatherine1994, line 21: Hi Ian, just wondering since the axios version for frontend and code-gen need to be the same, maybe we don't want to add "^" in front of the version number?

ianliuwk1019: Yes, I was questioning myself about that previously as well but feel might be ok (it seems to be a version high enough for stable).
But, I will change that to be fixed at this version, to be more comfortable not having break in the future. Thanks for mentioning that.

ianliuwk1019: This is now at fixed version.

MCatherine1994 (approved):

Issue #1131, opened 2024-01-16
Additional FAM Admin Endpoints For Admin Management API.


ianliuwk1019:
For Application Administrators management, we need additional two Admin Management API endpoints:
- GET: All applications (that can be assigned to the person who will be the application administrator: FAM, FOM, etc...). 
- GET: All administrators for applications (existing administrators for different applications: FAM, FOM, etc...).

**Additional Context**
- Due to previous security segregation for API to be separated for regular access management and high privilege access management, the code were initially mostly duplicated at both APIs base, and the two endpoints were missing out from the new context (for FAM Admin) but discovered.
-  These two endpoints are needed so that the FAM Admin can see a list of administrators on different applications (not just FAM) and can assign users for being application administrators.

**Dependencies**
- Blocking: This issue blocks #887 and #1059 

**Acceptance Criteria**
- [x] GET: All applications created at Admin Management API.
- [x] GET: All administrators for applications at Admin Management API.
- [x] Basic authentication/authorization check in place: authorized only for FAM admin access.
- [x] Swagger update and API-Client code generated.
- [x] Configure the generated API-Client into frontend API Factory/Service.

ianliuwk1019:
Will do the code and open for review and merge first so it does not block other tickets.
Will add tests at different branch.

PR #1136, opened 2024-01-18
feat: #1131 additional get admin management endpoints

ianliuwk1019:
- Add new "/applications" router for admin_management api and new "get_applications" endpoint.
- Add new endpoint "get_application_admins" for admin_management api.
- Add service/repository code to get applications and application admins.
- Add router_utils and add service instances for dependency injection.
- Minor adjustment on router path "/application_admins" and fix tests.
- Update openapi json on admin_management api.
- Recreated api-client(s) for both.
- Configure frontend ApiServiceFactory.ts to include new endpoint service.


ianliuwk1019:
I think we can ignore the files show up on `client-code-gen/gen/app-access-control-api` folder. I regenerated again and most of files have no difference but fore some reason show up in git.

server/admin_management/api/app/main.py
```diff
@@ -67,9 +66,14 @@ def custom_generate_unique_id(route: APIRouter):
 )
 app.include_router(
     router_application_admin.router,
-    prefix=apiPrefix + "/application_admin",
+    prefix=apiPrefix + "/application_admins",
```
ianliuwk1019, line 69: I slightly adjust this to follow the REST path convention.

frontend/src/services/ApiServiceFactory.ts
```diff
@@ -63,6 +64,10 @@ export default class ApiServiceFactory {
                 FAMApplicationAdminApi,
                 adminManagementBaseURL
             ),
+            applicationsApi: this.createApiInstance(
+                FAMApplicationsApi,
```
MCatherine1994, line 68: maybe doesn't matter, but if we want to use alias, shall we change to use alias in all places? also at line 22

server/admin_management/api/app/schemas.py
```diff
@@ -44,6 +44,12 @@ class FamApplicationBase(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 
 
+class FamApplicationGet(FamApplicationBase):
+    application_id: int
```
MCatherine1994, line 48: I feel we need to return the application name as well, cause if we display in the dropdown, we have to display the name?

server/admin_management/api/app/main.py
MCatherine1994: okay, no problem

MCatherine1994:
> I think we can ignore the files show up on `client-code-gen/gen/app-access-control-api` folder. I regenerated again and most of files have no difference but fore some reason show up in git.

yeah not sure why we have many empty file with difference here... but if you feel it's okay then maybe it doesn't matter, thanks!!

frontend/src/services/ApiServiceFactory.ts
ianliuwk1019: All these API names were actually generated from client-code-gen.
Frontend only imports them.
The reason I put the comment and use alias on line 7 was both `fam-admin-mgmt-api` and `fam-app-acsctl-api` all contains generated api(s) called "**FAMApplicationsApi**" and so both conflicts each other when import into the same file.
I prefer to stay the name they are generated from, however, in this case it happens to be the same so I had to choose using alias to avoid the conflict (in this case, I intended to let `fam-admin-mgmt-api` have the original name). This change will not affect external code that uses both api(s).
I can ask you tomorrow, maybe you have better idea or maybe I misunderstand your thought.

server/admin_management/api/app/schemas.py
ianliuwk1019: Yes, indeed that's needed :) and it does have the application name because it is inherit from the `FamApplicationBase` which produces:
image

MCatherine1994: oh right, thanks Ian I missed that

frontend/src/services/ApiServiceFactory.ts
MCatherine1994: Hi Ian, now I got it, all good, thanks!!

MCatherine1994 (approved):

Issue #1135, opened 2024-01-18
Update FAM Readme For Difference On Instruction For Setting Up Workspace With Windows OS.


ianliuwk1019:
**Describe the task**
As one of our team member having trouble with Windows WSL (Linux Subsystem) installation and permissions, the alternative was to setup working Workspace on Windows directly.
After the workspace is setup and working on Windows, we need to update README so that others can follow and setup correctly in the future.

**Acceptance Criteria**
- [x] README update for any difference or tips on setting up workspace on Windows than WSL (Linux).

PR #1137, opened 2024-01-19
chore: #1135 update FAM readme for Windows OS workspace setup

timisenco2015:
This PR is to update the FAM backend readme. The readme is updated to include steps to set up and start running both FAM backend and frontend.


server/backend/local-dev-window.env.bat
```diff
@@ -0,0 +1,16 @@
+SET POSTGRES_USER=***
+SET  POSTGRES_PASSWORD=****
```
MCatherine1994, line 2: I wonder maybe some variable value in this file no need to be hidden? Could just copy from the `server/backend/local-dev.env`. And we might want to remove the extra white space at this line, thanks!!

server/backend/readme.md
```diff
@@ -178,3 +178,35 @@ for test Output,
...
+```
+  - cd server/backend
+  - run docker compose up -d fam-flyway
+  - run this command in the same directory: python3 -m venv venv
+  - change your directory to:
+    - `cd server/backend/venv/Scripts`
+    - then enter(run) activate.bat to activate venv
+    - you will see: (venv) <Your base directory>\server\backend\venv\Scripts>
+  - cd back to server/backend
+  - Create a copy of `local-dev.env` and rename the file to `local-dev-window.env.bat.` Ask one of the developers for the environment properties in the local-dev.env, and copy and paste these properties into the local-dev-window.env.bat. Change and prefix each line with 'SET' e.g.
```
MCatherine1994, line 198: I wonder if we already provide a `local-dev-window.env.bat` file in our repo, do we still need to do this copy and rename step? thanks!!

server/backend/readme.md
```diff
@@ -178,3 +178,35 @@ for test Output,
...
+  - In the same directory, enter(run) local-dev-window.env.bat
+  - To run start the backend, run python3 server start.py
+```
+
+To Start or run the frontend
+ ```
+ - cd frontend
+ - npm run install-frontend
+ - npm run dev
+ ```
```
MCatherine1994, line 211: the frontend piece could add to `frontend/readme.md`, looks like we have it, but just need to update this part to be `npm run install-frontend`
image

server/backend/readme.md
timisenco2015: > I wonder if we already provide a `local-dev-window.env.bat` file in our repo, do we still need to do this copy and rename step? thanks!!

I made change but I did not save it before committing it to repo

frontend/readme.md
```diff
@@ -24,7 +24,7 @@ See [Vite Configuration Reference](https://vitejs.dev/config/).
 ## Project Setup
 
 ```sh
-npm install
+npm run install-frontend
```
ianliuwk1019, line 27: Hm... I didn't notice frontend has two readme before.
I think we could delete this one and just keep the first README.md

server/backend/readme.md
```diff
@@ -178,3 +178,27 @@ for test Output,
 
 - find the `output` tab
 - select `Python Test Log` in from the pulldown, top right of the output window
+
+
+## ----------------------------------------------------------------------------
+## -------------- Windows Configuration ----------------------------------------
+## ----------------------------------------------------------------------------
```
ianliuwk1019, line 185: Hi Ayobami, the header section looks a bit strange. Maybe it is ok to follow the one on top section?
image

server/backend/readme.md
```diff
@@ -178,3 +178,27 @@ for test Output,
...
+
+Before you follow the steps below, ensure you have Python installed or updated to the latest version. Install and start your Docker desktop for Windows.
+
+```
+  - cd server/backend
+  - run docker compose up -d fam-flyway
+  - run this command in the same directory: python3 -m venv venv
+  - change your directory to:
+    - `cd server/backend/venv/Scripts`
```
ianliuwk1019, line 194: There is probably a way to **NOT** going into child folder (/venv/Scripts) and still can execute the activate.bat.
Have you tried that?

server/backend/readme.md
```diff
@@ -178,3 +178,27 @@ for test Output,
...
+    - then enter(run) activate.bat to activate venv
+    - you will see: (venv) <Your base directory>\server\backend\venv\Scripts>
+  - cd back to server/backend
+  - Ask one of the developers for the environment properties in the local-dev.env, and update the properties in local-dev-window.env.bat with them
+    SET POSTGRES_USER=***
+    SET  POSTGRES_PASSWORD=***
+    SET POSTGRES_HOST=***
```
ianliuwk1019, line 201: Hi Ayobami, probably no need to mention these 3 "SET" environments here.

ianliuwk1019 (changes requested):

frontend/readme.md
ianliuwk1019: Please remove this readme.

ianliuwk1019: Checked with Ayobami on this readme duplication.
It is strange that I don't see that on his system. (Maybe something to do with same name files but UPPER case and lower case file names issue on Windows.)
I will resolve this one and delete it in other PR since Ayobami cannot see it.

ianliuwk1019 (approved):

server/backend/readme.md
timisenco2015: Fixed

server/backend/readme.md
timisenco2015: Fixed

Issue #1089, opened 2023-12-12
Add delegated admin POST endpoint


MCatherine1994:
**Describe the task**
FAM needs a mechanism to create and delete delegated administrator records. Create a POST endpoint that allows the creation of a new record in fam_access_control_privilege

**Acceptance Criteria**
- [x] the POST endpoint allows the creation of a new record in fam_access_control_privilege
- [x] make sure the endpoint require the application admin role 
- [x] make sure that the user can't manage their own records
- [x] add the audit similar to current application admin POST and DELETE actions
- [x] add unit tests

**Additional context**
- Add any other context about the task here.
- Or here

MCatherine1994:
When create a delegated admin, need the following steps:
- check user can't manage their own records
- check user is authorized (need to be app admin)
     - check if the role id in the request exists
     - check if the application the role belongs to exists
     - check if user is the admin of the application that get by role id
- check if user exists or create one
- check if role require child role
    - if require child role
        - check if forest client number is provided
        - check if child role exists or add a new child role
            - if child role not exists, find or create forest client number in the fam forest client table, and then create the child role
        - check if user privilege already exists, if not create the record
    - if not require child role
        - check if user privilege already exists, if not create the record

MCatherine1994:
will create a new ticket for adding forest client validator

PR #1138, opened 2024-01-20
feat: #1089 add post delegated admin endpoint

MCatherine1994:
Added POST method for creating access control privilege, this request allow creating access control privilege for multiple forest client numbers, refs: #1089.

- Request params has `username, user_type_code, role_id and forest client number list`
- Return result like below. For now we're not sure what information we'll display in the frontend, so return the detail user and role information for now, once we have the mock up, we can adjust to only return what we need.
```
[
  {
    "status_code": 409,
    "detail": {
      "access_control_privilege_id": 27,
      "user_id": 12,
      "role_id": 102,
      "user": {
        "user_type_code": "B",
        "user_name": "IANLIU",
        "user_type": {
          "code": "B",
          "description": "BCEID"
        }
      },
      "role": {
        "role_id": 102,
        "role_name": "FOM_SUBMITTER_00001011",
        "client_number": {
          "client_name": null,
          "forest_client_number": "00001011",
          "status": null
        },
        "parent_role": {
          "role_name": "FOM_SUBMITTER",
          "role_type_code": "A",
          "application_id": 2
        }
      }
    },
    "error_message": "User already has the requested access control privilege."
  },
  {
    "status_code": 200,
    "detail": {
      "access_control_privilege_id": 32,
      "user_id": 12,
      "role_id": 107,
      "user": {
        "user_type_code": "B",
        "user_name": "IANLIU",
        "user_type": {
          "code": "B",
          "description": "BCEID"
        }
      },
      "role": {
        "role_id": 107,
        "role_name": "FOM_SUBMITTER_00001016",
        "client_number": {
          "client_name": null,
          "forest_client_number": "00001016",
          "status": null
        },
        "parent_role": {
          "role_name": "FOM_SUBMITTER",
          "role_type_code": "A",
          "application_id": 2
        }
      }
    },
    "error_message": null
  }
]
```
- Audit log:
{"auditEventTypeCode": "CREATE_ACCESS_CONTROL_PRIVILIEGE", "auditEventResultCode": "SUCCESS", "application": {"applicationId": 2, "applicationName": "FOM_DEV", "applicationEnv": "DEV"}, "role": {"roleId": 3, "roleName": "FOM_SUBMITTER", "roleType": "A", "forestClientNumber": ["00001011", "00001016"]}, "targetUser": {"userGuid": null, "userType": "B", "idpUserName": "IANLIU", "cognitoUsername": null}, "requestingUser": {"userGuid": null, "userType": "I", "idpUserName": "CMENG", "cognitoUsername": "dev-idir_a9888e8ac6a0430e9f2200135df625bf@idir"}, "requestIP": "127.0.0.1"}
- Add tests for forest client, role and access control privilege related methods
- Add required permission for admin management api db user, it needs permission to create role and forest client record
- Create a follow up ticket to add forest client validator, #1139


server/admin_management/api/app/models/model.py
```diff
@@ -233,8 +233,8 @@ class FamAccessControlPrivilege(Base):
         onupdate=datetime.datetime.utcnow,
         comment="The date and time the record was created or last updated.",
     )
-    role = relationship("FamRole", back_populates="fam_access_control_privilege")
-    user = relationship("FamUser", back_populates="fam_access_control_privilege")
+    role = relationship("FamRole", back_populates="fam_access_control_privilege", lazy="joined")
```
MCatherine1994, line 236: turns out we have to add `lazy="joined"`, otherwise we can't get role information from the fam_access_control_privilege. By adding it makes the connection.

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -0,0 +1,150 @@
...
+                not hasattr(request, "forest_client_number")
+                or request.forest_client_number is None
+            ):
+                error_msg = (
+                    "Invalid access control privilege request, missing forest client number."
+                )
+                utils.raise_http_exception(HTTPStatus.BAD_REQUEST, error_msg)
+
+            for forest_number in request.forest_client_number:
+                # Check if child role exists or add a new child role
```
MCatherine1994, line 77: We suppose to validate the forest client number here. But there are too much code in this ticket, created a follow up ticket #1139 to add the validation.

server/admin_management/api/app/models/model.py
ianliuwk1019: Thanks for verifying that again Catherine!!
Then I guess that's probably the reason `lazy="joined"` was added in other model.
Not important for now, but would be curious how SqlAlchemy can allow to control lazy behaviour.

server/admin_management/api/app/schemas.py
```diff
@@ -96,3 +152,41 @@ class FamAppAdminGet(BaseModel):
...
+    user_type_code: famConstants.UserType
+    role_id: int
+    forest_client_number: Union[
+        List[Annotated[str, StringConstraints(min_length=1, max_length=8)]], None
+    ] = None
+
+    model_config = ConfigDict(from_attributes=True)
+
+
+class FamAccessControlPrivilegeCreate(BaseModel):
```
ianliuwk1019, line 169: Hi Catherine, this `FamAccessControlPrivilegeCreate` looks confusing with `FamAccessControlPrivilegeCreateRequest`. By the name from them I can't tell what their purpose are for which endpoints. Normally `schema.py` is where endpoints' request body classes and response body classes located.

It also looks a bit strange as "user_id", "create_user" are in its properties. I would not assume frontend at this point pass them in request body for granting access. Maybe this is for some other purpose?

server/admin_management/api/app/schemas.py
```diff
@@ -96,3 +152,41 @@ class FamAppAdminGet(BaseModel):
...
+
+    model_config = ConfigDict(from_attributes=True)
+
+
+class FamAccessControlPrivilegeCreateResponse(BaseModel):
+    status_code: int
+    detail: FamAccessControlPrivilegeGet
+    error_message: Optional[str] = None
+
+    model_config = ConfigDict(from_attributes=True)
```
ianliuwk1019, line 192: Hi Catherine, this `FamAccessControlPrivilegeCreateResponse` seems to be different from normal schema.py objects.
Questions are:
- What is this `status_code`? Is it our internal status code and different than Http Status Code so we need to defined a property for it?
- Why is it to have success response model return from endpoint but having `error_message` when this seems to indicate that it should be an error response? For example, in case of success creation of a resource in backend, it should have (either status code 201 or 200, like what indicated from FastAPI):
image
In case of error(s) happens in backend, depending on what is failing, normally it would be "403", "404", "409" or "500". FastAPI has indicated some error handling in documentation as well.
 
- Do we need this `model_config`? By looking at the properties, maybe it does not qualify for this `model_config` anymore?

We can discuss what' this class purpose is. Thanks Catherine!

ianliuwk1019 (changes requested):

server/admin_management/api/app/schemas.py
ianliuwk1019: Hmm... Just saw the service code... I am getting a bit understanding what you are trying to do now... I think I remembered you mentioned before you are doing (this time) a bulk creation .... Let me understand a bit more on your code. Thanks!

server/admin_management/api/app/schemas.py
```diff
@@ -96,3 +152,41 @@ class FamAppAdminGet(BaseModel):
...
 
     model_config = ConfigDict(from_attributes=True)
+
+
+# -------------------------- FAM Access Control Privilege (Delegated Admin) -------------------------- #
+class FamAccessControlPrivilegeCreateRequest(BaseModel):
+    user_name: Annotated[str, StringConstraints(min_length=3, max_length=20)]
+    user_type_code: famConstants.UserType
+    role_id: int
+    forest_client_number: Union[
```
ianliuwk1019, line 162: Hi Catherine, I think I remembered you mentioned before you are doing (this time) a bulk creation now.
In this case, could we have `forest_client_numbers` as this is now a List of `forest_client_number`?

server/admin_management/api/app/main.py
```diff
@@ -74,6 +79,11 @@ def custom_generate_unique_id(route: APIRouter):
     prefix=apiPrefix + "/applications",
     tags=["FAM Applications"],
 )
+app.include_router(
+    router_access_control_privilege.router,
+    prefix=apiPrefix + "/access_control_privilege",
```
ianliuwk1019, line 84: Hi Catherine, it's better to use "/access_control_privilege**s**" as REST path.

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -0,0 +1,90 @@
...
+        Depends(enforce_self_grant_guard),
+    ],
+)
+def create_access_control_privilege(
+    access_control_privilege_request: schemas.FamAccessControlPrivilegeCreateRequest,
+    request: Request,
+    db: Session = Depends(database.get_db),
+    token_claims: dict = Depends(jwt_validation.authorize),
+    requester: Requester = Depends(get_current_requester),
+):
```
ianliuwk1019, line 42: Hi Catherine, now I understand, the new endpoint you are adding is different than normal ones.
You are essentially adding a "**bulk**" creation of REST resource**s** (**ACCESS_CONTROL_PRIVLIEGE**) for the endpoint, not as previously one at a time.
Can we use method name as "create_access_control_privilege_**many/bulk**" and add some method comment for it (as initially I wasn't aware by this method name and got confused and realized later you are creating bulk endpoint). Thanks Catherine!
Or, @basilv , maybe you have better naming or comment on bulk design on POST endpoint?

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -0,0 +1,150 @@
...
+            )
+            create_return_list.append(handle_create_return)
+
+        LOGGER.debug(
+            f"Creating access control privilege executed successfully: {create_return_list}"
+        )
+
+        return create_return_list
+
+    def handle_access_control_privilege_create(
```
ianliuwk1019, line 98: Hi Catherine, maybe it is ok and no need to call it "**handle_**" (just `create_access_control_provilege` or just `grant_privilege`. Initially when I see "**handle_**" I was confused and thought this was a callback function (to handle scenario when something happens), but it is not.

Might be better to add method comment, since you have special handling in this case (bulk scenario) for constructing special kind of "response" than normally would, to explain why `FamAccessControlPrivilegeCreateResponse`: `status_code` and `error_message` are needed in case of error happens.

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -0,0 +1,150 @@
...
+    ) -> schemas.FamAccessControlPrivilegeCreateResponse:
+        access_control_privilege_return = None
+
+        # Check if user privilege already exists
+        fam_access_control_privilege = self.get_acp_by_user_id_and_role_id(
+            user_id, role_id
+        )
+
+        if fam_access_control_privilege:
+            error_msg = "User already has the requested access control privilege."
```
ianliuwk1019, line 109: This might not be clear if message is intended for frontend to display.

server/admin_management/api/app/services/role_service.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class RoleService:
+    def __init__(self, db: Session):
+        self.role_repo = RoleRepository(db)
+        self.forest_client_service = ForestClientService(db)
+        self.forest_client_repository = ForestClientRepository(db)
```
ianliuwk1019, line 20: This repository probably can be removed.

server/admin_management/api/app/utils/audit_util.py
```diff
@@ -11,6 +13,7 @@
 class AuditEventType(str, Enum):
     CREATE_APPLICATION_ADMIN_ACCESS = "Grant User Application Admin Access"
     REMOVE_APPLICATION_ADMIN_ACCESS = "Remove User Application Admin Access"
+    CREATE_ACCESS_CONTROL_PRIVILIEGE = "Create Access Control Privilege"
```
ianliuwk1019, line 16: Since the audience is for security auditor, maybe we can give description as "Grant Delegated Admin Privilege" or similar?

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -0,0 +1,90 @@
...
+from api.app.services.role_service import RoleService
+from api.app.utils.audit_util import AuditEventLog, AuditEventOutcome, AuditEventType
+
+LOGGER = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.post(
+    "",
```
ianliuwk1019, line 27: Maybe beneficial to add either one of these so on swagger it will be more descriptive (than by default using method name):
```
    name="Grant Delegated Admin Privileges",
    description="Grant Delegated Admin Privileges",
```

server/admin_management/api/app/schemas.py
```diff
@@ -78,7 +77,64 @@ class FamUserInfo(FamUserBase):
     model_config = ConfigDict(from_attributes=True, populate_by_name=True)
 
 
-# -------------------------------------- FAM Application Admin --------------------------------------- #
+# ----------------------------------- FAM Forest Client ------------------------------------ #
+class FamForestClientCreate(BaseModel):
+    # Note, the request may contain string(with leading '0')
+    forest_client_number: Annotated[str, StringConstraints(max_length=8)]
+    # client_name: str
```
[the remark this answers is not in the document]
MCatherine1994: as we discussed, will remove this comment. And for now we don't seem to use/need the client name (organization name) of the forest client number.

server/admin_management/api/app/services/access_control_privilege_service.py
MCatherine1994: Updated to be like ` "error_message": "User already has the requested access control privilege for FOM_SUBMITTER_00001013"`

server/admin_management/api/app/schemas.py
MCatherine1994: renamed to FamAccessControlPrivilegeCreateDto, and added comment to both

ianliuwk1019 (approved):

Issue #1083, opened 2023-12-11
Create component testing for UsernameInput component


MCatherine1994:
**Describe the task**
Since we separate the user name input form field into a new component in #1069, we can test it separately.

**Acceptance Criteria**
- [x] test the username input, set the input with a value and verify the input is expected
- [x] test the props, receive props correctly
- [x] test the emit function
    - [x] "change", when input change, the emit function is called 
    - [x] "setVerifyResult", when username verified, the emit function is called
- [x] test enter in the input field will enable the verify button
- [x] test the verify button shows "loading" while waiting for api result
- [x] test the username verification
    - [x] give a valid username, mock the api to return some user information (test the success case), check if user information card is displayed with expected information
    - [x] give a not found username,  mock the api to return not found response, check if user information card is displayed with expected "user does not exists" information
- [x] test when props.domain change, user information card will be removed, setVerifyResult will be called with different parameter

**Additional context**
- Add any other context about the task here.
- Or here

PR #1152, opened 2024-01-24
feat: #1083 component testing for usernameInput component 

J0taFerreira:
- added test to check if:
  - input is changing value as expected
  - props are being received as expected
  - change and emit setVerifyResult are being called
  - button will be enabled/desabled based on input change
  - button loading label is showing
  - user information card is displayed with correct information
  - user information card is displayed with user does not exists when fail to receive info
  - user information card is removed when domain changes


frontend/src/tests/common/fixPrimevueCssErr.ts
```diff
@@ -0,0 +1,12 @@
...
+//https://stackoverflow.com/questions/69906136/console-error-error-could-not-parse-css-stylesheet/69958999#69958999
+export const fixPrimevueCssError = () => {
+const originalConsoleError = console.error;
+const jsDomCssError = 'Error: Could not parse CSS stylesheet';
+console.error = (...params) => {
+    if (!params.find((p) => p.toString().includes(jsDomCssError))) {
+        originalConsoleError(...params);
+    }
+};
+}
```
J0taFerreira, line 12: After Primevue switched to CSS Layers the tests are having problems parsing the stylesheet. This deals with the errors on the console. I'm using this fix since a did not find a better one
https://github.com/primefaces/primevue/issues/4512
https://stackoverflow.com/questions/69906136/console-error-error-could-not-parse-css-stylesheet/69958999#69958999

MCatherine1994: yeah, just put a link here as our notes... I didn't find a better solution for now, most of the examples use this way. We can chat more tomorrow. thanks!! https://github.com/primefaces/primeng/issues/14085#issuecomment-1888629612

MCatherine1994: let's rename this file to be `fixJsdomCssErr`, it's actually the jsdom package is causing the error

frontend/src/tests/UserIdentityCard.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    beforeEach(async () => {
+        wrapper = mount(UserIdentityCard, {
+            props,
+            global: {
+                plugins: [PrimeVue],
+            },
+        });
+
+        card = wrapper.find('.custom-card');
+        cardEl = card.element as HTMLInputElement;
```
MCatherine1994, line 54: Hi Jota, do we use this one? if not could we remove it?

frontend/src/tests/UserIdentityCard.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        username = wrapper.find('#userId');
+        usernameEl = username.element as HTMLSpanElement;
+        firstName = wrapper.find('#firstName');
+        firstNameEl = firstName.element as HTMLSpanElement;
+        lastName = wrapper.find('#lastName');
+        lastNameEl = lastName.element as HTMLSpanElement;
+        checkmarkIcon = wrapper.find('#checkmarkIcon');
```
MCatherine1994, line 61: I think not all of these elements are used by both test cases, so maybe we don't need to put them all in beforeEach.

frontend/src/tests/UserIdentityCard.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        errorIcon = wrapper.find('#errorIcon');
+    });
+
+    it('Should show correct info on card based on props', () => {
+        //it renders
+        expect(card.element).toBeTruthy();
+
+        expect(usernameEl.textContent).toContain(props.userIdentity.userId);
+        expect(firstNameEl.textContent).toContain(props.userIdentity.firstName);
```
MCatherine1994, line 70: So here it can just be `expect(wrapper.find('#firstName').element.textContent).toContain(props.userIdentity.firstName);` something like this

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -28,21 +30,21 @@ const props = defineProps<{
...
+            <div class="col-2" v-if="props.userIdentity.found">
+                <label class="row" id="lastName">Last Name</label>
                 <span class="row">{{ props.userIdentity.lastName }}</span>
             </div>
             <div
                 class="col-6 d-flex"
                 v-if="!props.userIdentity.found"
             >
-                <span class="px-0 invalid"> User does not exist </span>
+                <span class="px-0 invalid" id="userNotExist"> User does not exist </span>
```
MCatherine1994, line 47: I'm fine to add ids this time, but normally we don't add id just because of testing. We could just verify the expect the text is rendered. And in the case we want it to be more specific, could also find by span and row and go through the return list. But normally as long as can find the text should be fine.

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        data: {
+            firstName: "Name",
+            found: true,
+            lastName: "LastName",
+            userId: 'userId'
+        },
+        status: 200,
+        statusText: 'Ok',
+        headers: {},
+        config: {},
```
MCatherine1994, line 27: I got a warning in VSCode, you might need to add `headers: {} as AxiosRequestHeaders`, `frontend/src/tests/ForestClientInput.spec.ts` has an example

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    let wrapper: VueWrapper;
+    let emitChange: unknown[][] | undefined;
+    let emitSetVerifyResult: unknown[][] | undefined;
+
+    let usernameInputText: DOMWrapper<HTMLElement>;
+    let usernameInputTextEl: HTMLInputElement;
+    let verifyButton: DOMWrapper<HTMLElement>;
+    let verifyButtonEl: HTMLButtonElement;
+
+    let cardUsernameEl: HTMLSpanElement;
```
MCatherine1994, line 42: we don't need to declare all variables at this level, if the variable is only used by a single test case, it can be declared inside the test case.

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        fieldId: 'newFieldId',
+    };
+
+    const newValue = 'testIdir';
+
+    beforeEach(async () => {
+        wrapper = mount(UserNameInput, {
+            props,
+            global: {
+                plugins: [PrimeVue],
```
MCatherine1994, line 63: I think you might try this to avoid the css error, but think we can remove it now.

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        expect(usernameInputTextEl.value).toBe(newValue);
+    });
+
+    it('Should receive the correct prop', async () => {
+        //default props
+        expect(wrapper.props()).toEqual(props);
+
+         await wrapper.setProps(newProps);
+         expect(wrapper.props()).toEqual(newProps);
+         expect(wrapper.props()).not.toEqual(props)
```
MCatherine1994, line 89: might need format the file, the space is not correct

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        verifyButtonEl = verifyButton.element as HTMLButtonElement;
+    });
+
+    afterEach(() => {
+        vi.clearAllMocks();
+        wrapper.unmount();
+    });
+
+    it('Should change usernameInput value', async () => {
+        await usernameInputText.setValue(newValue);
```
MCatherine1994, line 79: when we verify input change, we better verify it's empty at the beginning, and then it changes to the new value

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+
+         await wrapper.setProps(newProps);
+         expect(wrapper.props()).toEqual(newProps);
+         expect(wrapper.props()).not.toEqual(props)
+    });
+
+    it('Should call and emit correct value' , async () => {
+        await usernameInputText.setValue(newValue);
+        emitChange = wrapper.emitted('change');
+        expect(wrapper.emitted('change')).toBeTruthy();
```
MCatherine1994, line 95: let's do a review, so if we declare a variable, then let's use the variable

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        // test the given parameters when emitChange has been called
+        // i.e. emitChange = [ [ 'B' ] ]
+        expect(emitChange![0][0]).toEqual(newValue);
+
+        emitSetVerifyResult = wrapper.emitted('setVerifyResult')
+        await wrapper.setProps(newProps);
```
MCatherine1994, line 101: and maybe no need to set the props again?

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        vi.spyOn(
+            AppActlApiService.idirBceidProxyApi,
+            'idirSearch',
+        ).mockImplementation(async () => {
+            setLoadingState(true);
+            return userInputMock();
+        });
+
+        await usernameInputText.setValue(newValue);
+        expect(usernameInputTextEl.value).toBe(newValue);
```
MCatherine1994, line 129: maybe no need to set value here, cause only set user id props can enable the button. But we do want to check loading is false, and button text is not loading before trigger the click

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    it('Should remove card and emit different value when domain changes', async () => {
+        emitSetVerifyResult = wrapper.emitted('setVerifyResult');
+        // default props
+        expect(wrapper.props()).toEqual(props);
+
+        await usernameInputText.setValue(newValue);
+        expect(usernameInputTextEl.value).toEqual(newValue);
+
+        await wrapper.setProps(userInputMock().data);
```
MCatherine1994, line 149: um.... I'm not sure if we need to the above three lines... also here, you could just set the username. It's confused to change all props.

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+        await verifyButton.trigger('click');
+        await flushPromises();
+
+        verifiedUserIdentity = wrapper.findComponent({ name: 'UserIdentityCard'}).vm;
+        verifiedUserIdentity = userInputMock().data
+        cardUsernameEl = wrapper.find('#userId').element as HTMLSpanElement;
+        expect(cardUsernameEl).toBeTruthy();
+        expect(cardUsernameEl.textContent).toContain('userId');
+
```
MCatherine1994, line 158: We need another test for the user not found case.

frontend/src/tests/common/fixPrimevueCssErr.ts
J0taFerreira: Sure, I'll rename it!

frontend/src/tests/UserIdentityCard.spec.ts
J0taFerreira: Hello! I forgot to remove it. Thanks!

frontend/src/tests/UserIdentityCard.spec.ts
J0taFerreira: I'll refactor so that only used both are on beforeEach

frontend/src/tests/UserIdentityCard.spec.ts
J0taFerreira: I made the change to use this way

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: just did, thanks!

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll see which ones are in only one test and refactor

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: oh, that's correct! I forgot!

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: sure, I'll do that

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: without the props I would need to change the expect to look for the emit when domain is "I"

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll refactor it

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll add this test, thank you!

frontend/src/tests/UserIdentityCard.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+            firstName: 'First Name',
+            lastName: 'Last Name',
+        },
+    };
+
+    const propsNotFound = {
+        userIdentity: {
+            userId: 'userId',
+            found: false,
+            firstName: null,
```
MCatherine1994, line 32: in the not found case, I feel the return data might only have userId, and found false, no firstName and lastName, but better double check

frontend/src/tests/UserIdentityCard.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+            lastName: null,
+        },
+    };
+
+    beforeEach(async () => {
+        wrapper = mount(UserIdentityCard, {
+            props,
+            global: {
+                plugins: [PrimeVue],
```
MCatherine1994, line 41: do we still need this plugin?

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+            config: {
+                headers: {} as AxiosRequestHeaders,
+            },
+        };
+    }
+};
+
+describe('UserNameInput', () => {
+    let wrapper: VueWrapper;
+    let emitSetVerifyResult: unknown[][] | undefined;
```
MCatherine1994, line 48: we don't need to declare it here, cause it's not shared between test cases, and we declare it in each test case already,  feel we can remove it, thanks!

frontend/src/tests/UsernameInput.spec.ts
MCatherine1994: um.... I think the wrapper by default is rendered with props with domain "I"?

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: Do you mean that I should declare the emit emitSetVerifyResult on the two different tests?

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: yes, it is domain "I" by default

frontend/src/tests/UsernameInput.spec.ts
MCatherine1994: yes, just remove from there, because we do `emitSetVerifyResult = wrapper.emitted('setVerifyResult');` in the test cases, and this is not sharable.

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: oh! I see, I'll fix it!

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: did not know that, thanks! I'll remove it

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        emitSetVerifyResult = wrapper.emitted('setVerifyResult');
+        await wrapper.setProps(newProps);
+        await verifyButton.trigger('click');
+
+        expect(emitSetVerifyResult).toBeTruthy();
+        expect(emitSetVerifyResult![1][0]).toEqual(true);
+    });
+
+    it('Should enable verify btn when username is inputted', async () => {
+        // button starts as disabled
```
MCatherine1994, line 118: here we need to expect the input is empty initially

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    it('Should show loading on the verify btn while we call the api', async () => {
+        vi.spyOn(
+            AppActlApiService.idirBceidProxyApi,
+            'idirSearch'
+        ).mockImplementation(async () => {
+            setLoadingState(true);
+            return userInputMock(true);
+        });
+
+        setLoadingState(false);
```
MCatherine1994, line 137: we might don't need to set loading to be false, by default it should be false

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        await verifyButton.trigger('click');
+        expect(isLoading()).toBe(true);
+        expect(verifyButtonEl.textContent).toContain('Loading');
+
+        // cleanup state variable
+        setLoadingState(false);
+        expect(isLoading()).toBe(false);
+    });
+
+    it('Should show not found on card when user is not verified', async () => {
```
MCatherine1994, line 150: just the title, maybe when user is not found?

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        vi.spyOn(
+            AppActlApiService.idirBceidProxyApi,
+            'idirSearch'
+        ).mockImplementation(async () => {
+            return userInputMock(false);
+        });
+
+        await wrapper.setProps({ userId: newProps.userId });
```
MCatherine1994, line 158: let's add a comment here, to remind ourselves, this is used to enable the veirfyButton

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        await verifyButton.trigger('click');
+        await flushPromises();
+
+        userIdentity = wrapper.findComponent({
```
MCatherine1994, line 162: and we only use this userIdentity variable in this test case, let's remove it from the very top, and declare here `const userIdentity  = ....`
Screen Shot 2024-01-31 at 2 32 01 PM

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+            name: 'UserIdentityCard',
+        }).vm;
+
+        const cardEl = wrapper.find('.custom-card').element as HTMLSpanElement;
+        expect(cardEl).toBeTruthy();
+        expect(cardEl.textContent).toContain('User does not exist');
+    });
+
```
MCatherine1994, line 170: do we have a test to check the successful case when user is found? just saw we tested the not found case.

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    it('Should remove card and emit different value when domain changes', async () => {
+        let cardUsernameEl: HTMLSpanElement;
```
MCatherine1994, line 172: we don't have to declare it here, just do `const cardUsernameEI: HTMLSpanElement = ...` when you use it on line 180

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        // default props
+        expect(wrapper.props()).toEqual(props);
+
+        await wrapper.setProps(userInputMock(true).data);
+        await verifyButton.trigger('click');
+        await flushPromises();
+
+        cardUsernameEl = wrapper.find('#userId').element as HTMLSpanElement;
+        expect(cardUsernameEl).toBeTruthy();
+        expect(cardUsernameEl.textContent).toContain('userId');
```
MCatherine1994, line 182: let's add some comments here and do some cleanup, so no need to test the default props, and the whole test could just be:
```
         // show user identity card to prepare for the test 
         await wrapper.setProps({ userId: newProps.userId });
         await verifyButton.trigger('click');
         await flushPromises();
         const cardUsernameEl = wrapper.find('#userId').element as HTMLSpanElement;
         expect(cardUsernameEl).toBeTruthy();
         
         // change the domain to be B
         await wrapper.setProps({domain: UserType.B});
         // for BCeID should emit true
         expect(emitSetVerifyResult![1][0]).toEqual(true);
         // UserIdentityCard not on page anymore
         expect(wrapper.findAll('#UserIdentityCard')).toHaveLength(0);

         // change the domain to be I
         await wrapper.setProps({domain: UserType.I});
         // for IDIR should emit false
         expect(emitSetVerifyResult![1][0]).toEqual(false); 
```

MCatherine1994: just a reminder to format the file again after the fixes, thanks so much!!

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I can add this test

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: sure, will do

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll remove it

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll change to "user is not found"

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: sure

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: I'll do that! Thank you!

MCatherine1994 (approved):

PR #1157, opened 2024-01-26
fix: admin management api minor refactoring

ianliuwk1019:
- Rename some tests filenames.
- Add comments for schema.py for convention.
- Add suffix for some schema objects.
- Update admin management api openapi.json.
- Recreate admin management api client code.
- Configure frontend for FAMAccessControlPrivilegeApi
- Fix frontend issue due to backend schema change.
- Use service instance dependency in routers.


server/admin_management/api/app/schemas.py
```diff
@@ -57,60 +65,60 @@ class FamUserBase(BaseModel):
...
 
-class FamUserCreate(FamUserBase):
+class FamUserDto(FamUserBase):
     create_user: Annotated[str, StringConstraints(max_length=60)]
 
     model_config = ConfigDict(from_attributes=True)
 
 
-class FamUserType(BaseModel):
+class FamUserTypeDto(BaseModel):
```
MCatherine1994, line 74: Hi Ian, I wonder if "DTO" schema is when we use to create the record in db? or it's just a naming convention, if it's not used at router level, then we can call it with "Dto"?

MCatherine1994:
thanks Ian! I just got a question, looks good!

MCatherine1994 (approved):

server/admin_management/api/app/schemas.py
ianliuwk1019: Hi @MCatherine1994 , in layered pattern, "DTO" object is a "transfer" object. It is **not specific** to repository insert/update/select. It only means a **transfer object between layers** (presentation layer, service layer, repository layer etc...). Normally you would see DTO lives in service layer **a lot** but having conversion from presentation layer and repository layer. Layered pattern uses it to abstract away business/responsibility/dependency between each layer.
Although I stick to convention using "DTO", but in python, our code, we did not have strict disinclination for layers yet (and probably does not need to go that far). If you prefer, we can drop the "DTO" and just keep the (__request and __response) if they are really being used at endpoints.

MCatherine1994: I'm fine with it, thanks Ian!!

PR #1162, opened 2024-01-30
feat: #1059 application admin management UI

NickSaglioni:
- Implemented new page `GrantApplicationAdmin` to handle Application Admin permissions;
- Created a new component `ApplicationSelect` that is shared in two different pages (`ManagePermissions` and `GrantApplicationAdmin`);
- Accessing new admin endpoints to get all applications and creating app admin;
- Updated router to include new route `grant-app-admin`


frontend/src/router/index.ts
```diff
@@ -84,6 +85,19 @@ const routes = [
             };
         },
     },
+    {
+        path: routeItems.grantAppAdmin.path,
+        name: routeItems.grantAppAdmin.name,
+        meta: {
+            requiresAuth: true,
+            requiresAppSelected: false,
```
MCatherine1994, line 93: Hi Nick, I think we require app to be selected

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
+<script setup lang="ts">
+import { onMounted, ref } from 'vue';
+import { ErrorMessage, Field, Form as VeeForm } from 'vee-validate';
+import { object, string } from 'yup';
+import router from '@/router';
+import { AdminMgmtApiService } from '@/services/ApiServiceFactory';
+import { UserType, type FamApplication } from 'fam-app-acsctl-api';
+import type { FamAppAdminCreateRequest } from 'fam-admin-mgmt-api/model/fam-app-admin-create-request';
+import type { FamApplicationGetResponse } from 'fam-admin-mgmt-api/model/fam-application-get-response';
+import ApplicationSelect from '@/components/common/form/ApplicationSelect.vue';
```
MCatherine1994, line 10: Hi Nick, just wondering did you add this file? I couldn't see it `components/common/form/ApplicationSelect.vue`

NickSaglioni: Thanks Catherine. I was refactoring the file location and forgot to re-add it in git. Fixed now.

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -91,7 +84,7 @@ const deleteAdmin = (admin: FamAppAdminGetResponse) => {
         <div class="custom-data-table">
             <DataTableHeader
                 btnLabel="Add application admin"
-                :btnRoute="routeItems.grantUserPermission.path"
+                :btnRoute="routeItems.grantAppAdmin.path"
```
J0taFerreira, line 87: Hello, Nick!
I think you forgot to add this path to the side nav as well

NickSaglioni: Nice find! Thanks, Jota!

MCatherine1994:
Hi Nick, I got this warning in the console, not sure if you get the same or just me?
Screen Shot 2024-01-31 at 12 42 55 PM

MCatherine1994:
Also I think the space between the title and the application dropdown is different than what's in the FAM dev, and I saw an extra space at the bottom. So I think it's when refactor the application select component, the style change might impact that. 
image

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
...
+        as="div"
+    >
+        <div class="page-body">
+            <form id="grantAccessForm" class="form-container">
+                <StepContainer
+                    title="User information"
+                    subtitle="Enter the user information to add a new admin to [Application name]"
+                >
+                    <UserNameInput
+                        domain="I"
```
MCatherine1994, line 123: let's use the `import { UserType } from 'fam-app-acsctl-api';` and `UserType.I`

MCatherine1994:
We don't need to make change to this pr, but just a question for @OlgaLiber2, I wondering if we removed user's admin access of an application, do we want to say which application in the notification message?

like `You removed IANLIU's admin privilege of FOM_TEST`?
Screen Shot 2024-01-31 at 1 11 43 PM

MCatherine1994:
Hi Nick, just wondering maybe no need ":" at the end? thanks!!
Screen Shot 2024-01-31 at 1 24 39 PM

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
...
+const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
+const formValidationSchema = object({
+    userId: string()
+        .required('User ID is required')
+        .min(2, 'User ID must be at least 2 characters')
+        .nullable(),
+    application: object().required('Application is required'),
+});
+
+const selectedApplication = ref<FamApplication>();
```
MCatherine1994, line 31: I wonder if we could rename this variable maybe? cause it's the same name as the one in our application state store. Or I wonder if we could use `formData.application` directly, without adding an extra variable

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
...
+const applications = ref<FamApplicationGetResponse[]>(); // clone default input
+
+onMounted(async () => {
+    applications.value = (
+        await AdminMgmtApiService.applicationsApi.getApplications()
```
MCatherine1994, line 36: Just a note here, we should make the call to get applications in the router handler, same as the `beforeEnterGrantUserPermissionRoute`, where it grabs all roles. However since later on we'll use the new "get_my_fam_access" api to cache the result for all applications, I'm fine to leave it here, we'll change it anyway later. This is just for your reference that how we handle the on mounted call currently. 

image

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
...
+        ref="form"
+        v-slot="{ errors, meta }"
+        :validation-schema="formValidationSchema"
+        as="div"
+    >
+        <div class="page-body">
+            <form id="grantAccessForm" class="form-container">
+                <StepContainer
+                    title="User information"
+                    subtitle="Enter the user information to add a new admin to [Application name]"
```
MCatherine1994, line 120: the [Application name] need to be the real application name
Screen Shot 2024-01-31 at 1 43 42 PM

frontend/src/components/common/form/ApplicationSelect.vue
```diff
@@ -0,0 +1,67 @@
...
+const model = ref<FamApplication>();
+
+const emit = defineEmits(['onApplicationSelected']);
+
+onMounted(() => {
+    model.value = props.model;
+});
+
+const onApplicationSelected = async (e: DropdownChangeEvent) => {
+    emit('onApplicationSelected', e);
```
MCatherine1994, line 33: Hi Nick, somehow I feel we're missing something, even though this component seems work fine. However, when we select application, we make the emit call, but your local state variable "model" is not changed. We only set the local model when on mounted to let it equals props.model. But `onApplicationSelected`, the emit method will update props.model, but there is no where to update the local model?

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,182 @@
...
+import Button from '@/components/common/Button.vue';
+import { IconSize } from '@/enum/IconEnum';
+import { Severity } from '@/enum/SeverityEnum';
+
+import { isLoading } from '@/store/LoadingState';
+import { setNotificationMsg } from '@/store/NotificationState';
+
+const defaultFormData = {
+    userId: '',
+    application: null as number | null,
```
MCatherine1994, line 20: this type might not be correct, cause later one we set it to equals the value returns from dropdown selection

```
const onApplicationSelected = async (e: any) => {
    formData.value.application = e.value;
};
```

frontend/src/components/common/form/ApplicationSelect.vue
```diff
@@ -0,0 +1,67 @@
+<script setup lang="ts">
```
MCatherine1994, line 1: I feel it's okay to put this file under common, no need to create a sub form folder, thanks!!

frontend/src/router/routeHandlers.ts
```diff
@@ -71,9 +71,17 @@ const beforeEnterGrantUserPermissionRoute = async (
     return true;
 };
 
+const beforeEnterGrantApplicationAdminRoute = async (
+    to: RouteLocationNormalized
+) => {
+    populateBreadcrumb([routeItems.dashboard, routeItems.grantAppAdmin]);
```
MCatherine1994, line 77: I think before enter this router, we have to check the selected application is FAM. Cause only FAM admin is allowed to go to this route. 

In the below example, I selected FOM_DEV, but I go to the `/grant-app-admin` router directly, it still lets me in, it's super confused. So for this ticket, let's just restrict only allow enter this router when the selected application is FAM. 
 Screen Shot 2024-01-31 at 2 05 29 PM
Screen Shot 2024-01-31 at 2 06 18 PM

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
NickSaglioni: Thanks, I'll change it.

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
NickSaglioni: I am so sorry about that. I don't know how I could possibly have let this pass. I'll fix it right away.

frontend/src/components/common/form/ApplicationSelect.vue
NickSaglioni: Ok, I'll refactor it. Thanks.

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
NickSaglioni: Thanks Catherine, the variable is not needed indeed. I'll remove it.

frontend/src/components/common/form/ApplicationSelect.vue
NickSaglioni: Your suggestion does work, I'll fix it. Thanks.

frontend/src/router/routeHandlers.ts
NickSaglioni: I agree with you. I haven't thought about it, my mistake. I'll update it to restrict access.

ianliuwk1019: Most of the routes should have checks/guards to protect at very beginning. But just to inform, the guards (including this one) on frontend will be shift on its logic soon due to our new access change for users.

frontend/src/components/common/form/ApplicationSelect.vue
```diff
@@ -0,0 +1,67 @@
...
+        required: true,
+    },
+    placeholder: {
+        type: String,
+        required: false,
+        default: 'Choose an application',
+    },
+});
+
+const model = ref<FamApplication>();
```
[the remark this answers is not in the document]
NickSaglioni: This is a nice suggestion but the defineModel() was first introduced in `Vue 3.3`. I believe we are currently using `Vue 3.2.38`, maybe it could be refactored in the future when we bump our Vue version.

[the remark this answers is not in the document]
ianliuwk1019: If the new ApplicationSelect component is working fine, and if you prefer, then can do that in additional ticket.
Our frontend uses `3.4.6` currently, so it is available.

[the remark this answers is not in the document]
NickSaglioni: Oh, I see. Could I update the Vue version number in `package.json`? I got confused by it. @MCatherine1994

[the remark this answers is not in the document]
NickSaglioni: Thanks for the clarification everyone!

ianliuwk1019:
It's minor, but I see this on this dimension.
image

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,178 @@
...
+import Button from '@/components/common/Button.vue';
+import { IconSize } from '@/enum/IconEnum';
+import { Severity } from '@/enum/SeverityEnum';
+
+import { isLoading } from '@/store/LoadingState';
+import { setNotificationMsg } from '@/store/NotificationState';
+
+const defaultFormData = {
+    userId: '',
+    application: Number,
```
ianliuwk1019, line 20: You might be already fixing this...
I saw this "Invalid Prop" when I click cancel button.
image

MCatherine1994: I got the same

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -46,24 +50,32 @@ const userRoleAssignments = shallowRef<FamApplicationUserRoleAssignmentGet[]>(
...
+);
 
 onUnmounted(() => {
     resetNotification();
 });
 
-const onApplicationSelected = async (e: DropdownChangeEvent) => {
-    setSelectedApplication(e.value ? JSON.stringify(e.value) : null);
-    if (e.value.application_id === FAM_APPLICATION_ID) {
+watch(manageApplicationSelected, async (newManageApplicationSelected) => {
```
MCatherine1994, line 65: I wonder maybe let's not refactor this application dropdown to a reusable component? cause I feel it complicated the thing a little bit... the one we use here will change a global state variable, the other place we use for grant app admin will change the local form variable. I think maybe it's more clear to just keep them separated.  Cause we use the primevue dropdown anyway, and just pass different data. And if we want to apply the same style, we can give the same class name, and move the style to layout.scss maybe or somewhere that they can share.  Not sure how you feel? thanks!!

NickSaglioni: It seemed a good idea at first to make it a reusable component but I think I agree with you that the complexity has increased a bit. I'll revert it back to what it was and change the new page accordingly. Thanks.

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
NickSaglioni: As the component does not exist anymore, this warning is gone.

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -0,0 +1,188 @@
...
+                    formData.value.application.application_name
+                }`
+            );
+        })
+        .catch((error) => {
+            if (error.response?.status === 409) {
+                setNotificationMsg(
+                    Severity.warning,
+                    error.response.data.detail
+                );
```
ianliuwk1019, line 81: For now, maybe it is ok, but this logic inside catch() block look strange: on this `409` case and on `else { }` block at line 90.
If the backend returns errors and are generic enough for frontend and does not need to be processed, frontend components are not necessary to handle them individually.

ianliuwk1019 (approved):

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
NickSaglioni: I agree, maybe this could be a conversation to improve the error notification handling.

MCatherine1994 (approved):

Issue #1086, opened 2023-12-11
Add delegated admin endpoints to user management api


MCatherine1994:
**Describe the task**
FAM needs a mechanism to create and delete delegated administrator records.

- [ ] Create a POST endpoint that allows the creation of a new record in fam_access_control_privilege
- [ ] Create a DELETE endpoint that allows the deletion of an existing record in fam_access_control_privilege
- [ ] Create a GET endpoint that gets a list of delegated admin users for an application
- [ ] Make sure that all the endpoints require the application admin role 
- [ ] Make sure that the user can't manage their own records
- [ ] Make sure that POST and DELETE actions are audited similar to the current role POST and DELETE actions
- [ ] Test locally

OlgaLiber2:
@MCatherine1994 split this ticket into 3 tickets: 
- 1 ticket for GET 
- 1 ticket for POST 
- 1 ticket for DELETE

include tests for each of these 3 tickets

MCatherine1994:
The return result looks like below:
```
[
  {
    "access_control_privilege_id": 1,
    "user_id": 4,
    "role_id": 92,
    "user": {
      "user_name": "IANLIU",
      "user_type": {
        "code": "I",
        "description": "IDIR"
      }
    },
    "role": {
      "role_id": 92,
      "role_name": "FOM_SUBMITTER_00001011",
      "client_number": {
        "forest_client_number": "00001011"
      },
      "parent_role": {
        "role_name": "FOM_SUBMITTER",
        "role_type_code": "A"
      }
    }
  },
  {
    "access_control_privilege_id": 2,
    "user_id": 4,
    "role_id": 93,
    "user": {
      "user_name": "IANLIU",
      "user_type": {
        "code": "I",
        "description": "IDIR"
      }
    },
    "role": {
      "role_id": 93,
      "role_name": "FOM_SUBMITTER_00001013",
      "client_number": {
        "forest_client_number": "00001013"
      },
      "parent_role": {
        "role_name": "FOM_SUBMITTER",
        "role_type_code": "A"
      }
    }
  }
]
```

MCatherine1994:
Create a follow up ticket #1172 to update router name for admin management to follow the [API naming standard](https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/FAM+API+Naming+Standard)

PR #1167, opened 2024-01-30
feat: #1086 get delegated admin endpoint

MCatherine1994:
refs: #1086

An example return:
```
[
  {
    "access_control_privilege_id": 1,
    "user_id": 4,
    "role_id": 92,
    "user": {
      "user_name": "IANLIU",
      "user_type": {
        "code": "I",
        "description": "IDIR"
      }
    },
    "role": {
      "role_id": 92,
      "role_name": "FOM_SUBMITTER_00001011",
      "client_number": {
        "forest_client_number": "00001011"
      },
      "parent_role": {
        "role_name": "FOM_SUBMITTER",
        "role_type_code": "A"
      }
    }
  },
  {
    "access_control_privilege_id": 2,
    "user_id": 4,
    "role_id": 93,
    "user": {
      "user_name": "IANLIU",
      "user_type": {
        "code": "I",
        "description": "IDIR"
      }
    },
    "role": {
      "role_id": 93,
      "role_name": "FOM_SUBMITTER_00001013",
      "client_number": {
        "forest_client_number": "00001013"
      },
      "parent_role": {
        "role_name": "FOM_SUBMITTER",
        "role_type_code": "A"
      }
    }
  }
]
```


server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -99,3 +105,23 @@ def create_access_control_privilege_many(
             )
 
         audit_event_log.log_event()
+
+
+@router.get(
+    "/{application_id}/access_control_privileges",
```
ianliuwk1019, line 111: Hi Catherine, we probably like to move this endpoint to other router (application) or at least rename the base-path and sub-path for better sticking to the **REST standard** as this does not follow...
image
I myself find in this case is also blur to what REST should be followed...

Some suggestions:
These two probably are more follow from standard.
```
GET: /applications/{application_id}/delegated-admins/privileges
GET: /access_control_privileges?{application_id}
```

These two I am not sure myself if they can are acceptable:
```
GET: /access_control_privileges/applications/{application_id}/privileges
GET: /delegated-admin-management/applications/{application_id}/privileges
```
Maybe @basilv can provide some good ideas? Thanks!

ianliuwk1019 (changes requested):

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -99,3 +105,23 @@ def create_access_control_privilege_many(
...
+def get_access_control_privileges_by_application_id(
+    application_id: int,
+    token_claims: dict = Depends(jwt_validation.authorize),
+    access_control_privilege_service: AccessControlPrivilegeService = Depends(
+        access_control_privilege_service_instance
+    ),
+):
+    LOGGER.debug(
+        f"Get access control privileges for application: {application_id}, requestor: {token_claims}"
+    )
```
ianliuwk1019, line 125: We don't need this debug log I think. Also `requester: {token_claims}` does not match I think.

server/admin_management/tests/repositories/test_access_control_privilege_repository.py
```diff
@@ -120,3 +122,20 @@ def test_create_access_control_privilege_invalid_role_id(
             )
         )
     assert str(e.value).find(ERROR_VOLIATE_FOREIGN_KEY_CONSTRAINT) != -1
+
+
+def test_get_acp_by_application_id(access_control_privilege_repo: AccessControlPrivilegeRepository,):
```
ianliuwk1019, line 127: Maybe it does not matter, there is a "," at the end on the closing ")"

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: we might still want to keep it under this router, cause the return result is a list of users with roles, not users and applications. I feel it makes sense to keep under this router. But we could rename it, `access_control_privileges/{application_id}/privileges`? I also think about the path before but has no good idea....

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -99,3 +105,23 @@ def create_access_control_privilege_many(
             )
 
         audit_event_log.log_event()
+
+
+@router.get(
+    "/{application_id}/access_control_privileges",
+    response_model=List[schemas.FamAccessControlPrivilegeDto],
```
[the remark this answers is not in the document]
ianliuwk1019: Yes, that's fine. I did put a file comment for the convention at the schem.py.

server/admin_management/api/app/routers/router_access_control_privilege.py
ianliuwk1019: By standard, if `access_control_privileges` is the **REST "resource"**, then the path **"/{application_id}"** isn't correct, that would indicate the "**id**" to identify that REST **resource**. (Just like on the network tab, you will only see "/access_control_privileges/**19**/" and that "19" isn't (`application_id`) for others ( it is the resource for `access_control_privileges.get(id=19)` ).

server/admin_management/api/app/routers/router_access_control_privilege.py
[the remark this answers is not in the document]
MCatherine1994: thanks Ian, just checked, I'll update

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: `access_control_privileges/privileges/?application_id={application_id}` ?

MCatherine1994: discussed with team, use `/access_control_privileges?application_id={application_id}` and `/application_admins?application_id={application_id}`

MCatherine1994: image

basilv: There's one more level of potential debate in terms of the actual resource naming. /application_admins isn't actually returning a list of admin users, its returning the access held by these users. So perhaps /application_admin_accesses

/access_control_privileges reflects access, not users so that's more in line, but could be more clearly named /delegated_admin_accesses to better align with our domain language.

ianliuwk1019: I feel the same, "delegated_admin_accesses" is a better name, and the existing endpoints was inherited from the past using python database entity name, with not really a naming convention followed from domain language.
However, probably up to Catherine if she wants to change or leave it to the new ticket to change them all together (will impact frontend).

ianliuwk1019 (approved):

ianliuwk1019 (approved):

Issue #1121, opened 2024-01-11
Breadcrumb Component Not Routing


J0taFerreira:
**Describe the bug**
The breadcrumb component is not routing the user to Manage Permissions when clicked.

**To Reproduce**
1. Go to 'Add User Permission'
2. Click on Manage Permission link present on the breadcrumb

![breadcrumb-bug](https://github.com/bcgov/nr-forests-access-management/assets/126905995/5d951293-cead-459d-b048-57a6b2574391)

**Additional context**
I believe the problem is the routeItem.ts file in the path of routeItems, if I'm not mistaken primevue expects an object with a "to" instead of "path".

J0taFerreira:
Primevue introduced breaking changes when vue-router support from Menu components was removed in version 3.40.0.
Note: For some reason the breaking changes was not mentioned in the 3.40 changelog.

some useful links about the changes:
https://github.com/primefaces/primevue/releases/tag/3.40.0
https://github.com/primefaces/primevue/issues/4739
https://github.com/primefaces/primevue/issues/4633#issuecomment-1827780322

PR #1169, opened 2024-02-01
fix: #1121 breadcrumb component not routing

J0taFerreira:
- Fixed breadcrumb component not routing.

The breadcrumb component now receives the home as a prop, in the primevue breadcrumb component documentation has exemples using the name "`route`" for the the path but the component it self expects "`item.url`". Tried using "`route`" and "`path`" as the `routeItems` object key but it did not work, "`url`" worked though.

https://github.com/primefaces/primevue/commit/ad9b437003e0d3ae26047d2cfd1a6a6ecc9037d7#diff-33bf4bb142b615a34557b44002c2eb3ac4393bd3fa4099aede38fabbf4e5527f
https://primevue.org/breadcrumb/

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -8,7 +8,7 @@
             "name": "fam-frontend",
             "version": "0.0.0",
             "dependencies": {
-                "@bcgov-nr/nr-theme": "^1.8.3",
+                "@bcgov-nr/nr-theme": "^1.8.4",
                 "@carbon/icons-vue": "^10.72.1",
                 "aws-amplify": "^5.3.6",
                 "axios": "1.6.5",
@@ -7638,9 +7638,9 @@
             }
         },
         "node_modules/@bcgov-nr/nr-theme": {
-            "version": "1.8.3",
-            "resolved": "https://registry.npmjs.org/@bcgov-nr/nr-theme/-/nr-theme-1.8.3.tgz",
-            "integrity": "sha512-8RPcZew0TDM/Rek4nBgMwCoNl2iL+i+ksrBXhOLpgIco2Vagus7b+oB/99VNlIjktiNMUep9/be2f6/1TGgQrA=="
+            "version": "1.8.4",
+            "resolved": "https://registry.npmjs.org/@bcgov-nr/nr-theme/-/nr-theme-1.8.4.tgz",
+            "integrity": "sha512-otXy8RsKZ1GdMIGlJRP5XvY5AdxkkNFI74DQYJ4hpz4ED7XYDpDM9Ab4C+w935HmWpAtaTMX/1W6f6nA8PAYyw=="
         },
         "node_modules/@bcoe/v8-coverage": {
             "version": "0.2.3",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,7 +15,7 @@
         "test-coverage": "vitest run --coverage"
     },
     "dependencies": {
-        "@bcgov-nr/nr-theme": "^1.8.3",
+        "@bcgov-nr/nr-theme": "^1.8.4",
         "@carbon/icons-vue": "^10.72.1",
         "aws-amplify": "^5.3.6",
         "axios": "1.6.5",
--- a/frontend/src/assets/styles/default-theme.scss
+++ b/frontend/src/assets/styles/default-theme.scss
@@ -20,7 +20,6 @@ $dark-text-primary: map.get(darkTheme.$dark-theme, 'text-primary');
 $light-layer-selected-01: map.get(lightTheme.$light-theme, 'layer-selected-01');
 $light-link-primary: map.get(lightTheme.$light-theme, 'link-primary');
 $light-link-primary-hover: map.get(lightTheme.$light-theme, 'link-primary-hover');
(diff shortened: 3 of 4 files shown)


MCatherine1994:
Hi Jota, I think there is an example in the primevue documentation https://primevue.org/breadcrumb/#router, we can follow that. So change the route item "path" to "route".  Also need to change all the places is using ".path"
image

And update the Breadcrumb component in the `frontend/src/components/common/PageTitle.vue` file to something like below, it should work. Then we don't have to add additional home props.
image

J0taFerreira:
> Hi Jota, I think there is an example in the primevue documentation https://primevue.org/breadcrumb/#router, we can follow that. So change the route item "path" to "route". Also need to change all the places that is using ".path" image
> 
> And update the Breadcrumb component in the `frontend/src/components/common/PageTitle.vue` file to something like below, it should work. Then we don't have to add additional home props. Just a reminder, in this example we might don't need all of it, like some classname, let's just add what we need. Thanks so much!! image

route will work if we use the template, but it might broke some styling. I need to check to be sure

J0taFerreira:
> > Hi Jota, I think there is an example in the primevue documentation https://primevue.org/breadcrumb/#router, we can follow that. So change the route item "path" to "route". Also need to change all the places that is using ".path" image
> > And update the Breadcrumb component in the `frontend/src/components/common/PageTitle.vue` file to something like below, it should work. Then we don't have to add additional home props. Just a reminder, in this example we might don't need all of it, like some classname, let's just add what we need. Thanks so much!! image
> 
> "route" or "path" will work using the template , but it will break styling. I need to check to be sure

using the template sintax works with "path" as the routeItems key but will break the styling. Should I make the change and focus on the styling?

NickSaglioni (dismissed):
Looks good to me.

MCatherine1994:
Just add a note here, seems like Primevue breadcrumb can get the item.url and make the link works, so we changed "path" to be "url": https://github.com/primefaces/primevue/commit/ad9b437003e0d3ae26047d2cfd1a6a6ecc9037d7#diff-33bf4bb142b615a34557b44002c2eb3ac4393bd3fa4099aede38fabbf4e5527f

MCatherine1994:
Hi Jota, I think there are two more places is using ".path", we have to change them, otherwise like the sidebar navigation is not working
Screen Shot 2024-02-02 at 2 31 11 PM

frontend/src/router/index.ts
```diff
@@ -47,7 +47,7 @@ const routes = [
         component: LandingView,
     },
     {
-        path: routeItems.dashboard.path,
+        path: routeItems.dashboard.url,
```
ianliuwk1019, line 50: I am not sure I understand why changing from the key name from ".path" to ".url" matter in this breadcrumb issue.

1. To me, the key name shouldn't be "**url**" from the "**route**Items", since this is for **Vue Router**; and Vue Router is correct to use "**path**" as the key name, so even we supply from the routerItems object for the value, should still call it "**path**".

2. **Route's** "**path**" isn't a "**url**", path and url are different. **URL** contains **many parts** and is a link that you are able to access with it while "**path**" is from within the Route and is part of the URL. Mixing them isn't a good idea.

ianliuwk1019 (changes requested):

frontend/src/router/index.ts
J0taFerreira: Hello, Ian! The options were discussed at a past scrum. To use any other key name other than "url" we would need to use the templating syntax and that would break the style, The 'url' key name is expected by the primevue Breadcrumb component,
https://github.com/bcgov/nr-forests-access-management/pull/1169#issuecomment-1924803550

ianliuwk1019: Jota, I wasn't so sure what you were discussing last time so didn't comment on that. Primevue provides flexible options, including "url" (for different purpose), it does not mean we should use it to make app work (and I suspect we used the wrong option initially before this issue). If fixing style is expected, then we should fix styling. We should discuss again after scrum. 
@MCatherine1994

MCatherine1994: Hi Jota, I had a quick chat with Ian to have more understanding of this breadcrumb component. So we feel it's better to follow their example. Cause if we change to use "url", it feels like we mixed the internal link vs external link. "route" is for internal routes, and "url" is for external links. Even it works fine for now, but long term we might feel confused again. So let's switch to follow this example and fix our style, thanks so much and sorry for the confusion!!

image

J0taFerreira: I'll make the change to template.

MCatherine1994:
Hi Jota, please check the comment I put in the pr in the common repo, https://github.com/bcgov/nr-theme/pull/155#pullrequestreview-1869058420. We don't need to add these extra custom class, let's remove them, and upgrade to the latest theme package. Thanks!!

ianliuwk1019 (dismissed):
Approve. Just wondering if the addition of "$light-focus" in .scss is intended or is a left over?

J0taFerreira:
> Approve. Just wondering if the addition of "$light-focus" in .scss is intended or is a left over?

thanks Ian, just removed

MCatherine1994 (approved):

Issue #1139, opened 2024-01-20
Add forest client validator to admin management api


MCatherine1994:
**Describe the task**
We need to add forest client validator for admin management apis, so when create delegated admin access control privilege, we verify forest client number on backend as well.

**Acceptance Criteria**
- [x] Integrate with forest client api
- [x] Add forest client validator when create access control privilege
- [ ] ~Regenerate client code~ no difference
- [x] Add test

**Additional context**
- Can just copy over our existing forest client code under backend folder
- This is the left work from ticket #1089

PR #1176, opened 2024-02-02
feat: #1139 add forest client validator

MCatherine1994:
Integrate with forest client api to verify forest client number when creating access control privilege for delegated admin
refs: #1139

When create the access control privileges with multiple forest client numbers, as long as one of the forest client number is inactive or invalid, return the validation error
Screen Shot 2024-02-06 at 3 32 54 PM


server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -157,3 +176,45 @@ def grant_privilege(
             )
 
         return access_control_privilege_return
+
+
+class ForestClientValidator:
```
ianliuwk1019, line 181: Wondering maybe you like to have this new validator into a folder "/services/validators" so it does not need to be located in this service? Later it may come in handy for business rule validations for other services.

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -157,3 +176,45 @@ def grant_privilege(
...
+    """
+    Purpose: More validations on inputs (other than basic validations) and
+             business rules validations (if any).
```
ianliuwk1019, line 184: Probably need to adjust this comment from previous one.
Previous validator was intended to validate additional business rules for the whole request object for the endpoint but here it is only for forest_client_number validation.

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -157,3 +176,45 @@ def grant_privilege(
...
+    Cautious: Do not instantiate the class for more than one time per request.
+              It calls Forest Client API remotely if needs to.
+    """
```
ianliuwk1019, line 187: In access_control_privilege_service, it has a **loop** and instantiates this "**ForestClientValidator**" several times depending on number of the forest_client_number in the user requests.

I wonder how slow it is for this looping, since Forest Client API itself response is already slow enough....plus now each loop it needs to do a new extra instantiation for this  "ForestClientValidator" class and "ForestClientService" class as well for the setup for the whole classes...

In normal usage, a "service"(validator is also a service) is actually designed to be used as a **singleton** per request (which currently our python code does not enforce it yet).

Maybe you like to adjust the validator class a little so that you can move this validator instantiation outside of the loop at line 84 `validator = ForestClientValidator(forest_client_number)` and only pass forest_client_number to a validate function of the class.
Or you can have this class take an array of forest_client_number (including 1 item) and have the looping logic inside the class.

MCatherine1994: Hi Ian, thanks! that's a good point, I forgot it's a loop now, I'll try to make some adjustment!

server/admin_management/api/app/services/access_control_privilege_service.py
MCatherine1994: updated

server/admin_management/api/app/services/access_control_privilege_service.py
MCatherine1994: Hi Ian, I updated the validator do you mind check it again? thanks!!

ianliuwk1019 (approved):

Issue #820, opened 2023-08-23
Discuss how backend return error message for frontend to handle


MCatherine1994:
**Describe the task**
Currently our backend doesn't have a clear design of how we throw errors back to frontend to handle. We have different 401 and 403 errors, but not all of them are business logic errors that need to show to the frontend user. 409 errors so far seems like is intended for business error that should show in the frontend. But we need further discussion. 

**Acceptance Criteria**
- [ ] Have a clear guide when to throw 401, 403 and 409 errors. Return an error code, so frontend can add custom error message based on the error code
- [ ] Frontend knows how to handle the different error cases
- [ ] work with UX to make sure front end errors are descriptive and clear to the user

**Additional context**
- Further discussion resulting from ticket #735
- Frontend code `NotificationState.ts` is very confusing and hard to understand, redesign/refactor is needed or at least clear document should be added.

PR #1192, opened 2024-02-08
fix: fixing bugs from previous 1059 task

NickSaglioni:
- Adjusting some text from `GrantAccess` and `GrantApplicationAdmin` pages;
- Changing the 'User is admin already' message from warning to error and adjusting the text;
- Adjusting the `StepContainer` component to have the subtitle prop as optional;
- Fixed UserDomain label bug introduced in the task #1059;
- Updated the router guard to check for user roles before routing (this logic might need to be changed);


frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -75,10 +75,22 @@ const handleSubmit = async () => {
         })
         .catch((error) => {
             if (error.response?.status === 409) {
-                setNotificationMsg(
-                    Severity.warning,
-                    error.response.data.detail
-                );
+                // This is not a good way to check and replace a message error.
+                // Maybe we could implement a detail code in the backend's response,
+                // similar to what we have for 'self_grant_prohibited'
```
MCatherine1994, line 80: Hi Nick, this is a good idea, I added it to the ticket #820, thanks!!

ianliuwk1019: Didn't know that message is from backend?!
I think long time ago we created a ticket to discuss situation like this that how we really like to unify error handling backend and frontend; but we never did have chance on that discussion.

Please add a //TODO with nosonar on your comment so we don't ignore this comment. Thanks.

MCatherine1994: I pushed a fix, think for now 409 is just used for conflict, so did some cleanup, we can keep improve later

frontend/src/enum/SeverityEnum.ts
```diff
@@ -2,4 +2,15 @@ export enum Severity {
...
     error = 'error',
-}
\ No newline at end of file
+}
+
+export enum ErrorCode {
+    conflict = 'conflict',
+    selfGrantProhibited = 'selfGrantProhibited',
+    default = 'default',
+}
```
ianliuwk1019, line 11: Maybe better use Typescript convention for enum: **PascalCase** or **CONSTANT_CASE** for enum's property.

frontend/src/enum/SeverityEnum.ts
```diff
@@ -2,4 +2,15 @@ export enum Severity {
...
+
+export enum ErrorDescription{
+    selfGrantProhibited = 'Granting admin privilege to self is not allowed.',
+    default = 'An error has occured.',
+}
```
ianliuwk1019, line 16: Rarely seen using enum like this `selfGrantProhibited = 'Granting admin privilege to self is not allowed.',`. Often should be constant.

frontend/src/router/routeHandlers.ts
```diff
@@ -29,6 +29,11 @@ import type { RouteLocationNormalized } from 'vue-router';
  * state for it to be handled elsewhere.
  */
 
+const ACCESS_RESTRICTED_ERROR = new FamRouteError(
+    RouteErrorName.ACCESS_RESTRICTED,
+    "This access is restricted"
```
ianliuwk1019, line 34: Maybe just say " Restricted Access".

frontend/src/services/AuthService.ts
```diff
@@ -128,6 +128,13 @@ function storeFamUser(famLoginUser: FamLoginUser | null | undefined) {
     }
 }
 
+const userHasAccessByRole = (role: string): boolean => {
```
ianliuwk1019, line 131: Very minor, can we name it as "hasAccessRole"?

NickSaglioni: Thanks, I'll change it.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -234,10 +211,7 @@ const composeAndPushNotificationMessages = (
     >
         <div class="page-body">
             <form id="grantAccessForm" class="form-container">
-                <StepContainer
-                    title="User information"
-                    :subtitle="`Enter the user information to add a new user to ${selectedApplicationDisplayText}`"
-                >
```
ianliuwk1019, line 240: The removal of the `${selectedApplicationDisplayText}` might make this "GrantAccess" page confusing for the admin as they will not know easily which "application" they are assigning users for.
Might need to discuss with @OlgaLiber2 what's the better text to indicate the "application" was selected for granting.
@MCatherine1994

NickSaglioni: I agree with @ianliuwk1019  on this matter, maybe we need to come up with a better text for the subtitle item, not removing it altogether. My suggestion is to revert it to what it was while we discuss it. @MCatherine1994 @OlgaLiber2

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -120,11 +120,10 @@ const areVerificationsPassed = () => {
 
 const handleSubmit = async () => {
     const successForestClientIdList: string[] = [];
-    const warningForestClientIdList: string[] = [];
 
     // msg override the default error notification message
     const errorNotification = {
-        msg: '',
+        code: 'default',
```
ianliuwk1019, line 126: Since there is a new ErrorCode, is this meant to use "ErrorCode.Default"?

frontend/src/store/Constants.ts
```diff
@@ -1,3 +1,9 @@
...
 export const FAM_APPLICATION_NAME = 'FAM';
-export const FAM_APPLICATION_ID = 1;
\ No newline at end of file
+export const FAM_APPLICATION_ID = 1;
+export const FAM_ADMIN_ROLE = 'FAM_ADMIN';
+
+export const ErrorDescription = {
+    SelfGrantProhibited: 'Granting admin privilege to self is not allowed.',
+    Default: 'An error has occured.',
+}
```
ianliuwk1019, line 9: It's minor, but better have this "ErrorDescription" be in where the "ErrorCode" located.

ianliuwk1019 (approved):

MCatherine1994 (approved):

Issue #1174, opened 2024-02-02
Add Tests Code for Build get_my_FAM_access_API Ticket


ianliuwk1019:
**Describe the task**
This is additional ticket from original #1143 to add tests code at backend.
Tests should cover scenarios for the endpoints intended:
* Test with user that has no role
* Test with user that has only FAM-ADMIN role
* Test with user that has only/multiple app admin roles
* Test with user that has only/multiple delegated admin access (ie; FOM_SUBMITTER)
* Test with user that has multiple access privileges.

**Acceptance Criteria**
- [x] Router code tests added.
- [x] Service code tests added.
- [x] Repository code tests added (if needed).

PR #1170, opened 2024-02-01
feat: #1143 my fam access api

ianliuwk1019:
#1143 
- Add router router_admin_user_access:: get_admin_user_access, ("/admin-user-access").
- Add few schema objects for "FAM Admin User Access" endpoint
- Add "get_user_delegated_admin_grants" method for repository to get grants for the user with
	     SqlAlchemy query.
- Add "get_user_app_admin_grants" method for repository to get grants for the user with SqlAlchemy
		 query.
- Add "get_base_roles_by_app_id" method for repository to get roles from the application for only
	     the base role.
- Add new "fam_admin_user_access_service.py" service and "get_access_grants" method to construct
	     all necessary grants for the logged on admin user.
- Remove "_DEV"/"_TEST"/"_PROD" from application.name (processed from the database value)
- Add client-code generation for this endpoint.
- Configure frontend to be ready to use this endpoint.
- Didn't remove environment from "description" in json return as I feel might be ok and good for frontend (such as"Forest Operations Map (TEST)")
====================================================================
Produced example:
* When logged on user has no grant for any of the admin privileges:
```
{
  "access": []
}
```
* When logged on user has grants for **FAM_ADMIN** privileges:

```
{
  "access": [
    {
      "auth_key": "FAM_ADMIN",
      "grants": [
        {
          "application": {
            "id": 1,
            "name": "FAM",
            "description": "Forests Access Management",
            "env": null
          },
          "roles": null
        },
        {
          "application": {
            "id": 2,
            "name": "FOM",
            "description": "Forest Operations Map (DEV)",
            "env": "DEV"
          },
          "roles": null
        },
        {
          "application": {
            "id": 3,
            "name": "FOM",
            "description": "Forest Operations Map (TEST)",
            "env": "TEST"
          },
          "roles": null
        },
        {
          "application": {
            "id": 4,
            "name": "FOM",
            "description": "Forest Operations Map (PROD)",
            "env": "PROD"
          },
          "roles": null
        },
        {
          "application": {
            "id": 5,
            "name": "SPAR",
            "description": "Seed Planning and Registry Application (DEV)",
            "env": "DEV"
          },
          "roles": null
        },
        {
          "application": {
            "id": 6,
            "name": "SPAR",
            "description": "Seed Planning and Registry Application (TEST)",
            "env": "TEST"
          },
          "roles": null
        },
        {
          "application": {
            "id": 7,
            "name": "SPA",
            "description": "Seed Planning and Registry Application (PROD)",
[...]

PR #1195, opened 2024-02-12
feat: #1174 add tests for get my fam access api

ianliuwk1019:
Tests added for get_admin_user_access.
- Test with user that has no role
- Test with user that has only FAM-ADMIN role
- Test with user that has only/multiple app admin roles
- Test with user that has only/multiple delegated admin access (ie; FOM_SUBMITTER)
- Test with user that has multiple access privileges.


server/admin_management/tests/conftest.py
```diff
@@ -130,6 +140,126 @@ def test_rsa_key_missing():
     return new_key.exportKey("PEM")
 
 
+@pytest.fixture(scope="function")
+def setup_new_user(user_repo: UserRepository, db_pg_session: Session):
```
MCatherine1994, line 144: I wonder if we could add the return type for this method? as well as the setup_new_fom_delegated_admin?

server/admin_management/tests/conftest.py
```diff
@@ -130,6 +140,126 @@ def test_rsa_key_missing():
...
+
+
+@pytest.fixture(scope="function")
+def setup_new_fom_delegated_admin(
+    access_control_privilege_repo: AccessControlPrivilegeRepository,
+    role_service: RoleService
+):
+    """
+    Conveniently setup new "FOM" DELEGATED_ADMIN user for testing.
+    The fixture returns a function to be called with new FOM (evn) delegated
```
MCatherine1994, line 210: just a typo "evn"

server/admin_management/tests/routers/test_router_admin_user_access.py
```diff
@@ -0,0 +1,371 @@
...
+    fom_test_roles_grants = fom_test_grant["roles"]
+    # was setup for ROLE_TYPE_ABSTRACT (FOM_SUBMITTER), so should only have 1
+    assert len(fom_dev_roles_grants) == 1
+    fom_test_role_grants_submitter = fom_test_roles_grants[0]
+    assert fom_test_role_grants_submitter["id"] == TEST_FOM_TEST_SUBMITTER_ROLE_ID
+    assert fom_test_role_grants_submitter["name"] == TEST_APP_ROLE_NAME_FOM_SUBMITTER
+    assert fom_test_role_grants_submitter["type_code"] == RoleType.ROLE_TYPE_ABSTRACT.value
+    # should contains exact 2 forest_clients the new_user can administer
+    assert set(fom_test_role_grants_submitter["forest_clients"]) == \
+        set([TEST_FOREST_CLIENT_NUMBER, TEST_FOREST_CLIENT_NUMBER_TWO])
```
MCatherine1994, line 274: maybe we can use "dga_submitter_forest_clients" here?

server/admin_management/tests/services/test_admin_user_access_service.py
```diff
@@ -0,0 +1,208 @@
+import logging
+
+import pytest
```
MCatherine1994, line 3: this import is not in use

server/admin_management/tests/routers/test_router_admin_user_access.py
```diff
@@ -0,0 +1,371 @@
...
+                             TEST_FOM_TEST_SUBMITTER_ROLE_ID,
+                             TEST_FOREST_CLIENT_NUMBER,
+                             TEST_FOREST_CLIENT_NUMBER_TWO,
+                             TEST_NEW_BCEID_USER, TEST_NEW_IDIR_USER)
+
+LOGGER = logging.getLogger(__name__)
+test_end_point = f"{apiPrefix}/admin-user-accesses"
+
+# Below tests are based on endpoint json sample output at confluence design:
+# https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Delegated+Access+Administration+Design
```
MCatherine1994, line 28: Hi Ian, I wonder if you want to update this confluence page to make the json sample up to date? For example, in the below tests, for app admin, roles will have forest client number None, `assert fom_dev_roles_grants_submitter["forest_clients"] is None`, but it's not shown in the json sample. Think the json example in your last pr is the latest https://github.com/bcgov/nr-forests-access-management/pull/1170
Screen Shot 2024-02-12 at 4 06 48 PM
Screen Shot 2024-02-12 at 4 06 58 PM

server/admin_management/tests/repositories/test_access_control_privilege_repository.py
```diff
@@ -147,3 +146,53 @@ def test_get_acp_by_application_id(
         TEST_NOT_EXIST_APPLICATION_ID
     )
     assert len(found_acp) == 0
+
+
+def test_get_user_delegated_admin_grants(
+        access_control_privilege_repo: AccessControlPrivilegeRepository,
+        user_repo: UserRepository
```
MCatherine1994, line 153: maybe we format this file?

server/admin_management/tests/routers/test_router_admin_user_access.py
```diff
@@ -0,0 +1,371 @@
...
+    assert response.status_code == HTTPStatus.OK
+    result = response.json().get("access")
+    assert len(result) == 1
+    access = result[0]
+    assert access.get("auth_key") == AdminRoleAuthGroup.FAM_ADMIN  # FAM_ADMIN grants
+    grants = access.get("grants")
+    fam_app_grant = list(filter(lambda x: x["application"]["id"] == TEST_APPLICATION_ID_FAM, grants))[0]
+    assert fam_app_grant["application"] is not None
+    assert fam_app_grant["application"]["id"] == TEST_APPLICATION_ID_FAM
+    assert fam_app_grant["application"]["name"] == TEST_APPLICATION_NAME_FAM  # verify grante app is "FAM", not "FAM_DEV/FAM_TEST"
```
MCatherine1994, line 91: maybe it's a typo 'grante'

server/admin_management/tests/routers/test_router_admin_user_access.py
ianliuwk1019: Good suggestion. I updated the confluence.

However, on the side note, it actually has a layer of translation between python (FastAPI) and Swagger before outputting the swagger return json we see on api's doc (on swagger). I noticed using Python's "Optional" and "Union" types can have difference at the swagger's translation (for example, difference between: None becomes "null" or None becomes => the key/alue being removed). If we do need to make it: None becomes => the key/alue being removed, we can have another ticket to adjust (but not clear what to adjust, the python type, or is there existing swagger conversion configuration for this?).

MCatherine1994: I would prefer to make it optional, so if "None" then remove key/value pair

MCatherine1994 (approved):

server/admin_management/tests/routers/test_router_admin_user_access.py
ianliuwk1019: Thanks Catherine.
I also prefer "Optional" than "Union" in this case.
However, it is the "Optional" that Swagger is producing the "null", not removing the key/value pair.
This is the issue between FastAPI(python) and Swagger, not really in our code.
I will create a "nice-to-have tech debt" ticket so we can research/investigate if there is a way to tweak FastAPI for this at current or higher version.

Issue #1144, opened 2024-01-22
Call API after login and cache results in front-end


OlgaLiber2:
**Describe the task**
Call API after login and cache results in front-end

**Acceptance Criteria**
- [x] confirm front-end gets results
- [x] update users storage for access info
- [ ] 

**Additional context**
Story: The UI should use the new get_my_FAM_access API for determining what the user has access to do as a FAM admin or app admin.

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/a6db2df6-594b-4513-bde1-2c98d5b1e394)

PR #1200, opened 2024-02-14
feat: #1144 call api after login and cache results

J0taFerreira:
- added access atribute from the new adminUserAccesses api to FamLoginUser.

more context on the new api: https://github.com/bcgov/nr-forests-access-management/pull/1170


frontend/src/services/AuthService.ts
```diff
@@ -14,6 +15,7 @@ export interface FamLoginUser {
     idpProvider?: string;
     roles?: string[];
     authToken?: CognitoUserSession;
+    acesses?: FamAuthGrantDto[];
```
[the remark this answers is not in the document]
J0taFerreira: Hello, it is a typo. Sorry!

frontend/src/services/AuthService.ts
```diff
@@ -119,8 +121,11 @@ const removeFamUser = () => {
     localStorage.removeItem(CURRENT_SELECTED_APPLICATION_KEY);
 }
 
-const storeFamUser = (famLoginUser: FamLoginUser | null | undefined) => {
+async function storeFamUser(famLoginUser: FamLoginUser | null | undefined) {
```
MCatherine1994, line 124: Hi Jota, we refactor all function to write in the format of `const storeFamUser = () => {}`, for async, you can use const storeFamUser = async () => {}`, please keep in mind we have to keep consistency, thanks!!

frontend/src/services/AuthService.ts
```diff
@@ -119,8 +121,11 @@ const removeFamUser = () => {
...
     state.value.famLoginUser = famLoginUser;
+    state.value.famLoginUser!.accesses = (
+        await AdminMgmtApiService.adminUserAccessesApi.adminUserAccessPrivilege()
+    ).data.access;
```
MCatherine1994, line 128: also I wonder if you tested the case when this api returns error or it's down, what you will get here, you might need to consider the error handling.

frontend/src/services/AuthService.ts
J0taFerreira: Thank you, I already changed it.

frontend/src/services/AuthService.ts
J0taFerreira: If there is an error should it be a toast or a message?

frontend/src/services/AuthService.ts
```diff
@@ -111,29 +118,44 @@ const parseToken = (authToken: CognitoUserSession): FamLoginUser => {
...
+                        returnResult.data.access;
+                }
+            )
+            .catch((error: AxiosError) => {
+                setRouteToastError(error);
+            });
+        localStorage.setItem(
+            FAM_LOGIN_USER,
+            JSON.stringify(state.value.famLoginUser)
+        );
```
ianliuwk1019, line 147: I think, it is better off moving this retrieving user access logic within `storeFamUser` out to `AuthCallbackHandler.vue` or `AuthService.handlePostLogin()` (so the `storeFamUser` can only just be responsible for setting the cache). Maybe better in `handlePostLogin` function, since it really is meant for handling tasks, such as this, after user finally gets redirected back to app after Cognito.
For example, in `handlePostLogin` function, can do these steps:
```
 try {
       // DO await for : previous AWS stuff.
       // DO await for : previous refresh token stuff (this is when the app have the token ready)
       // DO await for: retrieving adminUserPrivilege() from backend.
       // DO: since the user privilege retrieved successfully at this point, we can set this into the famLoginUser. 
 }
 catch(error: any) {
        console.log('Not signed in');
        console.error('Error:', error);
        // DO: any other error handling you think it needs.
        return logout();
 }
```

This `handlePostLogin` was created at the early stage of this Vue project, the code looks old that it uses "Promise" style on return and error handling. Since it is an `async` function, it can ues "async-await` style and no need for `.then` (unless your feel having good reason to do so).

ianliuwk1019 (changes requested):

frontend/src/services/AuthService.ts
MCatherine1994: Hi Ian, yeah I'm fine to move it up, but then after call the api, set the access in state famLoginUser, we need to set local storage again. That why initially I just put it inside the storeFamUser so it's set at one place. But I'm fine to move it up cause I feel maybe the logic will be more clear.

frontend/src/router/routeHandlers.ts
```diff
@@ -148,5 +148,6 @@ export const beforeEachRouteHandler = async (
     if (AuthService.state.value.famLoginUser) {
         // condition needed to prevent infinite redirect
         await AuthService.methods.refreshToken();
+        await AuthService.methods.getUserAccess();
```
MCatherine1994, line 151: Add here, so in the case if a user already logged in,  and he has been granted a new admin role, then when refresh the token, and call get access again, the user should be able to manage the new app

ianliuwk1019: Thanks Catherine for this addition!
It is being called twice after login, but I think it is ok. When deployed, we can see if there is impact.
image

ianliuwk1019 (approved):

frontend/src/router/routeHandlers.ts
MCatherine1994: yeah I noticed that as well, it's been called when handle post login, that's when we call both refresh toke and get user access, and when enter each router, it will call refresh toke and get user access again

MCatherine1994: yeah let's see if there is any impact in dev, thanks so much!!

PR #1203, opened 2024-02-15
fix:/#1066 admin segregation cleanup

ianliuwk1019:
WIP...

### DIFF
--- a/server/flyway/sql/V38__remove_admins_from_user_role_xref_and_cleanup_fam_roles.sql
+++ b/server/flyway/sql/V38__remove_admins_from_user_role_xref_and_cleanup_fam_roles.sql
@@ -1,0 +1,43 @@
+-- After migrating the admins to fam admin management table (V34) and with code base migration completed,
+-- we can remove fam admin association for the users from fam_user_role_xref table and remove FAM roles
+-- from fam_role table.
+
+-- Just to be safe before deletion records, migrating admin user_role_xref again,
+-- if conflict do nothing.
+WITH selected_admin_xrefs AS (
+    SELECT user_role_xref.user_id, application.application_id, CURRENT_USER, CURRENT_DATE
+    FROM app_fam.fam_role role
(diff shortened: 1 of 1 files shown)


server/flyway/sql/V38__remove_admins_from_user_role_xref_and_cleanup_fam_roles.sql
```diff
@@ -0,0 +1,43 @@
+-- After migrating the admins to fam admin management table (V34) and with code base migration completed,
+-- we can remove fam admin association for the users from fam_user_role_xref table and remove FAM roles
+-- from fam_role table.
+
+-- Just to be safe before deletion of records, migrate remaining admin user_role_xref again (if any),
+-- if conflict on insert then do nothing.
+WITH selected_admin_xrefs AS (
+    SELECT user_role_xref.user_id, application.application_id, CURRENT_USER, CURRENT_DATE
```
MCatherine1994, line 8: Hi Ian, I wonder for this sql query, if the `application.application_id` will always be 1(FAM)? However if we want to grab app admin, the application id shouldn't always be 1. Maybe you can use the same query as in v34. So for example for FOM_DEV_ACCESS_ADMIN, role.application_id = 1 (this is a FAM role), but this is for FOM DEV ADMIN, when we insert into the fam_application_admin table, application_id should be 2(FOM_DEV) != role.application_id

ianliuwk1019: Hmm... make sense, I think I missed noticing that. Let me think a bit...

ianliuwk1019: I also tried not to use application_id = 1. Initially our FAM isn't defined to be in the first sequence in flyway, it was only due to the first in code to be executed.

MCatherine1994 (approved):

Issue #1145, opened 2024-01-22
Use API in front-end to get list of applications


OlgaLiber2:
**Describe the task**
Use API cached results in front-end to get list of applications

**Acceptance Criteria**
- [x] Stop using the get-applications API call
- [x] Used cached results to render application list
- [x] Tested with user that has only FAM-ADMIN role
- [x] Tested with user that has no role
- [x] Tested with user that has multiple app admin roles
- [x] Tested with user that has delegated admin access (ie; FOM_SUBMITTER)

**Additional context**
Refers to the endpoint done in #1144
Story: The UI should use the new get_my_FAM_access API for determining what the user has access to do as a FAM admin or app admin.

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/3a3e2369-6f8b-40f6-be55-f2301eca53b0)

**May need** to do 'manual' testing and separate ticket for component testing if it gets too complicated

**Test note:**
In ticket #1144, we cached the user access in the local storage "famLoginUser" (this could change depends on how #1144 is implemented). When we write component tests in this ticket, we can set the local storage "famLoginUser" with the access we want (based on different test scenario), and mount the "Manage Application" component, and verify if the dropdown returns the expected applications. Like in the screenshot below, the user has access in three applications, no matter the detail, the dropdown should return three apps. 
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/6a7c7a8d-108e-43fd-bf46-0ea0b8ff9b04)

ianliuwk1019:
Since I am replacing the way getting the applications, I also added this to be convenient/and obvious to user profile to know the user's level of admins like this. It is just a quick addition. If requires, we can have improvement ticket to make it better.
Would it be ok?
@MCatherine1994 @OlgaLiber2 
![Image](https://github.com/bcgov/nr-forests-access-management/assets/81595625/55a0d153-dc78-4cd0-94c0-c020870e77ab)

ianliuwk1019:
For this ticket, I think we won't do the last check `Tested with user that has delegated admin access (ie; FOM_SUBMITTER)` since current app does not have "Delegated Admin" screens built yet.
@MCatherine1994 @OlgaLiber2 

[...]

PR #1209, opened 2024-02-21
feat: #1145 use cached access for applications list

ianliuwk1019:


frontend/src/store/FamLoginUserState.ts
```diff
@@ -0,0 +1,168 @@
...
+        )
+    }
+};
+
+/**
+ * @param adminRole three admin levels: 'FAM_ADMIN', 'APP_ADMIN', 'DELEGATED_ADMIN'
+ * @returns array of applications the login user has been granted for
+ *          administering on based on adimn level.
+ */
+const getApplicationsAdministeredByAdminRole = (
```
MCatherine1994, line 56: Hi Ian, just personal thoughts... I wonder if we can just call it getAppsByAdminRole.... the applications administered is too long I feel....

ianliuwk1019: Sure, actually I like the short name initially but was afraid it is not so clear to others.
I will use the shorter name. 
I created all methods that are related and close to the state file to provide functionalities and would be easier to know where they are first before even touch components but either places are fine to me. I will move to the component.

MCatherine1994: oh sorry I just saw you use it in getApplicationsUserAdministers method as well, it confused me a bit.... I might prefer to separate it, so this call will just be used by fam admin

ianliuwk1019: Yes, those methods are useful as they are very closed related to the state. That's why it is much clear to me to be in the container of the same file and just provide them for whoever needs them (even just for one component, much like util may just be used for one function).
If we separate them out into components then component will be packed with complicated code that they are not suppose to be, and basic common logic will also need to be repeated in both places.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * If the way our forntend works is based on admin level to display
+ * management table instead of selecting "applicatoin" first, it won't need
+ * this particular function.
+ */
+const getApplicationsUserAdministers = () => {
+    const userAdminRoleGroups = getUserAdminRoleGroups();
+    let famApp;
+    const applicationList: Array<FamApplicationDto> = [];
+    if (userAdminRoleGroups) {
```
MCatherine1994, line 85: so maybe we can just go through the access array instead of go through the userAdminRoleGroups and call `getApplicationsAdministeredByAdminRole`, I feel it's a little bit hard to follow

frontend/src/store/FamLoginUserState.ts
```diff
@@ -0,0 +1,168 @@
...
+    // clean up local storage for selected application
+    localStorage.removeItem(CURRENT_SELECTED_APPLICATION_KEY);
+};
+
+/**
+ * To refresh User granted privileges when neccssary.
+ * FamLoginUser contains a property "accesses" that sometimes needs to be
+ * refreshed and stored into localStorage at FamLoginUser object.
+ */
+const refreshCachedUserAccess = async () => {
```
MCatherine1994, line 143: maybe we shouldn't call it refreshCached anymore, cause we only call it when we login, and there won't be any cached access

frontend/src/store/FamLoginUserState.ts
MCatherine1994: I'm fine to keep it here, but maybe restrict it to call used by fam admin, and in the getApplicationsUserAdministers method, just go through the whole access list, and check `access.grants.forEach`, then you don't need to reuse this method, and feel it's more clear

MCatherine1994:
Thanks Ian! Looks great! Just some small comment for simplify and clarification. Also maybe Nick @NickSaglioni if you want to check on this pr, Ian is getting the application options inside the component now instead of preparing it in the router handler, just regarding of your question this morning. Thanks so much!

frontend/src/store/FamLoginUserState.ts
ianliuwk1019: Good suggestion, although I was reusing other function for convenience but the extra mapping is unnecessary. the logic refactored.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -0,0 +1,166 @@
...
+            const accessGrants = access.grants;
+            if (access.auth_key == AdminRoleAuthGroup.FamAdmin) {
+                famApp = accessGrants.filter(
+                    (grant) => grant.application.name == FAM_APPLICATION_NAME
+                )[0].application;
+            } else {
+                accessGrants.forEach((grant) => {
+                    const app = grant.application
+                    let isNewItem = true
+                    for (let item of applicationList) {
```
MCatherine1994, line 92: Hi Ian, just wondering maybe we check if length of applicationList.filter((addedApp) => addedApp.id == app.id) > 0?  so we have less code? it doesn't impact functionality, I'll approve it, thanks!!

MCatherine1994 (approved):

frontend/src/store/FamLoginUserState.ts
ianliuwk1019: Hi Catherine, I think it is fine, I don't always use short-form, sometimes I favor for-loop for clearness.
Thanks Catherine! I will merge so others can resolve their conflict.

Issue #1159, opened 2024-01-26
Frontend display of "add delegated admin" form


OlgaLiber2:
**Describe the task**
Create a "add delegated admin" form (on the front end) 

**Acceptance Criteria**
- [x] Grant delegated admin access page exists (frontend)
- [x] Get the role list from the cached user accesses
- [x] Validation of required fields before submitting is working
- [x] New path to delegated admin page route included
- [x] Make sure only app admin can access to this route

**Additional context**
User story: As an app admin I want to grant a user access to be a delegated administrator for a particular application, role(s), and forest client(s).

Can reuse user domain select, username input, forest client input components

Depends on #1160

Links to mockups 

- add a delegated admin page: https://apps.nrs.gov.bc.ca/int/confluence/download/attachments/163414102/FAM-delegated-admin-form.png?version=1&modificationDate=1705432155000&api=v2 
- validation popup modal: https://apps.nrs.gov.bc.ca/int/confluence/download/attachments/163414102/FAM-delegated-admin-form-popup.png?version=1&modificationDate=1705432242000&api=v2
- display option on left nav bar: https://apps.nrs.gov.bc.ca/int/confluence/download/attachments/163414102/FAM-delegated-admin-button.png?version=1&modificationDate=1705432026000&api=v2

gormless87:
test close

Issue #1147, opened 2024-01-22
Use API to get list of roles for app admins


OlgaLiber2:
**Describe the task**
A clear and concise description of what the task is.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
Story: The UI should use the new get_my_FAM_access API for determining what the user has access to do as a FAM admin or app admin.

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/0c3230da-90ad-415b-a0d4-eb296852849f)

MCatherine1994:
Hi @J0taFerreira, just fyi, we don't need to write tests in this ticket, but you want to test it manually, thanks!!

PR #1212, opened 2024-02-26
feat: #1159 add delegated admin form

NickSaglioni:
- Created `GrantDelegatedAdmin` and `GrantDelegatedAdminView` components;
- Getting the app role list from the cached user accesses;
- Validating required fields before submit button is enabled;
- Created new new path to delegated admin page route;
- Only FAMMake sure only app admin can access to this route

### DIFF
--- a/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
+++ b/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
@@ -200,7 +200,6 @@ const composeAndPushNotificationMessages = (
         subtitle="All fields are mandatory"
     />
 
-    {{ delegatedRoleOptions }}
     <VeeForm
         ref="form"
         v-slot="{ errors, meta }"
@@ -236,8 +235,8 @@ const composeAndPushNotificationMessages = (
                     >
                         <Dropdown
                             :options="delegatedRoleOptions"
-                            optionLabel="role_name"
-                            optionValue="role_id"
+                            optionLabel="name"
+                            optionValue="id"
                             :modelValue="field.value"
                             placeholder="Choose an option"
                             class="w-100 custom-height"
--- a/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
+++ b/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
@@ -107,7 +107,7 @@ const handleSubmit = async () => {
         as="div"
     >
         <div class="page-body">
-            <form id="grantAdminForm" class="form-container">
+            <form id="grantDelegatedForm" class="form-container">
                 <StepContainer title="User information">
                     <UserDomainSelect
                         :domain="formData.domain"
@@ -154,7 +154,7 @@ const handleSubmit = async () => {
                 <div class="button-stack">
                     <Button
                         type="button"
-                        id="grantAccessCancel"
+                        id="grantDelegatedCancel"
                         class="w100"
                         severity="secondary"
                         label="Cancel"
@@ -164,7 +164,7 @@ const handleSubmit = async () => {
                     >
                     <Button
                         type="button"
-                        id="grantAccessSubmit"
+                        id="grantDelegatedSubmit"
                         class="w100"
                         label="Submit Application"
                         :disabled="
--- a/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
+++ b/frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
@@ -1,0 +1,303 @@
+<script setup lang="ts">
+import { ref, type PropType } from 'vue';
+import router from '@/router';
+import { number, object, string } from 'yup';
+import { ErrorMessage, Field, Form as VeeForm } from 'vee-validate';
+
+import {
+    type FamApplicationRole,
+    UserType,
+    type FamUserRoleAssignmentCreate,
+} from 'fam-app-acsctl-api';
+
+import Dropdown from 'primevue/dropdown';
+import Button from '@/components/common/Button.vue';
+import { IconSize } from '@/enum/IconEnum';
+import { Severity, ErrorCode } from '@/enum/SeverityEnum';
+import { AppActlApiService } from '@/services/ApiServiceFactory';
+
+import { FOREST_CLIENT_INPUT_MAX_LENGTH } from '@/store/Constants';
+import { isLoading } from '@/store/LoadingState';
+import { setGrantAccessNotificationMsg } from '@/store/NotificationState';
+
+const props = defineProps({
+    applicationRoleOptions: {
+        // options fetched from route.
+        type: Array as PropType<FamApplicationRole[]>,
+        default: [],
+    },
+});
+
+const defaultFormData = {
+    domain: UserType.I,
+    userId: '',
+    verifiedForestClients: [],
+    roleId: null as number | null,
+};
+const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
+const formValidationSchema = object({
+    userId: string()
+        .required('User ID is required')
+        .min(2, 'User ID must be at least 2 characters')
+        .nullable(),
+    roleId: number().required('Please select a value'),
+    forestClientNumbers: string()
+        .when('roleId', {
+            is: (_role_id: number) => isAbstractRoleSelected(),
+            then: () =>
+                string()
+                    .nullable()
+                    .transform((curr, orig) => (orig === '' ? null : curr)) // Accept either null or value
+                    .matches(/^[0-9,\b]+$/, 'Please enter a digit or comma')
+                    .matches(
+                        /^\d{8}(,?\d{8})*$/,
+                        'Please enter a Forest Client ID with 8 digits long'
+                    ),
+        })
+        .nullable(),
+});
+
+/* ------------------ User information method ------------------------- */
+const userDomainChange = (selectedDomain: string) => {
+    formData.value.domain = selectedDomain;
+    formData.value.userId = '';
+};
+
+const userIdChange = (userId: string) => {
+    formData.value.userId = userId;
+};
+
+const verifyUserIdPassed = ref(false);
+const setVerifyUserIdPassed = (verifiedResult: boolean) => {
+    verifyUserIdPassed.value = verifiedResult;
+};
+
+/* ------------------- Role selection method -------------------------- */
+const getSelectedRole = (): FamApplicationRole | undefined => {
+    // return props.applicationRoleOptions?.find(
+    //     (item) => item.role_id === formData.value.roleId
+    // );
+    return undefined;
+};
+
+const isAbstractRoleSelected = () => {
+    return getSelectedRole()?.role_type_code == 'A';
+};
+
+/* ----------------- Forest client number method ----------------------- */
+const setVerifiedForestClients = (verifiedForestClientNumber: string) => {
+    formData.value.verifiedForestClients.push(verifiedForestClientNumber);
+};
+
+const removeVerifiedForestClients = (index: number) => {
+    formData.value.verifiedForestClients.splice(index, 1);
+};
+
+const resetVerifiedForestClients = () => {
+    formData.value.verifiedForestClients = [];
+};
+
+/* ---------------------- Form method ---------------------------------- */
+const cancelForm = () => {
+    formData.value = defaultFormData;
+    router.push('/dashboard');
+};
+
+/*
+Two verifications are cunnretly in place: userId and forestClientNumber.
+*/
+const areVerificationsPassed = () => {
+    return (
+        // userId verification
+        verifyUserIdPassed.value &&
+        // forestClientNumber verification
+        (!isAbstractRoleSelected() ||
+            (isAbstractRoleSelected() &&
+                formData.value.verifiedForestClients.length > 0))
+    );
+};
+
+const handleSubmit = async () => {
+    const successForestClientIdList: string[] = [];
+
+    // msg override the default error notification message
+    const errorNotification = {
+        code: ErrorCode.Default,
+        errorForestClientIdList: [] as string[],
+    };
+    do {
+        const forestClientNumber = formData.value.verifiedForestClients.pop();
+        const data = toRequestPayload(formData.value, forestClientNumber);
+        await AppActlApiService.userRoleAssignmentApi
+            .createUserRoleAssignment(data)
+            .then(() => {
+                successForestClientIdList.push(forestClientNumber || '');
+            })
+            .catch((error) => {
+                if (error.response?.status === 409) {
+                    errorNotification.code = ErrorCode.Conflict;
+                } else if (
+                    error.response.data.detail.code === 'self_grant_prohibited'
+                ) {
+                    errorNotification.code = ErrorCode.SelfGrantProhibited;
+                }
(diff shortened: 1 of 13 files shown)


frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -0,0 +1,222 @@
...
+
+import Dropdown from 'primevue/dropdown';
+import Button from '@/components/common/Button.vue';
+import { IconSize } from '@/enum/IconEnum';
+
+import { isLoading } from '@/store/LoadingState';
+import type { FamRoleDto } from 'fam-admin-mgmt-api/model';
+
+const props = defineProps({
+    delegatedRoleOptions: {
```
MCatherine1994, line 17: Hi Nick, so here we don't need to set this value from the router anymore, if you check the `GrantApplicationAdmin.vue` file,  we now define and get its value inside the component, something like below:
Screen Shot 2024-02-26 at 2 14 51 PM

MCatherine1994: And I think we only assign abstract role with forest client numbers for now, ignore the concrete role, but let's ask Olga tomorrow to confirm.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -0,0 +1,222 @@
...
+                            removeVerifiedForestClients
+                        "
+                        @resetVerifiedForestClients="resetVerifiedForestClients"
+                    />
+                </StepContainer>
+
+                <div class="button-stack">
+                    <Button
+                        type="button"
+                        id="grantAccessCancel"
```
MCatherine1994, line 197: also the id for two buttons, need to change the button id in the `GrantApplicationAdmin.vue` component as well, those are also copied

frontend/src/router/index.ts
```diff
@@ -100,6 +101,25 @@ const routes = [
         component: GrantApplicationAdminView,
         beforeEnter: beforeEnterHandlers[routeItems.grantAppAdmin.name],
     },
+    {
+        path: routeItems.grantDelegatedAdmin.path,
+        name: routeItems.grantDelegatedAdmin.name,
+        meta: {
+            requiresAuth: true,
+            requiresAppSelected: true,
+            requiredPrivileges: [FAM_ADMIN_ROLE],
```
MCatherine1994, line 110: this will be app admin, only app admin can grant delegated admin

frontend/src/router/routeHandlers.ts
```diff
@@ -91,10 +91,30 @@ const beforeEnterGrantApplicationAdminRoute = async (
     return true;
 }
 
+const beforeEnterGrantDelegationAdminRoute = async (
+    to: RouteLocationNormalized
+) => {
+
+    if (!LoginUserState.hasAccess(APP_ADMIN_ROLE)) {
```
MCatherine1994, line 98: we don't need to check the user role here, cause if we set `requiredPrivileges: [APP_ADMIN_ROLE]` in the file above, the handler `beforeEachRouteHandler` already checks for permission. 

But we do need to check if the selected application is FAM or not, same as the `beforeEnterGrantUserPermissionRoute`, when FAM is selected, the user shouldn't access to grant delegated admin form

frontend/src/store/FamLoginUserState.ts
```diff
@@ -150,6 +162,14 @@ const cacheUserAccess = async () => {
     }
 };
 
+const delegatedCachedData = (application_id: number): FamRoleDto[] => {
```
MCatherine1994, line 165: let's rename this to `getCachedAppRoles`

frontend/src/store/FamLoginUserState.ts
```diff
@@ -150,6 +162,14 @@ const cacheUserAccess = async () => {
...
+    const delegatedCachedData = state.value.famLoginUser?.accesses?.find(key => key.auth_key === DELEGATED_ADMIN_ROLE)?.grants.find((item) => {
```
MCatherine1994, line 166: `state.value.famLoginUser?.accesses` is `getUserAccess()`

frontend/src/store/FamLoginUserState.ts
```diff
@@ -113,6 +113,18 @@ const hasAccessRole = (role: string): boolean => {
     return false;
 };
 
+const hasAccess = (role: any): boolean => {
```
MCatherine1994, line 116: So if we set `requiredPrivileges: [APP_ADMIN_ROLE]` above, and no need need method anymore

MCatherine1994:
The new `GrantDelegatedAdmin.vue` component has many duplication as the `GrantAccess.vue` component, but I feel it's fine for now. The handle submit will be sightly, and some titles are different. Later on when we work on delegated admin assign user access, there might have more difference. We can consider the further refactoring later.  

Let's see if we can make the `formValidationSchema` reusable, also the role selection dropdown.

frontend/src/router/routeHandlers.ts
NickSaglioni: Thanks Catherine. I'll correct that.

frontend/src/store/FamLoginUserState.ts
NickSaglioni: Yeah, it isn't needed in this case. Nice catch.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -0,0 +1,222 @@
...
+    />
+
+    <VeeForm
+        ref="form"
+        v-slot="{ errors, meta }"
+        :validation-schema="formValidationSchema"
+        as="div"
+    >
+        <div class="page-body">
+            <form id="grantAdminForm" class="form-container">
```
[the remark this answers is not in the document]
NickSaglioni: Changed to `grantDelegatedForm`

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
NickSaglioni: Added unique IDs to buttons

frontend/src/store/FamLoginUserState.ts
NickSaglioni: Thanks. Changed as suggested.

frontend/src/store/FamLoginUserState.ts
NickSaglioni: Thanks, I overlooked that method...

frontend/src/router/index.ts
NickSaglioni: Thanks, I have changed it.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
NickSaglioni: Hi Catherine. We can leave it unfiltered for now, right? @MCatherine1994

MCatherine1994: yeah, no need to filter it. But we don't need to set this value from the router, check the `GrantApplicationAdmin.vue` file, and we now define and get the role options value inside the component

frontend/src/router/routeHandlers.ts
```diff
@@ -91,10 +91,30 @@ const beforeEnterGrantApplicationAdminRoute = async (
...
 }
 
+const beforeEnterGrantDelegationAdminRoute = async (
+    to: RouteLocationNormalized
+) => {
+
+    if (selectedApplicationId.value === FAM_APPLICATION_ID) {
+        emitRouteToastError(ACCESS_RESTRICTED_ERROR);
+        return { path: routeItems.dashboard.path };
+    }
```
MCatherine1994, line 101: Hi Nick, I think I missed one situation, we have to verify the user is the admin of the selected application. For example, I'm the delegated admin of FOM_DEV, app admin of FOM_TEST, when I selected FOM_DEV, even though I still have the app admin role, but I'm not the app admin of FOM_DEV, then I shouldn't be able to access to this page.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -150,6 +150,14 @@ const cacheUserAccess = async () => {
     }
 };
 
+const getCachedAppRoles = (application_id: number): FamRoleDto[] => {
+    const delegatedCachedData = getUserAccess()!.find(key => key.auth_key === DELEGATED_ADMIN_ROLE)?.grants.find((item) => {
```
MCatherine1994, line 154: this should be APP_ADMIN role, so this is for APP_ADMIN to add a delegated admin

frontend/src/components/grantaccess/form/RoleSelect.vue
```diff
@@ -0,0 +1,62 @@
...
+        <label>{{ props.label }}</label>
+        <Field
+            :name="props.fieldId"
+            aria-label="Role Select"
+            v-slot="{ field, handleChange, errorMessage }"
+            v-model="computedRoleId"
+        >
+            <Dropdown
+                :options="roleOptions"
+                optionLabel="role_name"
```
MCatherine1994, line 41: also we need to adjust this, so the role options in the "GrantAccess" component is still getting from the api, so the data is like below, use `role_name`, and `role_id`. But we should now switch to get role options from the access attribute, the role option data is different, which is using `id` and `name`. So let's make it dynamic for now, and when we work on ticket #1147, we'll switch to use `id` and `name` for both cases.
Screen Shot 2024-02-27 at 2 29 12 PM

Screen Shot 2024-02-27 at 2 27 53 PM

frontend/src/components/grantaccess/form/RoleSelect.vue
```diff
@@ -0,0 +1,62 @@
...
+import { computed } from 'vue';
+import { ErrorMessage, Field } from 'vee-validate';
+import Dropdown from 'primevue/dropdown';
+import type { FamApplicationRole } from 'fam-app-acsctl-api';
+import type { FamRoleDto } from 'fam-admin-mgmt-api/model';
+
+const props = defineProps({
+    roleId: { type: Number, default: 0 },
+    roleOptions: {
+        type: [Array<FamApplicationRole[]>, Array<FamRoleDto[]>],
```
MCatherine1994, line 11: oh sorry also here, it should be` [Array<FamApplicationRole>, Array<FamRoleDto>]`

frontend/src/components/grantaccess/form/RoleSelect.vue
NickSaglioni: I am updating the `RoleSelect` component to work dynamically with both `FamApplicationRole` and `FamRoleDto` types. After completing task #1147, this dynamic behavior not going to be needed anymore and should be removed.

frontend/src/components/grantaccess/form/RoleSelect.vue
NickSaglioni: Thanks, I am going to update it.

frontend/src/store/FamLoginUserState.ts
NickSaglioni: Thanks, I'll correct this as suggested.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
NickSaglioni: Thanks, I'll update it as suggested.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -132,7 +132,7 @@ const deleteAppAdmin = async (admin: FamAppAdminGetResponse) => {
 
 <template>
     <ManagePermissionsTitle :isApplicationSelected="isApplicationSelected" />
-
+    {{ LoginUserState.state.value.famLoginUser?.accesses }}
```
MCatherine1994, line 135: Hi Nick, just wondering maybe your forgot to remove this line?

MCatherine1994: I removed it.

frontend/src/router/routeHandlers.ts
```diff
@@ -137,10 +150,18 @@ export const beforeEachRouteHandler = async (
...
         for (let role of to.meta.requiredPrivileges as Array<string>) {
-            if (!LoginUserState.hasAccessRole(role)) {
+            if (!LoginUserState.hasAccess(role)) {
+                emitRouteToastError(ACCESS_RESTRICTED_ERROR);
+                return { path: routeItems.dashboard.path };
+            }
+            // if require APP_ADMIN role, need to be the admin of the selected application
+            if (
+                role == AdminRoleAuthGroup.AppAdmin &&
+                !LoginUserState.isAdminOfSelectedApplication()
```
MCatherine1994, line 163: Hi Nick, I pushed a change regarding the check for "APP_ADMIN". If the router requires "APP_ADMIN", it actually means this router requires user to be the admin of the selected application. So we do a check here `beforeEachRouteHandler`, then no need to add the check in each individual router handler.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -0,0 +1,185 @@
...
+    console.log('Data to send to backend', formData.value);
+
+    router.push('/dashboard');
+};
+</script>
+
+<template>
+    <PageTitle
+        title="Add a delegated admin"
+        :subtitle="`Adding a delegated admin to ${selectedApplicationDisplayText}. All fields are mandatory`"
```
MCatherine1994, line 103: Adjusted this subtitle a little bit, added the name of the selected application, so it's clear to user.

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
```diff
@@ -65,7 +66,7 @@ const delegatedAdminSearchChange = (newvalue: string) => {
         <div class="custom-data-table">
             <DataTableHeader
                 btnLabel="Create delegated admin"
-                :btnRoute="'#'"
+                :btnRoute="routeItems.grantDelegatedAdmin.path"
```
MCatherine1994, line 69: updated the route

MCatherine1994:
Hi Nick! Looks good now, I pushed a commit for some small things, please take a look and verify again. Thanks!!

MCatherine1994 (approved):

frontend/src/components/managePermissions/ManagePermissions.vue
NickSaglioni: Sorry about that. This shouldn't be committed indeed, I think after merging with main, git added it anyway unnoticed. Sorry about the trouble and thanks for removing it.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
NickSaglioni: Thanks, that makes sense.

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
NickSaglioni: Thanks!

frontend/src/router/routeHandlers.ts
NickSaglioni: Sounds good. Thank you so much!

Issue #1163, opened 2024-01-30
Delete delegated admin function (frontend)


OlgaLiber2:
**Describe the task**
Delete icon click results in appropriate action

**Acceptance Criteria**
- [x] pop-up confirmation 
- [x] API is called to delete a record in fam access control privilege table (See #1090 )
- [x] refresh table (API call)
- [x] return notification of success/fail

**Additional context**
- Add any other context about the task here.
- Or here

PR #1219, opened 2024-03-01
feat: #1163 delete delegated admin function

J0taFerreira:
- show notification when deletion is successful and when fails
- refresh table after deletion
- pop-up to confirm action


frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -128,6 +129,26 @@ const deleteAppAdmin = async (admin: FamAppAdminGetResponse) => {
...
+    delegatedAdminAssignment: FamAccessControlPrivilegeGetResponse
+) => {
+    try {
+        delegatedAdmins.value = await deleteAndRefreshDelegatedAdmin(
+            delegatedAdminAssignment.access_control_privilege_id
+        );
+
+        setNotificationMsg(
+            Severity.Success,
+            `You removed ${delegatedAdminAssignment.role.role_name} access to ${delegatedAdminAssignment.user.user_name}`
```
MCatherine1994, line 143: I wonder if we should say "access" or "privilege"... this is like removing the privilege to manage that role

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
```diff
@@ -50,18 +58,45 @@ const delegatedAdminFilters = ref({
...
+    confirmDeleteData.adminName = delegatedAdmin.user.user_name;
+    confirmDeleteData.role = delegatedAdmin.role.parent_role
+        ? delegatedAdmin.role.parent_role.role_name
+        : delegatedAdmin.role.role_name;
+    confirm.require({
+        group: 'deleteDelegatedAdmin',
+        header: 'Remove Access',
+        rejectLabel: 'Cancel',
+        acceptLabel: 'Remove',
+        acceptClass: 'p-button-danger',
```
MCatherine1994, line 83: we might don't need this class

frontend/src/services/fetchData.ts
```diff
@@ -125,3 +126,20 @@ export const fetchDelegatedAdmins = async (
...
+/**
+ * This will call api to delete delegatedAdminPrivilege record from backend and fetch again
+ * to refresh the state.
+ * @param accessPrivilegegId id to delete delegated admin accesss record.
+ */
+export const deleteAndRefreshDelegatedAdmin = async (
+    accessPrivilegegId: number
+// ): Promise<FamAccessControlPrivilegeGetResponse[]> => {
+    ) => {
+    const t = await AdminMgmtApiService.delegatedAdminApi.deleteAccessControlPrivilege(
```
MCatherine1994, line 139: maybe no need "const t"

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
J0taFerreira: I followed the other ones, but you're right.

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: yeah, it make sense saying privilege instead of access

MCatherine1994: Hi Jota, let's change the access to privilege for now, `removed role_name from user_name`, also let's update the `deleteUserRoleAssignment` method, I feel it should be remove from, not remove to?

frontend/src/services/fetchData.ts
```diff
@@ -125,3 +126,20 @@ export const fetchDelegatedAdmins = async (
...
 };
+
+/**
+ * This will call api to delete delegatedAdminPrivilege record from backend and fetch again
+ * to refresh the state.
+ * @param accessPrivilegegId id to delete delegated admin accesss record.
+ */
+export const deleteAndRefreshDelegatedAdmin = async (
+    accessPrivilegegId: number
+// ): Promise<FamAccessControlPrivilegeGetResponse[]> => {
```
MCatherine1994, line 137: just wondering why we comment out this line? thanks!!

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: ok, I'll make the change! Thanks!

frontend/src/services/fetchData.ts
J0taFerreira: It should have been removed. Thanks!

MCatherine1994:
Hi Jota, this confirmation popup text need to change as well. The title and the text content, should use "privilege" not "access", thanks!! 
Screen Shot 2024-03-04 at 11 30 33 AM

frontend/src/services/fetchData.ts
MCatherine1994: I think we should keep it, other delete and refresh method all define the return type

MCatherine1994:
Hi Jota, the functionality works well, just that confirmation popup message need to adjust, thanks!!

frontend/src/services/fetchData.ts
J0taFerreira: I added the return type

frontend/src/components/managePermissions/ConfirmDialogText.vue
```diff
@@ -15,7 +15,7 @@ const props = defineProps({
 <template>
     <p>
         Are you sure you want to remove
-        <strong>{{ props.role }}</strong> access to
+        <strong>{{ props.role }}</strong> access from
```
MCatherine1994, line 18: Hi Jota, we shouldn't hardcode it here. Because this component is used by manage regular user access in `UserDataTable` , manage delegated admin in `DelegatedAdminTable` and manage admin access in `ApplicationAdminTable`. So you want to create a new property `customMsg`, by default it's value is "access", and in the case for delegated admin, pass "privilege"

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
```diff
@@ -50,18 +58,44 @@ const delegatedAdminFilters = ref({
...
+const deleteDelegatedAdmin = (
+    delegatedAdmin: FamAccessControlPrivilegeGetResponse
+) => {
+    confirmDeleteData.adminName = delegatedAdmin.user.user_name;
+    confirmDeleteData.role = delegatedAdmin.role.parent_role
+        ? delegatedAdmin.role.parent_role.role_name
+        : delegatedAdmin.role.role_name;
+    confirm.require({
+        group: 'deleteDelegatedAdmin',
+        header: 'Remove Access',
```
MCatherine1994, line 80: and the header should be `Remove Privilege` as well

MCatherine1994 (approved):

Issue #1179, opened 2024-02-06
Log into FAM with a Business BCeID account


OlgaLiber2:
**Describe the task**
Log into FAM with a Business BCeID account

**Acceptance Criteria**
- [x] Enable "Login with BCeID" option on FAM login screen
- [x] Enable business BCeID identity provider for FAM application
- [x] Have Cognito map attributes from the identity providers claim into the FAM identity token
- [x] Confirm FAM user profile screen shows available information for the current BCeID user based on the token. (create a new ticket if the profile doesn't display info properly)
- [x] Requires delegated admin  - same behavior as IDIR login without permissions 

**Additional context**

- Based on our BCeID Service Agreement, we'll have access to the following fields: user GUID, user ID, user display name, email, business GUID, business legal name (and asking for 'also known as'). We'll want to map all of these fields. 

- In the user profile screen we'll want to show all of these attributes except the GUIDs.

MCatherine1994:
We already have the attribute mapping, just need to add `custom:idp_business_id` to the read and write attributes so Cognito can override it and return to us. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/255a8791-fb3a-42bb-af3d-8fd6f50bb1eb)
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/080948b7-d7e0-4be7-afa5-ce796959b2e7)

PR #1225, opened 2024-03-06
feat: #1179 bceid login

MCatherine1994:
refs: #1179  
- Support business bceid login in FAM, use TEST-BUSINESSBCEID for local and dev environment, as we don't have any DEV business bceid for testing and playing around. To keep consistency we use TEST-IDIR for local and dev environment as well, so the same logout chain could work for both. Update the backend local cognito user id to use "test-idir", update backend test to use "test-idir" as well
- Update the profile to display more information
Screen Shot 2024-03-18 at 10 12 14 AM
Screen Shot 2024-03-18 at 10 25 23 AM

- Admin management endpoint `get_admin_user_access` doesn't need to check access group in the token, delegated admin won't have any access in the token. That API will return the access for the login user. Created a new router guard method to return requester data without check the access role in the token


frontend/public/env.json
```diff
@@ -43,6 +43,11 @@
         "type": "string",
         "value": "DEV-IDIR"
     },
+    "fam_console_idp_name_bceid": {
+        "sensitive": false,
+        "type": "string",
+        "value": "TEST-BCEIDBUSINESS"
```
MCatherine1994, line 49: use TEST-BCEIDBUSINESS for local, so we can test the bceid login, we don't have any dev business bceid

infrastructure/server/oidc_clients_fam.tf
```diff
@@ -3,22 +3,22 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
...
+  ))
+  logout_urls                                   = var.fam_logout_urls
   enable_propagate_additional_user_context_data = "false"
   enable_token_revocation                       = "true"
   explicit_auth_flows                           = ["ALLOW_REFRESH_TOKEN_AUTH"]
   id_token_validity                             = "60"
   name                                          = "fam_console"
   prevent_user_existence_errors                 = "ENABLED"
-  read_attributes                               = "${concat(var.minimum_oidc_attribute_list, ["custom:idp_display_name", "email"])}"
+  read_attributes                               = concat(var.minimum_oidc_attribute_list, ["custom:idp_display_name", "email", "custom:idp_business_id", "custom:idp_business_name"])
```
MCatherine1994, line 19: add business guid and business name to read and write attributes for Cognito to return back

infrastructure/server/oidc_clients_fam.tf
```diff
@@ -3,22 +3,22 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
...
   refresh_token_validity                        = "24"
-  supported_identity_providers                  = [ var.fam_console_idp_name ]
+  supported_identity_providers                  = [var.fam_console_idp_name, var.fam_console_idp_name_bceid]
```
MCatherine1994, line 21: add bceid as identity provider

server/admin_management/api/app/jwt_validation.py
```diff
@@ -211,3 +210,9 @@ def get_request_cognito_user_id(claims: dict = Depends(authorize)):
     cognito_username = claims[COGNITO_USERNAME_KEY]
     LOGGER.debug(f"Current requester's cognito_username for API: {cognito_username}")
     return cognito_username
+
+
+def get_request_cognito_user_id_without_access_check(claims: dict = Depends(validate_token)):
+    cognito_username = claims[COGNITO_USERNAME_KEY]
+    LOGGER.debug(f"Current requester's cognito_username for API: {cognito_username}")
+    return cognito_username
```
MCatherine1994, line 218: get user's cognito username without checking the access role in the token, so this method depends on method validate_token, instead of method authorize

terraform/dev/terragrunt.hcl
```diff
@@ -40,6 +40,7 @@ generate "dev_tfvars" {
     "${local.common_vars.inputs.idp_logout_chain_dev_url}http://localhost:5173"
   ]
   fam_console_idp_name = "DEV-IDIR"
+  fam_console_idp_name_bceid = "TEST-BCEIDBUSINESS"
```
MCatherine1994, line 43: dev and tools environment use TEST-BCEIDBUSINESS for us to play around

basilv:
@MCatherine1994 I didn't understand your comment "Note: Our logout chain in local and dev connects with the dev environment, so the logout for test business bceid will only log the user out from Cognito but still have the sitemider session. " - why wouldn't local/dev logout chain be configured for the test business BCeID environment? Is that because the dev keycloak instance is only pointing at dev BCeID, not test?

frontend/src/components/common/ProfileSidebar.vue
```diff
@@ -61,9 +64,14 @@ const adminRoles = computed(() => {
                 />
                 <div class="profile-info">
                     <p class="profile-name">{{ displayName }}</p>
-                    <p class="profile-idir">IDIR: {{ userName }}</p>
-                    <p class="profile-email">{{ email }}</p>
-                    <p class="profile-admin-level">Granted: <strong>{{ adminRoles }}</strong></p>
+                    <p class="profile-userid">UserID: {{ userName }}</p>
```
basilv, line 67: I think we should show the IDP type here (IDIR or Business BCeID).

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -83,6 +86,26 @@ async def get_current_requester(
     return requester
 
 
+# for delegated admin, there is no access group in the token, our auth lambda function only add app admin to the token
```
basilv, line 89: I thought we were adding a generic DELEGATED_ADMIN group to the token for delegated admins

MCatherine1994: Hi Basil! We don't add "DELEGATED_ADMIN" to the token currently, and I feel it's okay. Because when we check if the user can make the api call or not, he needs to be the delegated admin of the selected application, with just "DELEGATED_ADMIN" in the token is not enough. For app admin, we do show he's the admin of which application ("FOM_DEV_ADMIN", "FAM_ADMIN"), so that's fine to check the token. But for delegated admin, we can't just use the "DELEGATED_ADMIN" anyway, need to check the real access.

basilv:
> > @MCatherine1994 I didn't understand your comment "Note: Our logout chain in local and dev connects with the dev environment, so the logout for test business bceid will only log the user out from Cognito but still have the sitemider session. " - why wouldn't local/dev logout chain be configured for the test business BCeID environment? Is that because the dev keycloak instance is only pointing at dev BCeID, not test?
> 
> Hi Basil, that's just because for our local/dev environment, we suppose to use DEV-IDIR and DEV-BUSINESS-BCEID, and they work with the same DEV logout chain. But because we don't have dev bceid for testing, I setup to use DEV-IDIR and TEST-BUSINESS-BCEID for local/dev environment. So DEV-IDIR needs dev logout chain, but TEST-BUSINESS-BCEID needs test logout chain. The Amplify just configured with one redirect logout url doesn't support both in the current code. And I feel it's fine so I didn't change the Amplify config. Because for TEST env, IDIR and BCEID will both connect with TEST, and same for PROD. So the issue is only in DEV. Now I'm thinking maybe we can just use TEST-IDIR and TEST-BUSINESS-BCEID for local and dev, and config the test logout chain, so logout can work properly for both. And use DEV orTEST IDIR won't impact the login behaviour?

This makes sense, just as long as this is documented somewhere in the dev configuration.

server/admin_management/api/app/routers/router_guards.py
basilv: I know we were talking about it before, so I just wanted to make sure. What you've done above seems fine.

MCatherine1994: thanks!! So for the admin management apis, they all need app admin role, just that get my access api doesn't need to check role, because the purpose of that api is to return the login user's access. 

For the apis under our server/backend, which will be used by delegated admin/app admin to do user access management, we need to adjust the code to check the real access. We have ticket for that already in backlog. Just want to mention it here as a note.

basilv (approved):

Issue #1150, opened 2024-01-22
Menu item to navigate to grant delegated admin screen (frontend)


OlgaLiber2:
**Describe the task**
Left sidebar includes delegated admin option

**Acceptance Criteria**
- [x] sidebar item created for delegated admin
- [x] navigation to grant delegated admin page
- [x] Show the delegated admin management functionality only if I'm an app admin (not including FAM admins)
- [x] After creating a delegated admin and go back to the dashboard page, it should stay under the delegated admin tab. After granting a regular user access and get back to the dashboard page, it should stay under the users tab.

**Additional context**
Story: As an app admin I want to review the list of delegated admins for my application 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/40109b71-8816-4dac-9e78-9f0e8730b480)

Depends on #1159 for creation of delegated admin page

PR #1228, opened 2024-03-08
feat: #1150 navigation to grant delegated admin screen

NickSaglioni:
- sidebar item created for delegated admin page;
- navigation to grant delegated admin page from dashboard table button;
- show the delegated admin management functionality only if I'm an app admin (not including FAM admins);
- selecting correct tab in dashboard page after creating a delegated admin or granting a regular user access;

### DIFF
--- a/frontend/src/components/grantaccess/GrantAccess.vue
+++ b/frontend/src/components/grantaccess/GrantAccess.vue
@@ -146,7 +146,7 @@ const handleSubmit = async () => {
         errorCode,
         role
     );
-    setCurrentTabState(TabKey.User);
+    setCurrentTabState(TabKey.UserAccess);
     router.push('/dashboard');
 };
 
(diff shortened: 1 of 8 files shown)


MCatherine1994:
Hi Nick, I think the tab works well. Just the side nav, I'm not too sure why yet. If I'm the app admin of FOM_DEV, and delegated admin of FOM_TEST. When I first select FOM_DEV, side nav gives me options for "add user permission" and "add delegated admin", which is correct. But when I switched to FOM_TEST, the "add delegated admin" option is still there.... 

Screen Shot 2024-03-08 at 1 53 08 PM

MCatherine1994:
Oh so here, need to add if the user is not the admin of the selected app `disableSideNavOption('Add delegated admin', true);`   in a else condition

image

MCatherine1994 (approved):

PR #1229, opened 2024-03-08
feat: #1147 get list of roles from cache

J0taFerreira:
- switch from using api call to using cache information
- removed roles attribute from FamLoginUser
- removed unused fetchApplicationRoles

### DIFF
--- a/frontend/src/components/grantaccess/GrantAccess.vue
+++ b/frontend/src/components/grantaccess/GrantAccess.vue
@@ -30,7 +30,15 @@ const defaultFormData = {
 };
 const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
 const regularUserRoleOptions = computed(() => {
-    return FamLoginUserState.getCachedAppRoles(selectedApplicationId.value!);
+    if (FamLoginUserState.isAdminOfSelectedApplication()) {
+        return FamLoginUserState.getCachedAppRoles(
+            selectedApplicationId.value!
+        );
+    } else {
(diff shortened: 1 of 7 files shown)


frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -36,6 +29,9 @@ const defaultFormData = {
     roleId: null as number | null,
 };
 const formData = ref(JSON.parse(JSON.stringify(defaultFormData))); // clone default input
+const regularUserRoleOptions = computed(() => {
+    return FamLoginUserState.getCachedAppRoles(selectedApplicationId.value!);
```
MCatherine1994, line 33: Hi Jota, we might can't reuse the existing `getCachedAppRoles` method here. Because the current `getCachedAppRoles` is getting roles for app admin, which can have access to all roles. But in our case here, the user could just be the delegated admin of the selected application, and it's possible he can only manage a few roles, not all. Sorry I missed this part as well. So I think we need to first check if user is the app admin or delegated admin of the selected app, and then we decide to get roles from auth_key "APP_ADMIN" or "DELEGATED_ADMIN". 

So basically you can call the `isAdminOfSelectedApplication` method first, and if return true, then we keep what we have, and if false, then the user will be delegated admin. (Cause as long as he can select that app, he will be app admin or delegated admin of it.) And then you just create a new method `getCachedAppRolesForDelegatedAdmin` and call it

image

J0taFerreira: Hello, Catherine! I'll make the change. Thanks!

MCatherine1994 (approved):

PR #1255, opened 2024-03-19
fix: #1179 bceid login bug fix

MCatherine1994:
refs: #1179 

Fix the bug found after deployment to dev:
- added to show a toast error message in the case the login user has no access in FAM
Screen Shot 2024-03-19 at 12 59 54 PM

- fix the frontend redirect logout chain url config for environment variables

### DIFF
--- a/infrastructure/server/variables_provided.tf
+++ b/infrastructure/server/variables_provided.tf
@@ -206,12 +206,6 @@ variable "cognito_app_client_logout_chain_url" {
 
 # Variables for FAM front-end config
 
-variable "frontend_logout_chain_url" {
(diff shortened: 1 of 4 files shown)


infrastructure/server/variables_provided.tf
```diff
@@ -206,12 +206,6 @@ variable "cognito_app_client_logout_chain_url" {
 
 # Variables for FAM front-end config
 
-variable "frontend_logout_chain_url" {
```
MCatherine1994, line 209: I think we are not using this anymore

basilv (approved):

Issue #1182, opened 2024-02-06
Hide BCeID user details if not the same organization


OlgaLiber2:
**Describe the task**
As a delegated admin I want to only be able to see users within my organization (and their first and last names).

**Acceptance Criteria**
- [x] Update the Backend API for searching bceid user to check: if the requester's user type is B, then if his organization (business guid) is the same as the search result (return business guid)
- [x] Only return the search result with detail information if the business guid match
- [x] If not match, return error message to specify the permission issue because they are not in the same organization.
- [x] add test

**Additional context**
- Provide the FAM backends a way to access the user's identity type (IDIR vs Business BCeID) and a way to determine the Business GUID Main options are: 1. add to access token, 2. retrieve from Cognito via Cognito API (haven't validated if feasible but Conrad had mentioned as an option), 3. store and retrieve from user table - which would require a PIA update.
- FAM backend APIs dealing with BCeID user information need to check if the current user (if they are a BCeID user) has a business GUID (from their token) that matches the user record(s). If no match then remove attributes with personal information.
- This applies to any scenario where BCeID user details are displayed:

On validation of the id when granting access
On displaying user first & last name for granted user access. (we will try to make this check reusable, but this ticket can just target on the search bceid method)

We'll use the BCeID Business GUID attribute to check the organization of each user.

For any record not matching, redact / blank out / hide / remove attributes of personal information (PIA states that "no information about the result will be displayed", but we can at least confirm that the user id is valid). The result returned to the front-end will specify that information about the users cannot be displayed because they are not in the same organization.

Must be done after #1222 (backend)

See also #1184

basilv:
I was thinking we'd have the guid stored in the fam users table and could fetch it from there. Or add it as a claim to the access token, that's probably the best approach.

basilv:
Ian, you are right that we're missing this in the stories - I've added this to the delegated admin page. I keep flipflopping on which approach is best (add to access token vs lookup from user table). We should at least add the user identity type (IDIR vs Business vs ...) to the FAM access token, as that allows us for IDIR users to avoid needing to check the business GUID.

basilv:
There's a third option, which is to query Cognito for the user information (Business GUID). If we were to store in our Postgres database, that would require a PIA update.

ianliuwk1019:
Thank you Basil. 
Is the PIA is going to be another long process? It seems to me storing user's Business GUID is the best option.

[...]

PR #1269, opened 2024-03-26
feat: #1182 bceid search bceid need check org

MCatherine1994:
refs: #1182 

- Add check for bceid search for bceid, only return the search result if from same organization
- Add tests

Search for business bceid within same organization:
Screen Shot 2024-03-21 at 3 24 15 PM
Search for business bceid from different organization:
Screen Shot 2024-03-21 at 3 24 24 PM


MCatherine1994:
I wondering if we have the case the bceid user will search for himself? Because the IDIM api, if bceid search for self, the business guid will not be returned, I don't know why, maybe this scenario is not fully enabled, so it's not working properly?
In theory, the bceid user can't grant himself access anyway, but if it's still possible a bceid user lookup for himself, then I need to ask the IDIM consulting team why the business guid doesn't return in this case.

basilv:
> I wondering if we have the case the bceid user will search for himself? Because the IDIM api, if bceid search for self, the business guid will not be returned. I don't know why, maybe this scenario is not fully enabled, so it's not working properly? In theory, the bceid user can't grant himself access anyway, but if it's still possible a bceid user lookup for himself, then I need to ask the IDIM consulting team why the business guid doesn't return in this case.

From a business standpoint, we might want the delegated admin to be able to grant access to themselves. (From a security standpoint, the risk is limited because the roles they can grant are restricted, unlike for FAM admins/app admins.). This likely requires discussion with Olga.

If we do want to allow self-granting, then yes we'll need the BCeID user lookup to work, so I'd say ask IDIM consulting why this doesn't work currently.

basilv (approved):

Issue #1266, opened 2024-03-22
Update create_user and update_user length


MCatherine1994:
**Describe the task**
We use cognito_user_id as create_user and update_user, and the cognito_user_id length can up to 100 chars, but currently we only allow 60 chars for create_user and upate_user in all tables. When login with business bceid, the cognito_user-id is longer than 60 chars, db complains the the length.

**Acceptance Criteria**
- [ ] Update all the table has create_user and update_user to allow 100 chars
- [ ] Update data model in both `server/admin_management` and `server/backend` 
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

MCatherine1994:
`fam_app_environment`, `fam_role_type` and `fam_user_type_code` has no create_user and update_user column
`fam_application_admin` no need to change, cause no bceid can grant application admin

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/f13160a3-6e8d-48d1-9466-db4a98a2a11a)

MCatherine1994:
Hi Ian, yeah, but the real cognito user id is having the env in front of it, so it's like `test-bceidbusiness_532905de0aa24923ae535428f171bf13@bceidbusiness`. And the cognito_user_id column is having char(100)

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/8636155f-4cc7-4643-9a17-f5f37ecbdb0c)

PR #1275, opened 2024-03-27
fix: #1266 update create_user and update_user length in db

MCatherine1994:
refs: #1266

Update create_user and update_user length in tables, so business bceid user's cognito_user_id can be inserted as create/update user

`fam_app_environment`, `fam_role_type` and `fam_user_type_code` has no create_user and update_user column
`fam_application_admin` no need to change, cause no bceid can grant application admin, but for consistency, updated it as well. 

Image

### DIFF
--- a/server/flyway/sql/V45__create_and_update_user_length_increase.sql
+++ b/server/flyway/sql/V45__create_and_update_user_length_increase.sql
@@ -44,3 +44,6 @@ ALTER TABLE app_fam.fam_access_control_privilege
     ALTER COLUMN create_user SET DATA TYPE VARCHAR(100),
     ALTER COLUMN update_user SET DATA TYPE VARCHAR(100);
 
+ALTER TABLE app_fam.fam_application_admin
+    ALTER COLUMN create_user SET DATA TYPE VARCHAR(100),
+    ALTER COLUMN update_user SET DATA TYPE VARCHAR(100);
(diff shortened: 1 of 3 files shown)


MCatherine1994:
I wonder for the `fam_application_admin` table, there is no need to change the create/update user length, cause no bceid can grant application admin. But do we need to keep consistency for the length and update to 100 chars as well?

basilv:
I agree there's no need to change FAM_APPLICATION, but its easier for consintency to have all these audit columns the same length, so lets ensure that audit columns in all tables are updated.

server/flyway/sql/V45__create_and_update_user_length_increase.sql
```diff
@@ -0,0 +1,46 @@
+-- Increase create_user and update_user to (100) because business bceid user's cognito_user_id is longer than 60 chars
+
+ALTER TABLE app_fam.fam_application
+    ALTER COLUMN create_user SET DATA TYPE VARCHAR(100),
+    ALTER COLUMN update_user SET DATA TYPE VARCHAR(100);
+
+ALTER TABLE app_fam.fam_application_client
```
basilv, line 7: Missing alter table for fam_application_admin (model was changed).

basilv (changes requested):
add fam_application_admin and any other tables with audit columns to sql script

basilv (approved):

Issue #985, opened 2023-10-23
Check my own permissions for App/Delegated Admin 


OlgaLiber2:
![Capture-my-permissions](https://github.com/bcgov/nr-forests-access-management/assets/121979429/8eaad96f-f65e-4a26-9d05-20e7b87d3890)

AC:

- [x] add new router and check if need router handler
- [x] add to side nav bar
- [x] enable global table search
- [x] create new table for view own access using cached user access info
- [x] if user has no access in FAM, he can still access to this page, and we'll display a text message shows that the user has no access in FAM

Additional context:
user access sample: https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Delegated+Access+Administration+Design

OlgaLiber2:
[Uploading... my permissions - v1.png]

MCatherine1994:
fyi, Figma design page: https://www.figma.com/file/5BnF5T2ncoZVpSbCbj30Mr/FAM-mockups?type=design&node-id=1845-427569&mode=design&t=SJ9H0inI21b80q4n-0

J0taFerreira:
When the user don't have any permission, should it display something like this?
![Image](https://github.com/bcgov/nr-forests-access-management/assets/126905995/c4fa7c09-7dce-4e08-b659-a3715659d874)

or just show a text passed to the primevue table like that?
![Image](https://github.com/bcgov/nr-forests-access-management/assets/126905995/053773e9-6349-4765-9fdd-51f8150c2c0b)

MCatherine1994:
Hi @OlgaLiber2, I think we need a little bit more discussion, for this view my access screen, is it just for app admin or delegated admin to check their access to manage app or user?  Or it's also for general users, for example if a user has no access in FAM, but has a FOM_REVIEWER role, can this user check the access from this page?

PR #1276, opened 2024-04-02
feat: #985 check my own permissions

J0taFerreira:
- new router and check if need router handler
- "my permissions" to side nav bar
- global table search
- new table for view own access using cached user access info
- if user has no access in FAM, he can still access to this page, and we'll display a text message shows that the user has no access in FAM


frontend/src/components/myPermissions/MyPermissions.vue
```diff
@@ -0,0 +1,12 @@
...
+import myPermissionsTable from '@/components/myPermissions/myPermissionsTable.vue';
+import PageTitle from '@/components/common/PageTitle.vue';
+</script>
+
+<template>
+    <PageTitle
+        title="My permissions"
+        subtitle="Check all the applications and permissions that you have access to"
+    />
+    <myPermissionsTable />
```
MCatherine1994, line 11: Hi Jota, component name needs to use capital letter, MyPermissionTable

MCatherine1994:
Screen Shot 2024-04-02 at 2 48 27 PM
Hi Jota, I feel the data I get doesn't seem correct... but let's confirm with Olga tomorrow, to see if this table should return my admin access or my regular app access, thanks!!

frontend/src/components/myPermissions/MyPermissions.vue
J0taFerreira: oh, my bad! I'll fix it, thanks!

frontend/src/router/index.ts
```diff
@@ -110,6 +111,17 @@ const routes = [
         component: GrantDelegatedAdminView,
         beforeEnter: beforeEnterHandlers[routeItems.grantDelegatedAdmin.name],
     },
+    {
+        path: routeItems.myPermissions.path,
+        name: routeItems.myPermissions.name,
+        meta: {
+            requiresAuth: false,
```
MCatherine1994, line 118: Hi Jota, we still need authentication, so requiresAuth means this page needs user login first, all our routes need login (landing page doesn't need)

frontend/src/views/MyPermissionsView.vue
```diff
@@ -0,0 +1,7 @@
+<script setup lang="ts">
+import MyPermissions from '@/components/myPermissions/MyPermissions.vue';
+</script>
+
+<template>
+    <MyPermissions></MyPermissions>
```
MCatherine1994, line 6: This is small, but if we don't pass properties and has no child elements, it can just be  <MyPermissions />

frontend/src/components/myPermissions/MyPermissionsTable.vue
```diff
@@ -0,0 +1,110 @@
...
+import { FilterMatchMode } from 'primevue/api';
+import {
+    TABLE_CURRENT_PAGE_REPORT_TEMPLATE,
+    TABLE_PAGINATOR_TEMPLATE,
+    TABLE_ROWS_PER_PAGE,
+} from '@/store/Constants';
+import { isLoading } from '@/store/LoadingState';
+
+import Column from 'primevue/column';
+import DataTable from 'primevue/datatable';
```
MCatherine1994, line 12: and let's put the import of primevue component at top

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,24 @@ const getCachedAppRolesForDelegatedAdmin = (
             ? -1
             : 1;
     });
-}
+};
+
+const getMyCachedPermissions = () => {
```
MCatherine1994, line 212: Hi Jota, I think this method is still not correct. Not sure if you tested locally, but when I have just one FAM_ADMIN role, it's not showing in this table. 
Screen Shot 2024-04-05 at 10 07 26 AM

And when I have FAM_ADMIN and FOM_ADMIN access, this is also not correct. 
Screen Shot 2024-04-05 at 10 08 09 AM

So the `access` example you can check our confluence page at the bottom section https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Delegated+Access+Administration+Design, basically it will have three types, you need to first check the `auth_key`, auth_key can be "FAM_ADMIN", "APP_ADMIN" and "DELEGATED_ADMIN", and we need to handle them separately

First of all, you want to show an example of you data structure, like what you'll pass to the table component. And then for each auth key cases, we need to grab the information and format them 
1. If auth_key is FAM_ADMIN, then you just find the FAM's application description from the grants, filter the grant and return the application where name is FAM, the role name will just be `ADMIN` (hardcode this)
2. If auth_key is APP_ADMIN, then you want the information of each application in the grants, the role name will be `ADMIN` as well, no need to check the
3. If auth_key is DELEGATED_ADMIN, then for each grant, you want to application information as well as the role information

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,24 @@ const getCachedAppRolesForDelegatedAdmin = (
...
+                Object.assign(grant.application, {
+                    roles: role,
+                });
+                myRoles.push({
+                    application: grant.application,
+                });
+            });
+        });
+    });
+    return myRoles;
```
MCatherine1994, line 226: For example your return data can be in the format of 
```
{
application: {id: 1, name: 'FAM', description: 'Forests Access Management', env: null},     // this is exact same as the application info in user's access
role: {id: 4, name: 'Reviewer',type_code: 'C','forest_clients': null} // this is also the same as the role info in user's access
}
```
For FAM_ADMIN and APP_ADMIN case, you just add a role field, `role: {name: 'Admin'}`

frontend/src/views/MyPermissionsView.vue
J0taFerreira: Sure, I'll change it to self closing

frontend/src/store/FamLoginUserState.ts
J0taFerreira: oh, I see. I'll fix it following your comment

frontend/src/store/FamLoginUserState.ts
J0taFerreira: got it! Thanks!

NickSaglioni (changes requested):
I think the table margins are not correct, it should have some spacing from the sidenav and the right page border.

Screenshot 2024-04-09 at 15 17 19

J0taFerreira:
> I think the table margins are not correct, it should have some spacing from the sidenav and the right page border.
> Screenshot 2024-04-09 at 15 17 19

Hello! That's the way in the design, and do seen out off place compared to the other tables, maybe we should ask @OlgaLiber2

OlgaLiber2:
@NickSaglioni @J0taFerreira yes, it should have some spacing on the right and left like in this example:
My permissions

frontend/src/components/myPermissions/MyPermissionsTable.vue
```diff
@@ -0,0 +1,116 @@
...
+    },
+    'env': {
+        value: null,
+        matchMode: FilterMatchMode.CONTAINS,
+    },
+    'clientId': {
+        value: null,
+        matchMode: FilterMatchMode.CONTAINS,
+    },
+    'adminRole': {
```
MCatherine1994, line 30: Hi Jota, let's not called it admin Role, because later on we'll use the same field for regular role as well, let's just call it role

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,61 @@ const getCachedAppRolesForDelegatedAdmin = (
...
+        if (item.auth_key === AdminRoleAuthGroup.DelegatedAdmin) {
+            item.grants.map((grant) => {
+                grant.roles?.forEach((role) => {
+                    if(!role.forest_clients) {
+                        myRoles.push({
+                            adminRole: 'Delegated Admin',
+                            application: grant.application.description,
+                            env: grant.application.env,
+                            clientId: null,
+                            role: role.name,
```
MCatherine1994, line 246: oh I see what you're trying to do here, maybe let's just use `role: 'Delegated Admin, ' + role.name`

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,61 @@ const getCachedAppRolesForDelegatedAdmin = (
             ? -1
             : 1;
     });
-}
+};
+
+const getMyCachedPermissions = () => {
+    let myRoles: any = [];
```
MCatherine1994, line 213: maybe it's better to call it myPermissions?

MCatherine1994:
Hi Jota, just some suggestions for future, so for this kind of pr that we have 4 difference cases (no permission, FAM admin, app admin, delegate admin), it will be great if we can share some screenshots of how it looks like in the pr description or in the comment, it will help the review process and in the future when we look back. Thanks so much!

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,61 @@ const getCachedAppRolesForDelegatedAdmin = (
             ? -1
             : 1;
     });
-}
+};
+
+const getMyCachedPermissions = () => {
```
MCatherine1994, line 212: and let's rename this to getMyAdminPermission, cause later on we'll need to get regular permission as well, thanks!

MCatherine1994:
This looks pretty good! thanks!!
image

frontend/src/store/FamLoginUserState.ts
```diff
@@ -207,7 +207,61 @@ const getCachedAppRolesForDelegatedAdmin = (
...
+    let myRoles: any = [];
+    getUserAccess()?.map((item) => {
```
MCatherine1994, line 214: I was checking the different between array.map vs array.forEach, I think if we're not returning any value, no need to use map?
Screen Shot 2024-04-10 at 10 28 58 AM

frontend/src/components/myPermissions/MyPermissionsTable.vue
J0taFerreira: Hello, I'll change it to role.

frontend/src/store/FamLoginUserState.ts
J0taFerreira: good one! Thanks!

frontend/src/store/FamLoginUserState.ts
J0taFerreira: hum...good to know, I'll use forEach in this case Thanks!

MCatherine1994 (approved):

NickSaglioni (approved):
Looks good now, thanks Jota!

Issue #1272, opened 2024-03-27
Spike: debug bcsc login issue in prod


MCatherine1994:
**Describe the task**
Meet with Wesley to check the bcsc login issue in production

**Acceptance Criteria**
- [x] Start with a meeting with Wesley 

**Additional context**
- Some thoughts: removing gender attribute might caused this error

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/af72a599-e45b-47a4-83f0-38bebe6814ee)

- Our api log actually returns the user's information 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/94439c9c-8045-4e15-a6c8-d6452722b03a)

MCatherine1994:
Tried a few things:

- Tried to add username to the read/write attribute permission, but got pipeline error username is not a valid read/write attribute.
- Tried to remove the username attribute mapping from terraform, and it's auto added in the mapping, and no difference
- Tried to remove all attribute mapping and login again, got the same error
- Also chatted with Wesley, prod login using mobile app shouldn't be the reason of this error

Submitted a AWS support request through AWS Prod, https://support.console.aws.amazon.com/support/home?region=ca-central-1#/case/?displayId=171218031100513&language=en

MCatherine1994:
Just attached the email from Alan as our note, it's the summary of our meeting



![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/8161ab01-f959-432a-8aed-7b429174e061)

MCatherine1994:
Ian found out we had an error in our code while reviewing the PROD-BCSC config in terraform: 

We hardcoded the IDP name for PROD-BCSC long time ago, but forgot to update it to the real one Wesley shared with us recently. So when the user_info endpoint tries to verify the "aud" received from bcsc, it throws a 401, and the PROD-BCSC IDP name doesn't match what we have in the code. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/4c1b8da3-c6bf-4574-9a15-e80043b5cb55)


![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/9291d5f8-e382-4cad-9b27-6abf14618095)

[...]

PR #1277, opened 2024-04-02
fix: #1272 try to add username attribute for mapping

MCatherine1994:
refs: #1272

One of my guess of the bcsc login error, is because of this "Attribute read and write permissions" mapping, it doesn't have the username in it. So if Cognito wants to do something with the username field but has no permission to do so, it might throw the error we got?  But I can't manually add it through the UI, so I wonder if I could just update the terraform to include it and deploy to DEV. The "maximum_oidc_attribute_read_list" and "maximum_oidc_attribute_write_list" are only used by forest client app integration, so it won't impact other apps. I also found out how to configure our AWS DEV cognito, so we can use AWS DEV to verify the bcsc login, no need to deploy to AWS TEST and PROD.

If it doesn't work I'll revert the code back. Again, forest-client-test and forest-client-prod has exact same read/write attribute permission, I don't know why TEST-BCSC login with forest-client-test works fine, but PROD-BCSC login with forest-client-prod doesn't work. This pr is just my guess. 

Screen Shot 2024-04-02 at 3 02 36 PM

### DIFF
--- a/infrastructure/server/variables_provided.tf
+++ b/infrastructure/server/variables_provided.tf
@@ -272,7 +272,8 @@ variable "maximum_oidc_attribute_read_list" {
     "profile",
     "updated_at",
     "website",
-    "zoneinfo"
+    "zoneinfo",
+    "username"
   ]
 
 }
@@ -305,7 +306,9 @@ variable "maximum_oidc_attribute_write_list" {
     "profile",
     "updated_at",
     "website",
-    "zoneinfo"]
+    "zoneinfo",
+    "username"
+  ]
 }
 
 # Variables for connecting Cognito to BCSC OIDC


basilv:
I'm concerned as I thought I read somewhere that permissions can't be removed (editted?) once the user pool is created. I'll have to double-check, but it might not be possible to roll back the changes within Cognito.

MCatherine1994:
> I'm concerned as I thought I read somewhere that permissions can't be removed (editted?) once the user pool is created. I'll have to double-check, but it might not be possible to roll back the changes within Cognito.

oh no then that will be bad.... but however this read/write attribute is used for app integration config, I feel it doesn't make sense to not allow change... it should be configurable 
image

basilv:
Its okay - as per other chat its the required attributes that can't be changed which is different.

basilv (approved):

PR #1279, opened 2024-04-03
fix: #1272 revert read write attribute

MCatherine1994:
refs: #1272 

Got a pipeline error, seems like "username" is not a valid read/write attribute, revert the code back
Screen Shot 2024-04-03 at 10 08 49 AM

Have another guess, noticed that for bceid user in Cognito, their user attributes have a `sub` field, and it is required, but for the test bcsc user, I didn't see a `sub` field in the user attributes. 
Screen Shot 2024-04-03 at 10 24 35 AM
Screen Shot 2024-04-03 at 10 24 40 AM

Screen Shot 2024-04-03 at 10 25 57 AM

And checked our IDP config terraform file, for IDIR and BCEID, we didn't do a username mapping to sub, think Cognito will auto map it, so want to try to remove the username mapping to sub from the BCSC IDP config as well, and see if that brings any difference. 
Screen Shot 2024-04-03 at 10 26 43 AM

### DIFF
--- a/infrastructure/server/oidc_idp_bcsc.tf
+++ b/infrastructure/server/oidc_idp_bcsc.tf
@@ -24,7 +24,6 @@ resource "aws_cognito_identity_provider" "dev_bcsc_oidc_provider" {
   }
 
   attribute_mapping = {
-    username = "sub",
     given_name = "given_name",
     family_name = "family_name",
     birthdate = "birthdate",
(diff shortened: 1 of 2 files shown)


infrastructure/server/oidc_idp_bcsc.tf
```diff
@@ -24,7 +24,6 @@ resource "aws_cognito_identity_provider" "dev_bcsc_oidc_provider" {
   }
 
   attribute_mapping = {
-    username = "sub",
```
MCatherine1994, line 27: remove the username sub mapping from the BCSC IDP mapping, cause we don't have this in the IDIR and BCEID config file, and Cognito auto adds it seems like

basilv (approved):

PR #1290, opened 2024-04-08
fix: fix the table header text for delegated admin

MCatherine1994:
fix the table header for delegated admin, we don't want to confuse user for this prod deployment

image

image

### DIFF
--- a/frontend/src/components/managePermissions/table/DataTableHeader.vue
+++ b/frontend/src/components/managePermissions/table/DataTableHeader.vue
@@ -2,6 +2,7 @@
 import { computed } from 'vue';
 import InputText from 'primevue/inputtext';
 import router from '@/router';
+import { routeItems } from '@/router/routeItem';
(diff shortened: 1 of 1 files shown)


frontend/src/components/managePermissions/table/DataTableHeader.vue
```diff
@@ -12,33 +13,37 @@ const props = defineProps({
...
     set(newValue) {
-        emit('change', newValue)
-    }
-})
+        emit('change', newValue);
+    },
+});
 
+const userLevelText = computed(() => {
+    if (props.btnRoute == routeItems.grantDelegatedAdmin.path)
```
MCatherine1994, line 35: I was thinking if pass another property or just check the button route to determine if this is for delegated admin table or regular table, feel both okay. And not sure if this naming makes sense, I'm bad at naming...

ianliuwk1019 (approved):
Thanks Catherine!

PR #1291, opened 2024-04-09
fix: bceid login for FOM production login issue

MCatherine1994:
FOM reports a bceid login issue:
Screen Shot 2024-04-09 at 9 41 46 AM

Checked the AWS PROD Cloudwatch Auth Lambda log, found an error for the audit log when try to get the business guid. So comment out the audit log for getting the business guid, we might could use `event["request"]["userAttributes"].get("custom:idp_business_id") `, but want to fix the login issue as soon as possible, will work on the audit log later. 

Screen Shot 2024-04-09 at 9 34 42 AM

Tested with my prod business bceid account, and it works fine. So the error could happen when the login user doesn't have a business_guid somehow, we need more investigation on that. Cause we'll need the business_guid for checking organizations later, we expect the business bceid login user should have a business_guid

image


ianliuwk1019 (approved):
Thanks Catherine. Approve it for a quick resolution for PROD issue.

Issue #1293, opened 2024-04-09
FAM support FOM production bceid login loop back error


MCatherine1994:
**Describe the task**
FOM reports a bceid login issue:
<img width="939" alt="Screen Shot 2024-04-09 at 9 41 46 AM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/26bfc933-751a-4d69-9229-5d57f55c73b3">

Checked the AWS PROD Cloudwatch Auth Lambda log, found an error for the audit log when try to get the business guid

<img width="1439" alt="Screen Shot 2024-04-09 at 9 34 42 AM" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/e698efba-a70f-49a4-9a4c-cf5162f2c496">

Tested with my prod business bceid account, and it works fine. So the error could happen when the login user doesn't have a business_guid somehow, we need more investigation on that. Cause we'll need the business_guid for checking organizations later, we expect the business bceid login user should have a business_guid (Updated: the error is not caused by login user doesn't have business_guid, it's because we didn't configure it to be returned from Cognito, see the note below in the additional context)

<img width="923" alt="image" src="https://github.com/bcgov/nr-forests-access-management/assets/23131735/ad13267e-1c02-4c4b-9389-d35081fc2b59">

**Acceptance Criteria**
- [x] Quick fix, remove the audit log for business guid
- [x] Update the auth lambda audit log to use "get" for optional attributes that not all applications are configured 
- [x] Add test to check the different log 

**Additional context**
- Business bceid user should have a business guid returned from login, but we have the read/write permission configured for each application integrated with FAM, and only FAM enabled to return the business_guid. So if I login to FAM with my business bceid already, Cognito already has my business_guid, so my login to FOM works as well. But for other bceid users, FOM is not configured to return the business_guid, so our Auth Lambda audit log won't be able to find the `custom:idp_business_id`
- Or here

ianliuwk1019:
Would it be useful trying these failing users' BCeIDs (from different organizations) on our prod IDIM-Proxy swagger and see what's the return (the user type and if it contains business GUID)?

MCatherine1994:
Verified the business bceid user through IDIM web service, and they have business_guid. And the detail of the cause can be found in the ticket description. 

Deployed the quick fixes to prod at v1.14.1, confirmed from Julius the bceid login works now.

PR #1295, opened 2024-04-09
fix: #1293 fix audit log and add test

MCatherine1994:
refs: #1293

- Update the audit log in Auth Lambda to print out attributes that not are configured for each application optional 
- Add two new Cognito user info file with only minimum user attributes, and add them to the test


server/auth_function/lambda_function.py
```diff
@@ -74,7 +77,10 @@ def lambda_handler(event: event_type.Event, context: Any) -> event_type.Event:
 
     LOGGER.debug(f"context: {context}")
 
+    LOGGER.info(f"event: {event}")
```
MCatherine1994, line 80: I'm not sure if we have any concern about the privacy data from this event data, but I feel this can help us debug if there is any issue in the future.

server/auth_function/test/login_event_bceid_min_attr.json
```diff
@@ -0,0 +1,31 @@
...
+                "ca-central-1_p8X8GdjKW_TEST-BCEIDBUSINESS"
+            ],
+            "iamRolesToOverride": [],
+            "preferredRole": null
+        },
+        "userAttributes": {
+            "cognito:user_status": "EXTERNAL_PROVIDER",
+            "custom:idp_name": "bceidbusiness",
+            "custom:idp_user_id": "81069F39B35B4861BCD010582MANGLED",
+            "custom:idp_username": "LOAD-2-TEST",
```
MCatherine1994, line 19: only included the minimum user attribute

server/auth_function/test/login_event_bcsc_min_attr.json
```diff
@@ -0,0 +1,30 @@
...
+            "groupsToOverride": [
+                "ca-central-1_p8X8GdjKW_TEST-BCSC"
+            ],
+            "iamRolesToOverride": [],
+            "preferredRole": null
+        },
+        "userAttributes": {
+            "cognito:user_status": "EXTERNAL_PROVIDER",
+            "custom:idp_name": "ca.bc.gov.flnr.fam.test",
+            "custom:idp_user_id": "6RRZFAU3ANJ5WASQ4N22GKJXSMANGLED",
```
MCatherine1994, line 18: only included the minimum user attributes

ianliuwk1019:
Hi @MCatherine1994 ,
Sorry, would you mind refactor the test file from "lamda_function_test.py" to "lam**b**da_function_test.py" (with a 'b' in lamda). I just found that original file name is wrong.

server/auth_function/lambda_function.py
ianliuwk1019: I am not sure either but feel might be okay to leave it ".info" there for a while since we currently starting to have active production apps users with different logins, might need this to debug.
I think preferably to remove this later and instead to figure out how to enable debug mode in production in future.
Thanks.

server/auth_function/test/lambda_function_test.py
```diff
@@ -70,6 +72,70 @@ def test_create_user_if_not_found(
...
+        cognito_user_id = {} and
+        user_name = {} and
+        business_guid = {};"""
+
+    replaced_query = sql.SQL(raw_query).format(
+        sql.Literal(test_idp_type_code),
+        sql.Literal(test_idp_user_id),
+        sql.Literal(test_cognito_user_id),
+        sql.Literal(test_idp_username),
+        sql.Literal(test_idp_business_guid),
```
MCatherine1994, line 136: Test that for business bceid user login with attribute `custom:idp_business_id` configured readable and writable, it will be added to the database

server/auth_function/test/lambda_function_test.py
```diff
@@ -126,12 +194,66 @@ def test_update_user_if_already_exists(
...
+        user_guid = {} and
+        cognito_user_id = {} and
+        user_name = {} and
+        business_guid = {};"""
+    query = sql.SQL(raw_query).format(
+        sql.Literal(test_idp_type_code),
+        sql.Literal(test_idp_user_id),
+        sql.Literal(test_cognito_user_id),
+        sql.Literal(test_idp_username),
+        sql.Literal(test_idp_business_guid),
```
MCatherine1994, line 240: Test that for business bceid user login with attribute custom:idp_business_id configured readable and writable, if the user is already exists, will update the business_guid for the user, and we can get the user record back

server/auth_function/lambda_function.py
```diff
@@ -155,9 +164,9 @@ def populate_user_if_necessary(db_connection, event) -> None:
         user_name = event["request"]["userAttributes"]["custom:idp_username"]
 
     raw_query = """INSERT INTO app_fam.fam_user
-        (user_type_code, user_guid, cognito_user_id, user_name,
+        (user_type_code, user_guid, cognito_user_id, user_name, business_guid,
         create_user, create_date, update_user, update_date)
-        VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name},
+        VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid},
```
MCatherine1994, line 169: just noticed that, if user not exists, we need to insert new user record with the business_guid

server/auth_function/lambda_function.py
basilv: We already capture the user's information in Cognito, so logging it doesn't significantly change the privacy risk. But maybe it should be debug level.

basilv: (Add a ticket in the backlog for enabling debug mode in production.)

MCatherine1994: yeah.... I was not sure how we can quickly enable debug mode in prod, okay I'll change this to debug log, and create a new ticket for enabling debug mode in production, thanks!

server/auth_function/lambda_function.py
ianliuwk1019: Good catch.

ianliuwk1019 (approved):
I see you figure out adding testing on attribute for original tests. Thanks!

Issue #1187, opened 2024-02-06
As a delegated - remove end user access


OlgaLiber2:
**Describe the task**
As a delegated admin I want to revoke access, but only for roles/forest clients I'm authorized to administer

**Acceptance Criteria**
- [ ] Remove access function backend security check working

**Additional context**
- Add any other context about the task here.
- Or here

MCatherine1994:
Put some screenshots here for the manual testing result:

**Login as business bceid user**
- When login as business bceid user and try to grant idir user access, throw error
    ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/5256c925-cfe4-4150-8723-5346d8ec32e7)
- Login as business bceid user with the privilege to grant role "FOM_SUBMITTER_00001011"
    ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/ed512bcc-6a67-43a9-8a6a-14d64932c74d)
    - Grant this access to business bceid user from a different organization, throw error
        ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/16b8e52b-8530-427c-8206-458abd6fef92)
    - Grant this access to business bceid user from same organization, pass
        ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/0b942ac4-a9f2-4881-8033-9941a6b1a58e)
     - Grant access to business bceid user from same organization but with a different forest client number, throw error
        ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/0b59d07b-f1e0-43cb-917e-a8f0c69866ea)

**Login as idir user**
- Login as an IDIR user with the privilege to grant role "SPAR_NONMINISTRY_ORCHARD_00001011" and role "FOM_REVIEWER"
    ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/9d2366ed-f43b-461e-a5dc-8e9c833c3f53)
    - Grant this access to an IDIR user, pass
       ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/2fc351e1-0f72-4af0-b42c-8648558d8775)
    - Grant access with the same parent role but a different forest client number, throw error
        ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/eb942257-b0d3-43c0-83c1-76b162aa6a2a)
    - Grant access with a concrete role, pass
[...]

Issue #1285, opened 2024-04-04
Pass user_guid from frontend to backend


MCatherine1994:
**Describe the task**
We need user_guid to valid the user on backend, and store it. Frontend needs to store the user_guid in the form data, and pass it when making the api call to create user role assignment, create delegated admin and create application admin. 

**Acceptance Criteria**
- [x] Backend update the schema of api `create_user_role_assignment`, `create_delegated_admin` and `create_application_admin` to accept user_guid 
- [x] Frontend store the user_guid in the local form data after verify the user, and pass to the api call when create user role assignment, create delegated admin and create application admin

**Additional context**
- This might need to wait #1183 to be done first, so we can get the user_guid for both IDIR and business BCEID user
- Or here

ianliuwk1019:
May need to verify if we need to increase the existing "user_guid" column and "business_guid" column length to be larger enough.
Currently they are:
"""
user_guid = Column(String(32))
business_guid = Column(String(32))
"""

ianliuwk1019:
I checked our model does not have constraint for user_guid column and currently is already storing it from the attribute "custom:idp_user_id" so it resolved my confusion above, 32 characters should be fine and no uniqueness conflict.

Issue #1270, opened 2024-03-26
Store and update business guid when create the user


MCatherine1994:
**Describe the task**
For business bceid user, when we grant them access or add them as delegated admin, we need to store their business guid. So even they never login to any FAM applications, we can still have their business guid, and will only return their information to the delegated admin from same organization. 

**Acceptance Criteria**
~~- [ ] Update api `create_user_role_assignment`, and `create_delegated_admin` to validate the user by user_guid, and get the business_guid~~
- [x] Update api `create_user_role_assignment`, and `create_delegated_admin` to store business guid for business bceid user when create the user, or update the business_guid if user already exists (business_guid shouldn't be changed? this will be the case that we add business_guid for existing business bceid users in our database which didn't record that in the past)
- [x] Update the router guard where we verify the business_guid to search user by user_guid, not by user_name
- [x] Update tests

**Additional context**
- Need to do #1285 first, to let frontend pass the user_guid to the api calls
- may require a second pass to implement testing fully

MCatherine1994:
Hi @basil, just thinking ahead, later we'll work on another ticket to store user's guid as well when create the user, then feel it will be simpler to still do a search (validation) by user name and type on backend, and grab both user guid and user business guid (if has), and store them. And no need to pass business guid or guid from frontend anymore. How you feel about this?

basilv:
Well, later we might have the user be able to do a search by user's first/last name on front-end and pick a specific account. Also, since the user name can be changed at any time, the safest option is to use user guid to pass to backend.

MCatherine1994:
> Well, later we might have the user be able to do a search by user's first/last name on front-end and pick a specific account. Also, since the user name can be changed at any time, the safest option is to use user guid to pass to backend.

oh yeah... I agree, that's good point, thanks!!

ianliuwk1019:
Also create a ticket to improve backend tests: #1358 .

PR #1299, opened 2024-04-11
feat: #1187 add security check for delegated admin remove access

MCatherine1994:
refs: #1187

- update the schema for create/update user to char 100 (missed this last time)
- update the char limit of the role purpose column to 300, cause for child role, the role purpose will be `parent's role purpose + forest client number`. For spar roles if we grant access with the abstract spar role and a forest client number, it will exceed the current limit 200, and throw error
- add router guard for checking privilege
- add router guard to prevent business bceid user manage idir user access
- add router guard to ensure business bceid user can only manage user from same organization

Will create a separate pr for adding tests


server/admin_management/api/app/schemas.py
```diff
@@ -72,7 +72,7 @@ class FamUserBase(BaseModel):
 
 class FamUserDto(FamUserBase):
     user_type_code: famConstants.UserType
-    create_user: Annotated[str, StringConstraints(max_length=60)]
+    create_user: Annotated[str, StringConstraints(max_length=100)]
```
MCatherine1994, line 75: this was left from previous tickets, just noticed them when doing test, the create/update user in database is having the limit 100, updated the schemas

server/backend/api/app/routers/router_guards.py
```diff
@@ -184,31 +202,69 @@ def authorize_by_application_role(
...
+                forest_client_number
+                and role.role_type_code == RoleType.ROLE_TYPE_ABSTRACT
+            ):
+                parent_role = role
+                forest_client_role_name = (
+                    crud_user_role.construct_forest_client_role_name(
+                        parent_role.role_name, forest_client_number
+                    )
+                )
+                role = crud_role.get_role_by_role_name_and_app_id(
```
MCatherine1994, line 249: I put the comment in the code, also want to mention here, so for granting access, it can be granting an abstract role with a forest client number, for example, the request data could be as the following. But when we check for privilege, we want to check if the requester has the privilege to grant the child role, which is "FOM_SUBMITTER_00001011", not the parent role "FOM_SUBMITTER". The code above is just to grab the child role for this special case. 
``` 
{
user_name: 'IANLIU',
user_type_code: 'I',
role_id: 3,
forest_client_number: '00001011'
}
```

server/flyway/sql/V47__update_role_purpose_length.sql
```diff
@@ -0,0 +1,4 @@
+-- Increase length of role purpose column to 300
+
+ALTER TABLE app_fam.fam_role
+    ALTER COLUMN role_purpose SET DATA TYPE VARCHAR(300);
```
MCatherine1994, line 4: found an error while doing the testing, so when grant access with the SPAR role `SPAR_NONMINISTRY_ORCHARD` and a forest client number `00001011`, the role purpose for the new child role `SPAR_NONMINISTRY_ORCHARD_00001011` will be "the role purpose of SPAR_NONMINISTRY_ORCHARD" + "for 00001011", longer than 200 chars. I'm not sure if 500 will be too long? feel 300 is safe for now

server/backend/api/app/routers/router_guards.py
```diff
@@ -273,3 +370,84 @@ async def enforce_self_grant_guard(
...
+            # target_user found, expect it to have business_guid
+            target_user_business_guid = target_user.business_guid
+        else:
+            # target_user does not exist in FAM database (new user)
+            # search the target user from IDIM proxy and get business_guid
+            # todo: update to search by user_guid instead of user_name
+            try:
+                rbody = await request.json()
+                target_user_business_guid = get_business_guid(
+                    requester, rbody["user_name"]
```
MCatherine1994, line 406: please note that, currently we're not passing user_guid from frontend, we can only verify the user by username. We have a new ticket #1285 to pass the user_guid to backend. And we have ticket another ticket #1270 for storing business guid, so added this todo to #1270, and we can update to use user_guid to do the search in that ticket.

server/backend/api/app/routers/router_guards.py
```diff
@@ -273,3 +370,84 @@ async def enforce_self_grant_guard(
                 },
                 headers={"WWW-Authenticate": "Bearer"},
             )
+
+
+async def enforce_bceid_by_same_org_guard(
+    request: Request,
+    # forbid business bceid user (requester) manage idir user's access
+    _enforce_user_type_auth: None = Depends(authorize_by_user_type),
```
MCatherine1994, line 378: if for the case business bceid user manage idir user access, will raise exception here. So for this call, it will only check the case when business bceid manage business bceid

server/backend/api/app/crud/crud_access_control_privilege.py
```diff
@@ -45,3 +45,23 @@ def get_delegated_admin_by_user_id(
         )
         .all()
     )
+
+
+def get_delegated_admin_by_user_id_and_role_id(
```
basilv, line 50: I'm not keen on how this is returning a random item (if there are multiple matches). Would this make more sense as is_delegated_admin_... returning a boolean?

basilv: Also, why are we checking this for a specific role? I'm missing the use case for this check right now.

server/backend/api/app/routers/router_guards.py
```diff
@@ -153,26 +163,33 @@ async def get_request_role_from_id(
     Some endpoints has path_params with "user_role_xref_id".
     Some endpoints has role_id in request body.
     """
-    user_role_xref_id = None
+    role = None
+
     if "user_role_xref_id" in request.path_params:
         user_role_xref_id = request.path_params["user_role_xref_id"]
```
basilv, line 169: This whole section looks more complicated than I like, and makes certain assumptions about request parameters supplied on the routes this is guarding. I think at a minimum this needs better commenting of what the method is doing.

server/backend/api/app/crud/crud_access_control_privilege.py
ianliuwk1019: Yes, Basil's suggestion is really good.
Since all these are used for `router_guards` to determine delegated admin or not, and not directly needed for the object return, with boolean type of "**is_delegated_admin**()" make a lot sense, but not sure if Python support method overloadding... (probably not).
In the sqlalchemy query, something like `....filter().scalar() is not None` might do the work.

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -18,11 +21,16 @@
...
+@router.post(
+    "",
     response_model=schemas.FamUserRoleAssignmentGet,
     # Guarding endpoint with Depends().
     dependencies=[
-        Depends(authorize_by_application_role),
+        Depends(authorize_by_application_role),  # requester needs to be app admin or delegated admin
+        Depends(authorize_by_privilege),  # if requester is delegated admin, needs to have privilge to grant access with the request role
+        Depends(authorize_by_user_type),  # check business bceid user cannot grant idir user access
+        Depends(enforce_bceid_by_same_org_guard),   # check business bceid user can only grant access for the user from same organization
```
ianliuwk1019, line 33: Not sure the `dependencies` for the order of each `Depends`, it could matter, I can't seem to find the doc mentioned this.
I hope it is the **natural ordering** of the Array (left to right). So, "enforce_self_grant_guard" might be the first to guard and comes the `authorize_by_application_role`, `authorize_by_user_type`, `authorize_by_privilege` then `enforce_bceid_by_same_org_guard` I guess.
I was wondering that, but if manual testing work, maybe in this case it does not have issue.

server/backend/api/app/routers/router_guards.py
```diff
@@ -1,10 +1,11 @@
+import array
```
ianliuwk1019, line 1: Not used I think.

server/backend/api/app/routers/router_guards.py
```diff
@@ -153,26 +163,33 @@ async def get_request_role_from_id(
...
+        role = user_role.role
     else:
         try:
             rbody = await request.json()
             role_id = rbody["role_id"]
             LOGGER.debug(f"Retrieving role by Request's role_id: {role_id}")
-            role = crud_role.get_role(db, role_id)
-            return role  # role could be None.
+            role = crud_role.get_role(db, role_id)  # role could be None.
+        except json.JSONDecodeError:  # When request does not contains body part.
```
ianliuwk1019, line 179: I can't recall since when we started using `try - except json.JSONDecodeError`...
I think this `try-except` is not needed.
The `Depends` method "assume" to get the id from two sources: The "**path_params**" and the json "**body**", then they should be there for the endpoint guard. If the body or path_params are malformed then this is not expected and just let the function raise/throw the original exception.

server/backend/api/app/routers/router_guards.py
```diff
@@ -246,6 +301,48 @@ async def get_target_user_from_id(
...
+        if target_user:
+            # in the case of granting/removing access to an existing user
+            target_user_type_code = target_user.user_type_code
+        else:
+            # in the case of granting access to a new user
+            try:
+                rbody = await request.json()
+                target_user_type_code = rbody["user_type_code"]
+            except json.JSONDecodeError:
+                target_user_type_code = None
```
ianliuwk1019, line 323: Same as previous `try-except JSONDecodeError` comment. 
We could just safely assume the request body can be decoded safely, and you are able to retrieve it from the "user_type_code" attribute or None. If it cannot decode the request "body" then something is going wrong, we don't handle it.

server/backend/api/app/routers/router_guards.py
```diff
@@ -273,3 +370,84 @@ async def enforce_self_grant_guard(
...
+            # search the target user from IDIM proxy and get business_guid
+            # todo: update to search by user_guid instead of user_name
+            try:
+                rbody = await request.json()
+                target_user_business_guid = get_business_guid(
+                    requester, rbody["user_name"]
+                )
+
+            except json.JSONDecodeError:
+                target_user_business_guid = None
```
ianliuwk1019, line 410: Same as previous try-except JSONDecodeError comment.
We could just safely assume the request body can be decoded safely, and we are able to retrieve value from the "user_name" attribute.

server/backend/api/app/routers/router_user_role_assignment.py
MCatherine1994: I think I tried before, the dependencies is in order, like it will return permission error first, and then the user type error

server/backend/api/app/routers/router_guards.py
ianliuwk1019: Thanks Catherine for implement this guard. I admit I was confused. This guard method is probably most difficult so it appears is quite complicated for now.
If it works (I believe), we probably later need to refactor to abstract some logic out for more clarity for future developers.

server/backend/api/app/routers/router_guards.py
```diff
@@ -246,6 +301,48 @@ async def get_target_user_from_id(
...
+                target_user_type_code = None
+
+        if not target_user_type_code:
+            error_description = f"{missing_key_attribute_error.detail['description']} Target user user type code is missing."
+            missing_key_attribute_error.detail["description"] = error_description
+            raise missing_key_attribute_error
+
+        if target_user_type_code == UserType.IDIR:
+            raise HTTPException(
+                status_code=HTTPStatus.BAD_REQUEST,
```
ianliuwk1019, line 332: Probably should use `HTTPStatus.FORBIDDEN`?

server/backend/api/app/crud/crud_access_control_privilege.py
```diff
@@ -8,40 +8,62 @@
...
-        db.query(FamAccessControlPrivilege)
+        True
+        if db.query(FamAccessControlPrivilege)
         .join(FamRole)
         .filter(
             FamAccessControlPrivilege.user_id == user_id,
             FamRole.application_id == application_id,
         )
-        .one_or_none()
+        .first()
```
MCatherine1994, line 29: so it will return the first row found or none, we only need to know the user has at least one record, then means the user is the delegated admin of the application

server/backend/api/app/routers/router_guards.py
```diff
@@ -184,31 +196,71 @@ def authorize_by_application_role(
...
+
+        requester_has_privilege = False
+        if role:
+            requester_has_privilege = (
+                crud_access_control_privilege.has_privilege_by_role_id(
+                    db, requester.user_id, role.role_id
+                )
+            )
+        # if user has no privilege of the role, throw permission error
+        if not requester_has_privilege:
```
ianliuwk1019, line 255: Just curious, maybe I am still confused.
Does this need to be `if (role is None) or not requester_has_privilege:`?

MCatherine1994: um... yeah, but feel no need to add role is None here? cause if role is None, we won't be able to get role_id, so the requester_has_privilege will be None?
Screen Shot 2024-04-11 at 4 03 36 PM
or it can be like the following, do you prefer this way maybe?
```
if not role or not crud_access_control_privilege.has_privilege_by_role_id(
    db, requester.user_id, role.role_id
):
```

server/backend/api/app/routers/router_guards.py
```diff
@@ -273,3 +361,78 @@ async def enforce_self_grant_guard(
...
+        requester_business_guid = requester.business_guid
+        target_user_business_guid = None
+
+        if target_user:
+            # target_user found, expect it to have business_guid
+            target_user_business_guid = target_user.business_guid
+        else:
+            # target_user does not exist in FAM database (new user)
+            # search the target user from IDIM proxy and get business_guid
+            # todo: update to search by user_guid instead of user_name
```
ianliuwk1019, line 393: It's very minor, "todo" might be better with "TODO" to be tracked by some system I think (like typescript: `[S1135 20](https://rules.sonarsource.com/typescript/RSPEC-1135): Track uses of “TODO” tags`].

server/backend/api/app/routers/router_guards.py
ianliuwk1019: Ah I see, I didn't follow it clearly.

ianliuwk1019 (approved):

Issue #1186, opened 2024-02-06
As a delegated admin, grant access to users within my organization, for Backend API


OlgaLiber2:
**Describe the task**
As a delegated admin I want to grant users access to roles for forest clients I'm authorized to administer. 

**Acceptance Criteria**
- [x] Filter list of roles on frontend (filter is done based on the cached access from get my access api)
- [x] Grant access function backend security check:
     - [ ] business bceid user can only grant access to user from the same organization
     - [ ] business bceid user (delegated admin) can only grant user with role they have privilege, the role should already exists (when we create the delegated admin, we already define what roles he can manage)
     - [ ] business bceid can't grant IDIR user access
- [x] Return permission error if not pass the security check
- [x] Add test for the security check, same vs diff organization, able to manage role vs not allow to manage role

**Additional context**
- Only allow access granted to forest clients with status = active.
- do we want to filter forest clients based on my access (obtained from get_my_FAM_access API), do we want to create a forest client multiple select? or ask the user typing in and we check if he's allowed to add that forest client number? (create a new ticket for the design and frontend to handling the error case)

PR #1312, opened 2024-04-16
feat: #1186 add test for delegated admin granting access

MCatherine1994:
refs: #1186 

Add test for delegated admin granting access


ianliuwk1019 (approved):
Awesome tests, and easy to understand. Thank you Catherine!!

Issue #1315, opened 2024-04-17
Application name returned in my access api is not correct


MCatherine1994:
**Describe the task**
Found that some of the application name return by the `admin-user-accesses` api is not correct. Checked the database, application name is correct. Should be `remove_app_env_suffix` method has something not correct when try to remove the environment from application name. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/c0b14294-1661-40a3-9801-42408788b9b5)

**Acceptance Criteria**
- [ ] Fix the application name

**Additional context**
- ![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/949c50a8-62ef-430e-90f6-f027459a537f)
- Or here

PR #1316, opened 2024-04-17
fix: #1315 fix application name error

MCatherine1994:
refs: #1315

It's still unclear to me why rstrip() doesn't work properly, but changed to another way seems work fine now. 
Screen Shot 2024-04-17 at 4 14 32 PM


ianliuwk1019 (approved):
Really strange issue, I am unclear as well.

PR #1317, opened 2024-04-19
fix: adjusting sidenav list for better viewing

NickSaglioni:
Adjusting the sidenav list appearance per request:

Capture-fam-nav-bar2

### DIFF
--- a/frontend/src/assets/styles/default-theme.scss
+++ b/frontend/src/assets/styles/default-theme.scss
@@ -19,6 +19,11 @@ $text-on-color: map.get(darkTheme.$dark-theme, 'text-on-color');
 
 //lightTheme
 $light-layer-selected-01: map.get(lightTheme.$light-theme, 'layer-selected-01');
+$light-layer-hover-01: map.get(lightTheme.$light-theme, 'layer-hover-01');
+$light-layer-selected-hover-01: map.get(
+    lightTheme.$light-theme,
+    'layer-selected-hover-01'
+);
 $light-link-primary: map.get(lightTheme.$light-theme, 'link-primary');
 $light-link-primary-hover: map.get(
     lightTheme.$light-theme,
--- a/frontend/src/components/common/SideNav.vue
+++ b/frontend/src/components/common/SideNav.vue
@@ -1,26 +1,21 @@
 <script setup lang="ts">
 import Sidebar from 'primevue/sidebar';
 import router from '@/router';
-import { IconSize } from '@/enum/IconEnum';
 import { sideNavState } from '@/store/SideNavState';
 import type { PropType } from 'vue';
 import type { RouteLocationRaw } from 'vue-router';
-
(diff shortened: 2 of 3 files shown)


ianliuwk1019:
Hi @NickSaglioni,

General look good. Olga had a brief look during scrum and mentioned this behaviour is a bit strange. Would you be able to adjust it? Thank you.
Please use your judgement if this behaviour should be on common repo or not.
image
reference from Carbon: https://carbondesignsystem.com/components/UI-shell-left-panel/usage/

NickSaglioni:
> Hi @NickSaglioni,
> 
> Generally look good. Olga had a brief look during scrum and mentioned this behaviour is a bit strange. Would you be able to adjust it? Thank you. Please use your judgement if this behaviour should be on common repo or not. image reference from Carbon: https://carbondesignsystem.com/components/UI-shell-left-panel/usage/

Hi @ianliuwk1019. Thanks for pointing it out. I have adjusted as requested. Please let me know if it looks ok. Thank you.

ianliuwk1019:
> > Hi @NickSaglioni,
> > Generally look good. Olga had a brief look during scrum and mentioned this behaviour is a bit strange. Would you be able to adjust it? Thank you. Please use your judgement if this behaviour should be on common repo or not. image reference from Carbon: https://carbondesignsystem.com/components/UI-shell-left-panel/usage/
> 
> Hi @ianliuwk1019. Thanks for pointing it out. I have adjusted as requested. Please let me know if it looks ok. Thank you.

Hi @NickSaglioni , thanks for adjusting.
The hover on the none-selected item seem to also "bold" the text. The Carbon style sample does not do that. Could you adjust it? If common repo does not have specific one for side navigation hover styling, maybe just change it as sideNave.vue. Thanks.

MCatherine1994 (approved):

MCatherine1994:
looks good, thanks Nick! My only concern is that, when there is only the "add user permission" option, the rest `<li>` elements are hidden, however they still exist, not sure from the accessibility perspective for screen reader, should we still render them? will that confuse the user? or it's okay to just hide them like this
Screen Shot 2024-04-22 at 11 58 48 AM

ianliuwk1019:
> looks good, thanks Nick! My only concern is that, when there is only the "add user permission" option, the rest `<li>` elements are hidden, however they still exist, not sure from the accessibility perspective for screen reader, should we still render them? will that confuse the user? or it's okay to just hide them like this Screen Shot 2024-04-22 at 11 58 48 AM

That's a good observation from @MCatherine1994! It is preferable conditional rendering the elements and not just hide the element. Not sure if the reader agent actually take into account if the element is "display: none", probably not.

MCatherine1994:
Will merge this pr, and @NickSaglioni please create a follow up ticket to make adjustment of hidden side navigation options, thanks!!

Issue #1184, opened 2024-02-06
As a delegated admin filter the list of users granted access by organization


OlgaLiber2:
**Describe the task**
As a delegated admin I want to review the list of users granted access for the application, restricted to the roles / forest clients I'm delegated to administer. 

**Acceptance Criteria**
- [x] Update get_user_role_assignment API to get a list of users and filter the result to only see bceid users within my organization
- [x]  Update get_user_role_assignment API to get a list of users and filter the result restricted to the roles/forest-clients that the delegated administer has privilege to grant.
- [x] add test 
- [x] Verify frontend dashboard only display users in the same organization for business bceid delegated admins

**Additional context**
- only applied to business BCeID users/logins
- we might need to ask for test business bceid account; might need to connect with Andrew Schwenker about existing business bceid account that he created
- For testing need at least two different business BCeID users in two different orgs

gormless87:
Further refinement of #1182 will impact this issue; may affect AC

MCatherine1994:
just fyi, I checked today we have 4 working test business accounts, and they are from different organizations.

MCatherine1994:
Got a quick question, wondering if business bceid user can see idir user’s access? Or will only see the business bceid user from the same organization?

basilv:
If they can't add IDIR users, they shouldn't see IDIR users in the table.

OlgaLiber2:
@MCatherine1994  only see users from the same org

ianliuwk1019:
The "Add User Permission" functions allows both granting IDIR users and BCeID users and no restriction for the APP for the delegated admin (as BCeID user), so so far I don't see why the BCeID user is not allowed to see IDIR users assigned to the APP.

MCatherine1994:
okay, then we need another frontend ticket, to hide the "IDIR" option when grant user access, if the login user is business bceid. Bceid user can't verify IDIR user, so from frontend, they can't add IDIR user anyway.

ianliuwk1019:
Add one more Acceptance Criteria for filtering: `Update get_user_role_assignment API to get a list of users and filter the result restricted to the roles/forest-clients that the delegated administer has privilege to grant`.

PR #1319, opened 2024-04-22
feat: #1184 delegated admin filtering same org tests

ianliuwk1019:
ref: #1184 
- Some refactoring on test constants and adjusted affected tests. 
- Added test "test_get_application_user_role_assignments_filtering_for_delegated_admin" for
   viewing application users access grants.


server/backend/testspg/router/test_router_application.py
```diff
@@ -244,4 +257,243 @@ def test_fam_application_endpoints_invlid_path_application_id_type(
         headers=jwt_utils.headers(token)
     )
     assert response.status_code == HTTPStatus.UNPROCESSABLE_ENTITY
-    assert invalid_path_router_msg in response.text
\ No newline at end of file
+    assert invalid_path_router_msg in response.text
+
+
+# ---------------- Test get application user role assignment filtering scenarios ------------ #
+
```
ianliuwk1019, line 264: Sorry @MCatherine1994 , there are lots of constants refactoring.
The real test is just below.
Thank you!!

server/backend/testspg/constants.py
```diff
@@ -42,40 +48,95 @@
...
-# note:
-# test idir and bceid username might need change to a
-# real one after we enable the verfication, same for
-# forest client number
-TEST_USER_ROLE_ASSIGNMENT_FOM_DEV_CONCRETE = {
+# Note:
+# The test idir and bceid username might need change to a real one after we
+#   enable the verfication, same for forest client number.
+#
+# Please refer to V1001__ flyway script for pre-test admin testers setup
```
MCatherine1994, line 60: shall we say under local_sql as well?

server/backend/testspg/conftest.py
```diff
@@ -160,3 +176,46 @@ async def _get_current_requester_by_token(access_token: str) -> Requester:
         return requester
 
     return _get_current_requester_by_token
+
+
+@pytest.fixture(scope="function")
+def create_test_user_role_assignments(
```
MCatherine1994, line 182: just wondering could we combine this one and the one below (I created before)? feel they almost do the same thing, thanks!!

MCatherine1994: oh sorry please ignore me, I see what you're trying to do here, all good

server/backend/testspg/router/test_router_application.py
```diff
@@ -244,4 +257,243 @@ def test_fam_application_endpoints_invlid_path_application_id_type(
...
+
+
+
+
+
+
+
+
+
```
MCatherine1994, line 499: maybe we don't need these many new lines?

ianliuwk1019: Hm.... what's going on.... I didn't remember I purposely hit enter so many times.... but maybe I was almost falling asleep when writing a test.... sorry, will remove that. Thanks for noticing it!!!

server/backend/testspg/jwt_utils.py
```diff
@@ -12,12 +12,18 @@
...
 IDIR_USERNAME = "COGUSTAF"
 
-# for testing purpose,
-# we added this user as FOM DEV delegated admin in local sql,
-# to manage role FOM REVIEWER and FOM SUBMITTER with forest client number 00001018
+# Users setup (key identifier: cognito_user_id) for testing purpose.
+# We added these user "TEST-3-LOAD-CHILD-1"(BCEID) and "PTOLLEST"(IDIR) as
+#   FOM DEV delegated admin in local flyway sql, to manage role for
+#   FOM REVIEWER and FOM SUBMITTER with forest client number 00001018
 COGNITO_USERNAME_BCEID = (
```
MCatherine1994, line 19: Hi Ian, sorry I just feel maybe we can also call them "COGNITO_USERNAME_BCEID_DELEGATED_ADMIN" and COGNITO_USERNAME_IDIR_DELEGATED_ADMIN"? so we know they are delegated admin

server/backend/testspg/constants.py
ianliuwk1019: Added :)

server/backend/testspg/jwt_utils.py
ianliuwk1019: I was not sure about it, 50/50, but no problem, I added now.

server/backend/testspg/router/test_router_application.py
```diff
@@ -244,4 +257,243 @@ def test_fam_application_endpoints_invlid_path_application_id_type(
...
+    test_rsa_key,
+    create_test_user_role_assignments,
+    fom_dev_access_admin_token,
+):
+    """
+    This test first create various user role assigments at FOM_DEV (IDIR/BCEID,
+    business organizations and forest client...) and verify different
+    visibility scenarios based on requester's admin level and
+    delegated admin privileges.
+    """
```
MCatherine1994, line 276: Hi Ian, thanks so much for writing this test, I totally get the complexity. Just thinking of the maintainability and for easy understanding, maybe we could think of separate this one big test case?

Like we can do three different tests:

- Verify fom dev admin can get all user role assignments (we already have the test case `test_get_fam_application_user_role_assignment_concrete_role` and `test_get_fam_application_user_role_assignment_abstract_role`, so could also skip this one)

- Verify IDIR delegate admin can get all user role assignment with the privilege roles
    I got what you're trying to verify here, so basically we check:
    - response data total length 
    - response only has user role assignments with fom_reviewer and fom_submitter_00001018 role
    - response data length of idir user (with abstract role and concrete role)
    - response data length of bceid user (with abstract role and concrete role)
  
    I'm thinking if we can check everything based on one list instead of multiple ones. I found it's a little bit hard to follow when we check the length based on different lists. And also if we can add simple comment of what we're trying to verify for the above 4 cases?
    ```
     access_grants = [
        ACCESS_GRANT_FOM_DEV_CR_IDIR,
        ACCESS_GRANT_FOM_DEV_CR_BCEID_L3T,
        ACCESS_GRANT_FOM_DEV_AR_00001018_IDIR,          
        ACCESS_GRANT_FOM_DEV_AR_00000001_IDIR,
        ACCESS_GRANT_FOM_DEV_AR_00001018_BCEID_L3T
    ]
    ```
- Verify BCEID delegate admin can get all user role assignment with the privilege roles and from same organization
    I'm thinking maybe we can just create two lists as below, and we should still be able to verify
    - response data total length == len(access_grants_able_to_see)
    - response only has user role assignments with fom_reviewer and fom_submitter_00001018 role
    - response data length of idir user (==0)
    - response data length of bceid user == len(access_grants_able_to_see)
    - response data should all come from access_grants_able_to_see list
    ```
     access_grants_able_to_see = [
        ACCESS_GRANT_FOM_DEV_CR_BCEID_L3T,
        ACCESS_GRANT_FOM_DEV_AR_00001018_BCEID_L3T
    ]
    access_grants_not_able_to_see = [
        ACCESS_GRANT_FOM_DEV_CR_IDIR,
        ACCESS_GRANT_FOM_DEV_CR_BCEID_L4T, 
        ACCESS_GRANT_FOM_DEV_AR_00001018_IDIR,
        ACCESS_GRANT_FOM_DEV_AR_00001018_BCEID_L4T,
        ACCESS_GRANT_FOM_DEV_AR_00000001_BCEID_L3T
    ]
    ```

We can touch base tomorrow, this is a complicated one, just thinking if we can make it simpler to understand, thanks so much!!

MCatherine1994: not sure how you like the idea of two lists`access_grants_able_to_see` and `access_grants_not_able_to_see`, maybe for the IDIR case we can also use these two lists, and verify the response data are all from the `access_grants_able_to_see`, maybe that's easier?

ianliuwk1019: No worry, Initially I was thinking about several tests but then decided to put them into one. I was hoping the added comments could be giving reader more understanding, but if it is still not clear for your to read, I can look at your suggestion and give it some changes.

ianliuwk1019: Hi @MCatherine1994 , you have some time?

ianliuwk1019: Hi Catherine, I split the test with some adjustment. Hopefully they are more clear to read now.
Thanks!

server/backend/testspg/constants.py
```diff
@@ -42,40 +49,109 @@
...
+ACCESS_GRANT_FOM_DEV_CR_BCEID_L4T = {
+    "user_name": USER_NAME_BCEID_LOAD_4_TEST,
     "user_type_code": "B",
-    "role_id": TEST_FOM_DEV_SUBMITTER_ROLE_ID,
-    "forest_client_number": CLIENT_NUMBER_EXISTS_ACTIVE_00001018,  # note this is matched with the test delegated admin privilege in local sql
+    "role_id": FOM_DEV_REVIEWER_ROLE_ID,
+}
+
+ACCESS_GRANT_FOM_DEV_AR_00000001_BCEID = {
+    "user_name": "fom_user_test",
```
MCatherine1994, line 122: Hi Ian, since we will add the user validation to backend soon to validate their user_guid, maybe we use a real test bceid user here?

server/backend/testspg/constants.py
```diff
@@ -42,40 +49,109 @@
...
+ACCESS_GRANT_FOM_DEV_CR_BCEID_L3T = {
+    "user_name": USER_NAME_BCEID_LOAD_3_TEST,
     "user_type_code": "B",
-    "role_id": TEST_FOM_DEV_REVIEWER_ROLE_ID,
+    "role_id": FOM_DEV_REVIEWER_ROLE_ID,
 }
-TEST_USER_ROLE_ASSIGNMENT_FOM_DEV_ABSTRACT_BCEID = {
-    "user_name": "LOAD-3-TEST",
+
+ACCESS_GRANT_FOM_TEST_CR_BCEID_L3T = {
```
MCatherine1994, line 109: I'm not sure if we actually used this one

server/backend/testspg/constants.py
```diff
@@ -42,40 +49,109 @@
...
+}
+
+ACCESS_GRANT_FOM_DEV_AR_00001018_BCEID_L4T = {
+    "user_name": USER_NAME_BCEID_LOAD_4_TEST,
+    "user_type_code": "B",
+    "role_id": FOM_DEV_SUBMITTER_ROLE_ID,
+    "forest_client_number": FC_NUMBER_EXISTS_ACTIVE_00001018,
+}
+
+ACCESS_GRANT_FOM_TEST_AR_00001018_BCEID_L4T = {
```
MCatherine1994, line 149: this one as well, not sure if we actually used it

server/backend/testspg/router/test_router_application.py
```diff
@@ -138,8 +156,8 @@ def test_get_fam_application_user_role_assignment_no_role_assignments(
     access_roles_fom_dev_only = ["FOM_DEV_ADMIN"]
 
     # test no user role assignment for the application
-    role_assignment_end_point = endPoint + \
-        f"/{TEST_FOM_DEV_APPLICATION_ID}/user_role_assignment"
+    role_assignment_end_point = end_point + \
+        f"/{FOM_DEV_APPLICATION_ID}/user_role_assignment"
```
MCatherine1994, line 160: oh maybe this endpoint is the "get_application_role_assignment_end_point" ?

server/backend/testspg/router/test_router_application.py
```diff
@@ -153,15 +171,15 @@ def test_get_fam_application_user_role_assignment_concrete_role(
 ):
     access_roles_fom_dev_only = ["FOM_DEV_ADMIN"]
 
-    role_assignment_end_point = endPoint + \
-        f"/{TEST_FOM_DEV_APPLICATION_ID}/user_role_assignment"
+    role_assignment_end_point = end_point + \
+        f"/{FOM_DEV_APPLICATION_ID}/user_role_assignment"
```
MCatherine1994, line 175: this one as well maybe? same as the `get_application_role_assignment_end_point`?

server/backend/testspg/router/test_router_application.py
```diff
@@ -187,15 +205,15 @@ def test_get_fam_application_user_role_assignment_abstract_role(
...
```
MCatherine1994, line 209: think here as well, thanks!!

server/backend/testspg/router/test_router_application.py
```diff
@@ -237,11 +255,313 @@ def test_fam_application_endpoints_invlid_path_application_id_type(
...
+    )
+
+    # Verify this IDIR delegated admin can view "IDIR" users for # of records
+    #   matches what were granted with FOM_REVIEWER and FOM_SUBMITTER_00001018
+    #   role.
+    assert len([data for data in response_data if (
+        data["role"]["application_id"] == FOM_DEV_APPLICATION_ID and
+        data["user"]["user_type_code"] == UserType.IDIR and
+        (data["role"]["role_name"] == ROLE_NAME_FOM_REVIEWER or
+         data["role"]["role_name"] == ROLE_NAME_FOM_SUBMITTER_00001018)
```
MCatherine1994, line 426: I feel we already assert the role name above, maybe here and below we only need to check the the user type code?

MCatherine1994:
I also tried to run the test locally but got these errors not sure why. So I think the 403 and "requester_not_exists" error could because of the miss match of cognito_uer_id.

Screen Shot 2024-04-24 at 4 59 31 PM

server/backend/testspg/jwt_utils.py
```diff
@@ -12,12 +12,18 @@
...
+# We added these user "TEST-3-LOAD-CHILD-1"(BCEID) and "PTOLLEST"(IDIR) as
+#   FOM DEV delegated admin in local flyway sql, to manage role for
+#   FOM REVIEWER and FOM SUBMITTER with forest client number 00001018
+COGNITO_USERNAME_BCEID_DELEGATED_ADMIN = (
+    # "TEST-3-LOAD-CHILD-1"(BCEID)
     "test-bceidbusiness_bda2a1e212244dc2b9f9522057mangledc58bbb@bceidbusiness"
 )
+COGNITO_USERNAME_IDIR_DELEGATED_ADMIN = (
+    # "PTOLLEST"(IDIR)
+    "test-idir_eb65e9d7828d4718aa6f4193c0mangled@idir"
```
MCatherine1994, line 25: Hi Ian, I think this cognito_user_id needs to match what we have for Pete in the `server/flyway/local_sql/sample_V1002__add_local_user_cognito_id.sql` file, otherwise if I have a V1002 on my local, the cognito_user_id won't match.

ianliuwk1019: Hi Catherine, I think "PTOLLEST" does not need to be in V1002 script, you can remove that from your V1002 as it will be created from V1000 script.
However, I did a minor change on the value to be consistent with same_ script as previously I only follow from V1000 script from "COGUSTAF". Please update again. Thanks.
But it does look strange from original sample_ script some of them having "mangled" at the end and some of them having "mangled" in the middle and the length are different.

server/backend/testspg/constants.py
ianliuwk1019: I think this was from previous, I only refactoring the name I think.
Probably does not matter which "user_name" to use, I will it there as it is used by other tests.

ianliuwk1019: On the other hand, I am still thinking, except "Forest Client", we should be able to mock users and should not rely on too much on external service (IDIM) for our testing; but maybe that will add a bit challenges for future testing. Anyway, can discuss that later.

server/backend/testspg/constants.py
ianliuwk1019: Ah~ I created it to be used first and decided not to use on TEST. Removed.

server/backend/testspg/router/test_router_application.py
ianliuwk1019: Hm... I was only refactoring for variable name with inconsistent python snake_case at "endPoint" using VScode build-in automatic refactoring, and wasn't intended to touch other tests to increase the scope; there were already many files for refactoring this time.
But, sure, this can be changed as it located at the same file same as the new tests since I did create one convenient variable for it. I will need to tests them.

MCatherine1994 (approved):

MCatherine1994:
thanks so much Ian!!

Issue #1183, opened 2024-02-06
Write front-end logic for a scenario that BCeID delegated admin is trying to add a BCeID


OlgaLiber2:
**Describe the task**
Write front-end logic for scenarios that the delegated admin is trying to add a BCeID user
**Acceptance Criteria**
- [x] Frontend add logic to require verify bceid user, call Backend API to verify
- [x] In the successful case, display the information, userid, first name, last name, organization
- [x] As an IDIR user I can validate any Business BCeID
- [x] As a BCeID user if I attempt to validate a user in a different organization, I get the error msg, Error message noting "information about the users cannot be displayed because they are not in your organization [organization name]"
- [x] User not found case has same behavior as for searching IDIR
- [x] implemented in UserNameInput component, create a follow up ticket to update/create the test for the UserNameInput component

**Additional context**
- This applies to any scenario where BCeID user details are displayed:

On validation of the id when granting access
On displaying user first & last name for granted user access.

We'll use the BCeID Business GUID attribute to check the organization of each user.

For any record not matching, redact / blank out / hide / remove attributes of personal information (PIA states that "no information about the result will be displayed", but we can at least confirm that the user id is valid). The result returned to the front-end will specify that information about the users cannot be displayed because they are not in the same organization.

NickSaglioni:
Do we have a test BCeID user? @MCatherine1994

MCatherine1994:
yes, I'll send you in the chat

J0taFerreira:
How should the message be displayed when the bceid user tries to verify outside of his organization? @MCatherine1994 @OlgaLiber2 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/126905995/b90b7eae-31f3-4a97-b1ac-2571cae8b896)


![Image](https://github.com/bcgov/nr-forests-access-management/assets/126905995/6123afc3-6705-4c67-a8f9-256dc61ae0b8)

OlgaLiber2:
@J0taFerreira I would do the format of the first screenshot you have without the user name. But I will give you better wording next week. We can change the text in the future, it should be quick.

PR #1350, opened 2024-05-03
feat: #1183 Frontend enable BCeID delegated admin verifying BCeID user

J0taFerreira:
- Frontend is calling Backend API to verify bceid user
- Success case, display userid, first name, last name and organization
- Not Found case behavior remain the same
- IDIR user can validate any Business BCeID
- Display error message if BCeID attempt to validate a user in a different organization


MCatherine1994:
Hi Jota, let's check the sonar cloud issues, there are two new ones other than the test coverage, thanks!

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -28,29 +36,89 @@ const props = defineProps<{
             <p>Verified user information</p>
         </template>
         <template #content>
-            <div class="col" style="margin-left: 2rem">
-                <label for="userId" class="row">Username</label>
-                <span id="userId" name="userId" class="row">
+            <div
+                v-if="!props.cardMgs"
+                class="col"
+                style="margin-left: 2rem"
```
MCatherine1994, line 42: Hi Jota, let's format this file

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -2,11 +2,19 @@
...
-import type { IdimProxyIdirInfo } from 'fam-app-acsctl-api';
+import type { IdimProxyBceidInfo, IdimProxyIdirInfo } from 'fam-app-acsctl-api';
+import type { PropType } from 'vue';
 
-const props = defineProps<{
-    userIdentity: IdimProxyIdirInfo;
-}>(); // Vue3 alternative way for Type the defineProps().
+const props = defineProps({
+    userIdentity: {
+        type: Object as unknown as PropType<IdimProxyBceidInfo>,
```
MCatherine1994, line 10: maybe just `Object as PropType<IdimProxyBceidInfo>`? do we need unknown?

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -2,11 +2,19 @@
...
+        required: true
+    },
+    cardMgs: {
```
MCatherine1994, line 13: maybe let's rename this to errorMsg

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -28,29 +36,89 @@ const props = defineProps<{
...
+                </label>
+                <span
+                    id="lastName"
+                    name="lastName"
+                    class="row"
+                >
+                    {{ props.userIdentity.businessLegalName }}
+                </span>
+            </div>
+            <div
```
MCatherine1994, line 113: feel this logic is a little bit messed up together, user not found and error message shouldn't share the same element. And let's make sure we updated all the ids and fors (the label for organization name as well)


Screen Shot 2024-05-03 at 3 30 12 PM

frontend/src/components/grantaccess/form/UserNameInput.vue
```diff
@@ -16,6 +18,8 @@ const props = defineProps({
     helperText: { type: String },
 });
 
+const PERMISSION_REQUIRED_FOR_OPERATION = 'permission_required_for_operation';
```
MCatherine1994, line 21: could we check if we can import this from 'fam-app-acsctl-api'? if not possible that's okay

frontend/src/components/grantaccess/form/UserNameInput.vue
```diff
@@ -28,20 +32,43 @@ const computedUserId = computed({
...
+                )
+            ).data;
+        } else {
+            verifiedUserIdentity.value = (
+                await AppActlApiService.idirBceidProxyApi.idirSearch(
+                    computedUserId.value
+                )
+            ).data;
+        }
+    } catch (error: any) {
```
MCatherine1994, line 54: this is a good catch, we missed error handling previously. Even now we catch the permission error and customize the error message, but for other type of errors, I didn't see the toast message popup, feel it's not got caught. But we can create a follow up ticket to handle the rest error cases, might need to wait for Olga's design cause it might not make sense to use toast here as well, might need to follow how we display errors for verify forest client numbers. So I think Jota if you can create a follow up ticket for handling other errors that should be okay for now. How you feel Ian @ianliuwk1019 ?

ianliuwk1019: Sorry, maybe I am not clear. 
What were the errors that need to be caught here other than the business rule that BCeID user can only very BCeID user within the same organization?
In the case of not directly related to the business rule here in this input, I would leave it for general handling (however, I seem to remember that was removed long time ago).
If it is for business rule validation, I feel it make sense if Olga wants to display in this page with some message; but if Olga designs to use toast message it can still be acceptable just need to be consistent with other pages..

However, the caught error: "**PERMISSION_REQUIRED_FOR_OPERATION**" seems a bit strange; the component should expect to catch some error related to (400, 409 or 422) in the component I think.

frontend/src/components/grantaccess/UserIdentityCard.vue
J0taFerreira: Typescript was complaining and people on stack overflow suggested that but I tried without now and was ok so I'm changing. Thanks!

frontend/src/components/grantaccess/form/UserNameInput.vue
J0taFerreira: I did not found it on 'fam-app-acsctl-api', but I'll take another look

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -2,11 +2,19 @@
 import Card from 'primevue/card';
 import Icon from '@/components/common/Icon.vue';
 import { IconSize } from '@/enum/IconEnum';
-import type { IdimProxyIdirInfo } from 'fam-app-acsctl-api';
+import type { IdimProxyBceidInfo, IdimProxyIdirInfo } from 'fam-app-acsctl-api';
+import type { PropType } from 'vue';
 
-const props = defineProps<{
-    userIdentity: IdimProxyIdirInfo;
-}>(); // Vue3 alternative way for Type the defineProps().
```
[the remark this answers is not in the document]
J0taFerreira: Hello Ian! I already removed the unknown. Every other component uses the other option to "defineProps" so I wanted to keep the pattern but I can change to this way.

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -28,29 +36,89 @@ const props = defineProps<{
             <p>Verified user information</p>
         </template>
         <template #content>
-            <div class="col" style="margin-left: 2rem">
-                <label for="userId" class="row">Username</label>
-                <span id="userId" name="userId" class="row">
+            <div
+                v-if="!props.cardMgs"
```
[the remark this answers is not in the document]
J0taFerreira: It does work but, makes senses to use "props.userIdentity.found". Thanks, Ian!

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -28,29 +36,89 @@ const props = defineProps<{
...
-                    User does not exist
+            <div
+                class="col"
+                v-if="
+                    props.userIdentity.found &&
+                    props.userIdentity.businessLegalName
+                "
+            >
+                <label
+                    for="lastName"
```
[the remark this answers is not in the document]
J0taFerreira: I noticed that was wrong and already changed it. Thanks!

frontend/src/components/grantaccess/form/UserNameInput.vue
MCatherine1994: yeah the general error is not triggering the toast error message currently, so Jota if you could create a follow up ticket for that, thanks. And for this pr, let's change to check both status code 409 and the detail code, thanks!!

J0taFerreira: Sure, will do!

frontend/src/components/grantaccess/form/UserNameInput.vue
```diff
@@ -67,12 +67,16 @@ const verifyUserId = async () => {
...
+        );
     }
-    if (verifiedUserIdentity.value?.found)
-        emit('setVerifyResult', true, verifiedUserIdentity.value.guid);
 };
 const resetVerifiedUserIdentity = () => {
-    errorMgs.value = ''
+    errorMgs.value = '';
     verifiedUserIdentity.value = null;
     if (props.domain == UserType.I) emit('setVerifyResult', false);
```
ianliuwk1019: Do we only reset this input to "false" when it is IDIR user?
I think I can search BCeID user with "LOAD-3-TEST" verified and then change mind to input "LOAD-2-TEST" and still can submit.

MCatherine1994: Hi Ian, it's got fixed

MCatherine1994:
I'll fix the test and add a few more

frontend/src/components/grantaccess/form/UserNameInput.vue
```diff
@@ -28,25 +32,50 @@ const computedUserId = computed({
     },
 });
 
-const verifiedUserIdentity = ref<IdimProxyIdirInfo | null>(null);
+const errorMgs = ref('');
```
ianliuwk1019, line 35: Hi Jota, do you mean "errorMsg" or you really mean "Mgs"? (no need to change if you mean Mgs)

frontend/src/components/grantaccess/UserIdentityCard.vue
```diff
@@ -40,29 +45,46 @@ const props = defineProps<{
...
+                <label for="organizationName" class="row">
+                    Organization Name
+                </label>
+                <span id="organizationName" name="organizationName" class="row">
+                    {{ props.userIdentity.businessLegalName }}
+                </span>
+            </div>
+            <div
+                class="col-6 d-flex"
+                v-if="!props.userIdentity.found && props.errorMgs == ''"
```
ianliuwk1019, line 70: Curious, for "User does not exist", isn't it just need to check for "!props.userIdentity.found", why need to check "errorMgs"?

frontend/src/components/grantaccess/form/UserNameInput.vue
MCatherine1994: I changed it

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -48,6 +51,69 @@ const idimIdirSearchMock = (isUserFound: boolean): AxiosResponse => {
...
+            },
+            status: 200,
+            statusText: 'Ok',
+            headers: {},
+            config: {
+                headers: {} as AxiosRequestHeaders,
+            },
+        };
+    } else if (isPermissionError) {
+        // todo: investigate how to mock a axios error
```
MCatherine1994, line 77: Hi Jota, I added the test for verifying bceid user, but we need to figure out how to mock an axios error, do you mind take some time to see if could find an example? thanks!!

frontend/src/tests/UsernameInput.spec.ts
```diff
@@ -184,6 +294,38 @@ describe('UserNameInput', () => {
         expect(wrapper.find('#lastName').element.textContent).toContain(
             LASTNAME
         );
+        expect(wrapper.find('#organizationName').element.textContent).toContain(
+            BUSINESSLEGALNAME
+        );
+    });
+
+    it.skip('Should show the user identity card with permission when BCEID user is not allowed to search', async () => {
```
MCatherine1994, line 302: once we figured how to mock the axios error could remove the "skip" and run this test

frontend/src/components/grantaccess/UserIdentityCard.vue
MCatherine1994: Hi Ian, because in the permission error cases, the props.userIdentity.found is also false, but in this case, we don't want to display "User does not exist". We only want to display the permission errors.

ianliuwk1019: I see. Thanks Catherine.
Although it works, the condition still is confusing. I think it is possible, and I would not using <template> in this way, probably will separate it into two <template> with mutually-exclusive condition (like,. when there is "userIdentity" result and the other one with "errorMsg".
Duplicating the columns in this case should be ok and easier to understand.

MCatherine1994: yeah I was thinking of that before as well, but I think the design will change anyway for display error messages, so I feel it's okay for now

MCatherine1994:
just fyi, our pr title will be used in the release note, so let's try to keep a human readable title from now, I'm trying as well, thanks!!

frontend/src/tests/UsernameInput.spec.ts
J0taFerreira: Hello, Catherine, just made some changes on the test that was skipped, it is passing now. Can you take a look when you have the time? Thanks!

MCatherine1994 (approved):

MCatherine1994:
thanks Jota, looks good!!

PR #1356, opened 2024-05-07
fix:  Resolved Typing Errors in Tests to Enable Successful Build on pr #1351

J0taFerreira:
- Fixed typing errors in our test suite that were preventing successful builds


ianliuwk1019 (approved):
Hi Jota, remember next time to use convention for branch naming: "fix/ticket-number-[your_fix].

For now, remember to do a frontend build before merge on the local. We will have ticket to deal with this. Thanks.

Issue #1282, opened 2024-04-03
Frontend hide user domain when bceid user grant access


MCatherine1994:
**Describe the task**
When a delegated admin is business bceid user, and when tries to grant user access, hide the user domain selection, and the user domain will be business bceid. Business bceid delegated admin can only grant access to business bceid user from the same organization. 

**Acceptance Criteria**
- [x] Hide the user domain option on the grant access page when the login user is business bceid
- [x] Set the user domain form data to business bceid
- [x] Check/Update tests

**Additional context**
- at this point we do not have scenarios where a BCeID will add an IDIR user therefore that option should be removed for BCeID delegates

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/1232b4dc-8378-4d6c-a70c-709ad36ff3b0)

ianliuwk1019:
Verified in Dev. Looks good.

PR #1363, opened 2024-05-08
feat: #1282 hide user domain radio button when bceid user is granting access

J0taFerreira:
- Hide user domain radio button option for business bceid user
- default user domain form data is based on their bceid or Idir

ref: #1282


frontend/src/store/FamLoginUserState.ts
```diff
@@ -45,6 +46,15 @@ const state = ref({
 
 // --- getters
 
+const getUserType = () => {
```
MCatherine1994, line 49: Hi Jota, I wonder maybe we rename get to getUserIdpPovider? I feel overall looks good, but this "usertype" makes me confused with the real user type

ianliuwk1019: I do feel confused as well.
From user point of view, it's their user type, but from code it is derived from token claim (the IDP).
Maybe as Catherine suggested, just use "**getUserIdpProvider**".
But would the code below works? (no need to declare extra variable `let userType =`)
```
       // FAM currently supports IDIR/BCeID Business login.
       const idpProvider = state.value.famLoginUser!.idpProvider!.split('-')[1];
	if (IdpProvider.IDIR == idpProvider)
		return IdpProvider.IDIR
	else 
		return IdpProvider.BCEIDBUSINESS
```

J0taFerreira: Hello! I made the name change to "**getUserIdpProvider**", and followed Ian's code suggested as it worked fine.
Thanks!

ianliuwk1019:
Thanks Jota. There is a SonarCloud issue, would you try to see if you can resolve it? Thanks.

J0taFerreira:
> Thanks Jota. There is a SonarCloud issue, would you try to see if you can resolve it? Thanks.

Sure can!

MCatherine1994:
> Thanks Jota. There is a SonarCloud issue, would you try to see if you can resolve it? Thanks.

I checked that issue and think it's from the existing code, not the changes we made in this pr. Maybe Jota you can give a try, but if it's taking more time, I'm also fine with it.

MCatherine1994 (dismissed):

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
             : 1;
     });
 };
+//--------- get my permissions
+const getMyFamAdminPermission = (access: FamAuthGrantDto): IMyPermission[] => {
+    const famGrant = access.grants.find(
+        (grant: FamGrantDetailDto) =>
+            grant.application.name === FAM_APPLICATION_NAME
+    );
+    if (!famGrant) return [];
```
MCatherine1994, line 237: Hi Jota, I think we could assume famGrant should always has value. so just return
```
return [
        {
            application: famGrant!.application.description,
            env: famGrant!.application.env,
            role: 'Admin',
        },
    ];
```

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
...
+const getMyAdminPermission = (): IMyPermission[] => {
+    const myPermissions: IMyPermission[] = [];
+
+    getUserAccess()?.forEach((access: FamAuthGrantDto) => {
+        switch (access.auth_key) {
+            case AdminRoleAuthGroup.FamAdmin:
+                myPermissions.push(...getMyFamAdminPermission(access));
+                break;
+            case AdminRoleAuthGroup.AppAdmin:
+                myPermissions.push(...getMyAppAdminPermission(access));
```
MCatherine1994, line 294: and here maybe no need to use push, you can use "concat" method to combine two arrays

frontend/src/store/FamLoginUserState.ts
```diff
@@ -28,10 +31,10 @@ export interface FamLoginUser {
 }
 
 interface IMyPermission {
-    application: string,
-    env: string,
-    clientId: number,
-    role: string,
+    application: string | null | undefined;
+    env: AppEnv | null | undefined;
```
ianliuwk1019, line 35: Maybe I misunderstand, why do they need "null | undefined"? I thought these two are necessary?

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
             : 1;
     });
 };
+//--------- get my permissions
```
ianliuwk1019, line 231: Very minor, maybe leaving some space above lin 231. For easy to read.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
...
                 });
-            });
+            }
+        });
+    });
+
+    return permissions;
+};
+
+const getMyAdminPermission = (): IMyPermission[] => {
```
ianliuwk1019, line 285: Can we have this method below the last method of the "// --- getters" area (which is for the getters of the state and they are public/visible)?
The intention was to organize them and not being placed all over the place, to be easier to find.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
...
+            role: 'Admin',
+        },
+    ];
+};
 
-        if (item.auth_key === AdminRoleAuthGroup.AppAdmin) {
-            item.grants.forEach((grant) => {
-                myPermissions.push({
-                    role: 'Admin',
+const getMyAppAdminPermission = (access: FamAuthGrantDto): IMyPermission[] => {
```
ianliuwk1019, line 248: For those internal methods: "getMyFamAdminPermission", "getMyAppAdminPermission" and "getMyDelegatedAdminPermission", can we simply placed them above the "// --- export"?
Thanks.

frontend/src/store/FamLoginUserState.ts
```diff
@@ -215,66 +228,93 @@ const getCachedAppRolesForDelegatedAdmin = (
...
-        permission.application = permission.application.replace(/\([^()]*\)/g, '')
-        return permission
+        permission.application = permission.application!.replace(
+            /\([^()]*\)/g,
+            ''
+        );
+        return permission;
     });
 };
+
```
ianliuwk1019, line 312: Hmm... all above changed methods seems to me quite specific for only one component as it takes current cached user accesses and re-mapping them into specific structure needs on display.
Maybe it was discussed before and felt this is the file to add these methods here, otherwise I am not really sure if this state file is really the place for it (as initially the first time looking at the method, "getMyAdminPermission", I was confused for its purpose and object structure.)
I will skip for this. Just a comment.

frontend/src/store/FamLoginUserState.ts
J0taFerreira: Hello! That's a great idea, thanks!

frontend/src/store/FamLoginUserState.ts
J0taFerreira: it will fail build without "**null | undefined**"

MCatherine1994: Seems like it's because the `FamGrantDetailDto` has application description and env type "string | null", and we need to match the type here as well, otherwise it will fail
Screen Shot 2024-05-10 at 9 48 48 AM

MCatherine1994 (approved):

frontend/src/store/FamLoginUserState.ts
MCatherine1994: I agree Ian, we can move it out later

frontend/src/store/FamLoginUserState.ts
ianliuwk1019: I see, looks like its the backend schema. Thanks Catherine!

ianliuwk1019 (approved):

frontend/src/store/FamLoginUserState.ts
J0taFerreira: sounds good!

Issue #1307, opened 2024-04-15
Enable the entry of multiple forest client numbers with or without space after the comma


OlgaLiber2:
**Describe the task**
Currently there is only one correct format to enter multiple forest client numbers: 
00001011,00001012

We need to enable the entry of multiple forest client numbers with or without space after the comma so it is less prone to user error: 
00001011, 00001012

Both ways should be valid. 

**Acceptance Criteria**
- [x] Either entry format (comma followed by a single space or no space) works
- [x] Update/create tests if need

**Additional context**
- Add any other context about the task here.
- Or here

gormless87:
@J0taFerreira  - adding this to sprint 12

PR #1377, opened 2024-05-09
feat: #1307 enable multiple forest client numbers with or without space after comma

J0taFerreira:
- Changed regex to match comma followed by space or no space
- Removed the "Please enter a digit or comma" since the helper text is there.(approved by P.O)

ref: #1307


frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -30,7 +30,9 @@ const forestClientNumberVerifyErrors = ref([] as Array<string>);
 
 const verifyForestClientNumber = async (forestClientNumbers: string) => {
     forestClientNumberVerifyErrors.value = [];
-    let forestNumbers = forestClientNumbers.split(/[ ,]+/);
+    // regex to match commas with optional space
+    let regex = /\s*,\s*|\s+/
```
MCatherine1994, line 34: Hi Jota,  I think this line gives us a security warning in sonar cloud, do you mind take a look? thanks!

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -141,6 +143,7 @@ watch(
                         v-bind="field"
                         class="w-100 custom-height"
                         @input="cleanupForestClientNumberInput()"
+                        @keypress.enter="verifyForestClientNumber(forestClientNumbersInput)"
```
ianliuwk1019, line 146: I don't see we need this. It looks like with malformed input you still allow frontend to hit the backend.

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: Hello, Ian! @OlgaLiber2 suggested that pressing 'enter' to validate would be desirable

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: Hello, Catherine! I'll try to fix it but might need some help

frontend/src/components/grantaccess/form/ForestClientInput.vue
ianliuwk1019: Hi Jota, Both '**click the button**' and '**pressing enter**' are all fine. That's usability consideration.

However, you missed a point, it is about **wrong format of data hitting the backend**.
If we rely all on the backend checks, without frontend validations, then that's fine.

But our frontend can already check if the input entered at least follow certain format (length 8 and all digits) and validates it before sending to the backend.

Initially, without looking at the code, what I see is this:
image
So it means the code still has a bug.

Please see if you can find it.

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -30,7 +30,11 @@ const forestClientNumberVerifyErrors = ref([] as Array<string>);
 
 const verifyForestClientNumber = async (forestClientNumbers: string) => {
     forestClientNumberVerifyErrors.value = [];
-    let forestNumbers = forestClientNumbers.split(/[ ,]+/);
+
+    // split by commas and spaces
+    let forestNumbers = forestClientNumbers
+        .split(',')
+        .flatMap((clientNum) => clientNum.trim().split(' '));
```
ianliuwk1019, line 37: It's good you find a way to avoid using regular expression in this case; but I am not sure I follow why using".flatMap" and split again with (' ')? Wondering, vave you try using just ".map()" with trim()?

J0taFerreira: I think it can be simplified to not use "flatMap", I'll take a look. Thanks!

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -141,6 +143,13 @@ watch(
                         v-bind="field"
                         class="w-100 custom-height"
                         @input="cleanupForestClientNumberInput()"
+                        @keypress.enter="
+                            field.value &&
+                                !errorMessage &&
+                                verifyForestClientNumber(
+                                    forestClientNumbersInput
+                                )
+                        "
```
ianliuwk1019, line 152: Thanks Jota, yes, if no error then the data can be submit to backend.

This is interesting, I was trying to see how "errorMessage" is produced and what's contains in there.
I thought the vee-validate validation schema is in this ForestClientInput.vue but I could not find it. Instead, it is at the utils.ts and I don't know how it validate the FORM from parent component to child components. But that's just I don't have too much experience with vee-validate. It's working but in a hidden way.

* could you align the conditions vertically?
* since this "**forestClientNumbersInput**" is a ref, probably no need to pass it. But that's ok, fine with that.
* In "verifyForestClientNumber()" method, I am not sure the validation for line: 38~42 is for. Are we expecting these number (in array) can still be string Not-A-Number? (I thought the vee-validate schema already checks). Please check if this is unnecessary.
* I believe Vee-validate also can add some validation messages when needed, you guys might want to utilize that next time for displaying error messages; but I don't know how easy it is. Just a suggestion, not for this PR.

ianliuwk1019:
I notice if we have extra space(s) at the beginning or extra space(s) at the end, the button will be disabled.
image
image

These probably are not in the acceptance criteria. Not sure if @J0taFerreira wants to continue improving it or we can leave it for another ticket. Ideally, it should work like this: it does not matter the space(s) are at the beginning, in the middle, or at the end, as long as each forest client number is 8-numeric digits, then consider they are good inputs.
What do you think @MCatherine1994 ?

@OlgaLiber2

OlgaLiber2:
@ianliuwk1019 I agree

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: Hello Ian! I'll align the code. The "**verifyForestClientNumber()**" method check if the array contain only valid numbers, I'm not sure if veeValidate does this kind of comparison but I'll take a look.

J0taFerreira:
> I notice if we have extra space(s) at the beginning or extra space(s) at the end, the button will be disabled. image image
> 
> These probably are not in the acceptance criteria. Not sure if @J0taFerreira wants to continue improving it or we can leave it for another ticket. Ideally, it should work like this: it does not matter the space(s) are at the beginning, in the middle, or at the end, as long as each forest client number is 8-numeric digits, then consider they are good inputs. What do you think @MCatherine1994 ?
> 
> @OlgaLiber2

It makes sense, I'll follow your suggestion

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: VeeValidate does check to if there are only numbers and the button is disabled if not, so I think it can be taken out

ianliuwk1019 (approved):
Thanks Jota. Approve for this.
However, I see that "**verifyForestClientNumber**" method might need to be improved for this case:
* When the IDs previous were added (and already shown in display table below), frontend can ignore these IDs and no need to call backend to validate them again.

Could you create a ticket to fix that next time? Thanks.

image

frontend/src/components/grantaccess/form/ForestClientInput.vue
MCatherine1994: oh Ian, sorry just saw the message, the validation schema is from the form level, so it's passed from GrantAccess and grant delegated admin component 
Screen Shot 2024-05-13 at 3 35 05 PM

MCatherine1994 (approved):

PR #1390, opened 2024-05-14
chore: #1270 Add tests verifing business_guid saved for BCeID user

ianliuwk1019:
- Add fixture "override_target_user_bceid_search" for mocking bceid search parsed return.
- Add test for user "update"
- Add tests for idim_proxy integration when search BCeID user by user_guid.
  However, test are skipped as IDIM proxy currently has not enable this yet.
- Add tests for verifying business_guid is saved for BCeID user when create user role assignment 
  and when create delegated admin user.


server/backend/testspg/crud/test_idim_proxy_service.py
```diff
@@ -22,25 +26,34 @@ class TestIdimProxyServiceClass(object):
...
+    # Valid test Business Bceid user.
     search_params_business_bceid_diff_org = IdimProxyBceidSearchParam(
         **{"searchUserBy": IdimSearchUserParamType.USER_ID,
-            "searchValue": TEST_VALID_BUSINESS_BCEID_USERNAME_TWO}
-    )  # Valid test Business Bceid user.
+            "searchValue": USER_NAME_BCEID_LOAD_2_TEST}
+    )
+    # IDIR test user_guid. Note, do not use it in other place.
+    # This is only for real integration test.
+    TEST_IDIR_USER_GUID = "E72A12C916A44A9581CF39E5DCDFFAE7"
```
ianliuwk1019, line 45: Hi @basilv , appreciate for your opinion in this case.
This test file is for real integration test for FAM connecting to IDIM-Proxy for IDIM search.
To do the real search, we need to use the **real** **USER_GUID** (for searching on BCeID user, based on IDIM requirement).
We previously already use **real** IDIR "**USER_NAME**" in other search IDIM search.
Do you feel it is ok here for Integration Test or should we have ticket to migrate these to Github secrets (to be feed on to the environment variables)?

basilv: I'd feel more comfortable if this was a github secret.

ianliuwk1019: Thanks Basil. Same, I felt not comfortable as well initially. Will move it to github secret.

server/admin_management/tests/routers/test_router_access_control_privilege.py
```diff
@@ -171,6 +173,60 @@ def test_get_access_control_privileges_by_application_id(
     )
 
 
+def test_assign_new_bceid_user_delegated_admin_save_business_guid(
```
MCatherine1994, line 176: Hi Ian, this is small, but just thinking of our naming consistent for our tests, maybe we could rename it to "test_create_access_control_privilege_for_bceid_user_should_save_business_guid" ?  Just feel normally we name our test case based on "test_[method name]_[conditions]"?

MCatherine1994:
Hi Ian, also wondering shall we rename our pr title?

server/admin_management/tests/routers/test_router_access_control_privilege.py
ianliuwk1019: Yes, I know we normally did that, but I was struggling that will have long name on test method so instead just give a name that's meaningful.
It's sometime even harder to name the condition in a short name. But for sure, I change it if you feel we should stick to that.

server/admin_management/tests/conftest.py
```diff
@@ -261,6 +262,20 @@ def __setup_new_fom_delegated_admin(
...
 
+@pytest.fixture(scope="function")
+def override_target_user_bceid_search(
+    test_client_fixture
+):
+    def _override_target_user_bceid_search(
+        mocked_data
+    ):
+        app = test_client_fixture.app
+        app.dependency_overrides[target_user_bceid_search] = (
```
MCatherine1994, line 273: thanks Ian! I like this idea!

MCatherine1994 (approved):

Issue #1385, opened 2024-05-14
Forest Client Numbers Verification Ignore Already Verified Client IDs


J0taFerreira:
**Task Description**
When the IDs previous were added (and already shown in display table below), frontend can ignore these IDs and no need to call backend to validate them again.

**Acceptance Criteria**
- [x] If the Client ID already exists in the Forest Client Card, verification won't trigger another backend validation for that Client ID.
- [x] Verify if any tests have been broken and update if needed.

**Additional context**
![Image](https://github.com/bcgov/nr-forests-access-management/assets/126905995/739c64eb-19d5-4445-b94d-f059f3d7ebd4)

PR #1393, opened 2024-05-16
fix: #1385 Forest Client Numbers Verification Ignore Already Verified Client Numbers

J0taFerreira:
- already verified ForestNumbers added to ForestClientCard will not call backend again

ref: #1385


frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -27,14 +27,22 @@ const emit = defineEmits([
 const forestClientNumbersInput = ref('');
 const forestClientData = ref<FamForestClient[]>([]);
 const forestClientNumberVerifyErrors = ref([] as Array<string>);
+const verifiedNumbers: Set<string> = new Set();
```
MCatherine1994, line 30: Hi Jota, we already has a method `isForestClientNumberNotAdded` in this component, so I think you just need to move the circled part in the screenshot to the place where before calling the api
image

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -27,14 +27,22 @@ const emit = defineEmits([
...
 
 const verifyForestClientNumber = async (forestClientNumbers: string) => {
     forestClientNumberVerifyErrors.value = [];
 
     // split by commas and spaces
-    let forestNumbers = forestClientNumbers.split(',').map(num => num.trim());
+    let forestNumbers = forestClientNumbers.split(',').map((num) => num.trim());
 
     for (const item of forestNumbers) {
```
MCatherine1994, line 38: and let's rename this "item", I think this is "forestClientNumber", type string?

MCatherine1994:
this should be a fix, not a feat, "fix: #1385 Fix forest client numbers verification to ignore already verified client numbers"

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: hello! I'll make the change

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: sure thing!

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -32,15 +32,22 @@ const verifyForestClientNumber = async (forestClientNumbers: string) => {
     forestClientNumberVerifyErrors.value = [];
 
     // split by commas and spaces
-    let forestNumbers = forestClientNumbers.split(',').map(num => num.trim());
+    let forestNumbers = forestClientNumbers.split(',').map((num) => num.trim());
 
-    for (const item of forestNumbers) {
+    for (const forestClientNumber of forestNumbers) {
+        if (!isForestClientNumberNotAdded(forestClientNumber)) {
```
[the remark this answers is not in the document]
J0taFerreira: We sure can! I'll make the change

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -32,15 +32,22 @@ const verifyForestClientNumber = async (forestClientNumbers: string) => {
...
     // split by commas and spaces
-    let forestNumbers = forestClientNumbers.split(',').map(num => num.trim());
+    let forestNumbers = forestClientNumbers.split(',').map((num) => num.trim());
 
-    for (const item of forestNumbers) {
+    for (const forestClientNumber of forestNumbers) {
+        if (!isForestClientNumberAdded(forestClientNumber)) {
+            forestClientNumberVerifyErrors.value.push(
+                `Client ID ${forestClientNumber} has already been added.`
+            );
```
ianliuwk1019, line 41: Jota, not just rename the method, got to be careful.
Did you check again for "if" that makes sense to you for what's being executed below?

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/form/ForestClientInput.vue
```diff
@@ -83,7 +81,7 @@ const verifyForestClientNumber = async (forestClientNumbers: string) => {
     forestClientNumbersInput.value = forestNumbers.toString();
 };
 
-const isForestClientNumberNotAdded = (forestClientNumber: string) => {
+const isForestClientNumberAdded = (forestClientNumber: string) => {
```
MCatherine1994, line 84: Hi Jota, I think the logic feels weird... so if we rename it to "isForestClientNumberAdded", then I feel it should return true when found the record. But in this case, the code logic returns false when found. So I think if we rename it, we want to remove the "!" from the code, and it will return true when found. So on line 38, it should be 
`if (isForestClientNumberAdded(forestClientNumber))`, if forest client is added, we print error message. Currently it's confused.

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: sorry about that, at the time i did not thought it would be hard to understand.

frontend/src/components/grantaccess/form/ForestClientInput.vue
J0taFerreira: Changed the logic, sorry about that! I did not thought that would make hard to understand

ianliuwk1019 (approved):

Issue #1323, opened 2024-04-23
Run frontend build in sonar pipeline


MCatherine1994:
**Describe the task**
We saw multiple times in the past that the frontend build failed by small errors, but we can only notice that when try to merge to dev. So add the command to run frontend build in the sonar pipeline, we will be able to verify that frontend build is successful when open the pr. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/e1b1c814-2ad5-42b6-a2cc-b6d217761c2a)

**Acceptance Criteria**
- [x] Add `npm run build` to sonar ci-frontend
- [x] Make sure the pipeline works as expected

**Additional context**
- Add any other context about the task here.
- Or here

PR #1394, opened 2024-05-16
feat: #1323 Run frontend build in sonar pipeline

J0taFerreira:
- sonar pipeline will run npm run build


MCatherine1994:
I'm not so sure what's the trigger to let it running, but seems like the we skip the test and build. I'll ask Derek tomorrow about this. 
Screen Shot 2024-05-16 at 1 42 47 PM

ianliuwk1019:
> I'm not so sure what's the trigger to let it run, but seems like that we skip the test and build. I'll ask Derek tomorrow about this. Screen Shot 2024-05-16 at 1 42 47 PM

It seems to be good now.
I added a dummy change to "frontend".
Derek has the trigger defined as:
image
@MCatherine1994 

Probably ok to be approved. However, we could improve to build first then run test-coverage. Otherwise if test run finished but build failed, it is a waste of time to test and it will not show up on Sonar coverage I think.

MCatherine1994:
thanks Ian! I agree, let's update to run "build" first and then test

MCatherine1994 (approved):

Issue #1227, opened 2024-03-08
Cleanup FAM APIs, remove the ones not in use


MCatherine1994:
**Describe the task**
Since we switched to use the get_admin_user_access API to get the access of the login user, and use the cached result for application and role selection option, we can remove the APIs that are not in use anymore. Have to check what exactly the api name.
- the api to get all applications (this was previously used in the grant app admin form, for the application dropdown)
- the api to get roles by application id (this was previously used in the grant user access form, for the role dropdown)

**Acceptance Criteria**
- [ ] deprecated endpoints are removed
- [ ] client code regenerated
- [ ] associated tests removed

**Additional context**
-

PR #1403, opened 2024-05-21
fix: #1227 Cleanup unused apis in admin management and app access management directories

MCatherine1994:


frontend/src/services/ApiServiceFactory.ts
```diff
@@ -24,7 +23,6 @@ type AppAccessControlApiType = {
 
 type AdminManagementApiType = {
     applicationAdminApi: FAMApplicationAdminApi;
-    applicationsApi: FAMApplicationsApi;
```
MCatherine1994, line 27: removed it, as we no longer use it anymore

server/admin_management/api/app/main.py
```diff
@@ -80,11 +80,6 @@ def custom_generate_unique_id(route: APIRouter):
     prefix=apiPrefix + "/application_admins",
     tags=["FAM Application Admin"],
 )
-app.include_router(
-    router_application.router,
-    prefix=apiPrefix + "/applications",
```
MCatherine1994, line 85: removed the applications router, as we no longer use it anymore. Previously we use this endpoint to get all the applications for FAM admin (when grant app admin), now this information is included in the return of that get my access api

server/backend/api/app/crud/crud_application.py
```diff
@@ -11,55 +11,6 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def get_applications(db: Session):
```
MCatherine1994, line 14: removed the methods that are not in use anymore

server/backend/api/app/schemas.py
```diff
@@ -10,35 +10,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class FamGroupPost(BaseModel):
```
MCatherine1994, line 13: removed the schemas not are not in use anymore, also regroup the schemas a little bit

server/backend/testspg/authorization/test_jwt_validation.py
```diff
@@ -20,33 +17,27 @@
...
 
 LOGGER = logging.getLogger(__name__)
-endPoint = f"{apiPrefix}/fam_applications"
+endPoint = f"{apiPrefix}/fam_applications/{FAM_APPLICATION_ID}/user_role_assignment"
 
 
-def test_get_application_success(
-    test_client_fixture_unit: starlette.testclient.TestClient,
+def test_get_application_user_role_assignment_success(
+    test_client_fixture: starlette.testclient.TestClient,
```
MCatherine1994, line 28: we don't have the fam_applications endpoint anymore, so use the get the user role assignment endpoint for testing instead. And we need to use the real db to check user's permission, cannot use the mocked db anymore.

server/backend/testspg/authorization/test_jwt_validation.py
```diff
@@ -20,33 +17,27 @@
...
-    response = test_client_fixture_unit.get(f"{endPoint}", headers=headers(token))
+    response = test_client_fixture.get(f"{endPoint}", headers=headers(token))
 
     assert response.status_code == 200
     data = response.json()
     assert data == []
 
 
-def test_get_application_no_authentication_failure(
+def test_get_application_user_role_assignment_no_authentication_failure(
```
MCatherine1994, line 40: just rename the tests so it makes sense, now we're using the get user role assignment endpoint for testing

client-code-gen/gen/admin-management-api/model/api-app-constants-user-type1.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * 
+ * @export
+ * @enum {string}
+ */
+
+export const ApiAppConstantsUserType1 = {
+    I: 'I',
+    B: 'B'
+} as const;
```
MCatherine1994, line 26: Hi Ian, I'm not sure where this is coming from...

client-code-gen/gen/admin-management-api/model/user-type-output.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * 
+ * @export
+ * @enum {string}
+ */
+
+export const UserTypeOutput = {
+    I: 'I',
+    B: 'B'
+} as const;
```
MCatherine1994, line 26: also not sure where is this coming from...

frontend/src/components/grantaccess/form/RoleSelect.vue
```diff
@@ -2,13 +2,12 @@
...
 import { ErrorMessage, Field } from 'vee-validate';
 import Dropdown from 'primevue/dropdown';
-import type { FamApplicationRole } from 'fam-app-acsctl-api';
 import type { FamRoleDto } from 'fam-admin-mgmt-api/model';
 
 const props = defineProps({
     roleId: { type: Number, default: 0 },
     roleOptions: {
-        type: [Array<FamApplicationRole>, Array<FamRoleDto>],
+        type: [Array<FamRoleDto>],
```
MCatherine1994, line 10: Check the return of the get_admin_user_access api, the role type is FamRoleDto, so the FamApplicationRole should be the previous type

Screen Shot 2024-05-21 at 2 09 33 PM

client-code-gen/gen/admin-management-api/model/api-app-constants-user-type1.ts
ianliuwk1019: This looks curious.
I assume nothing changed other than openapi.json.
I may need to run generation my end to see if I get the same result....
Did you notice if that causes frontend issue or changes needed by this strange generation?

MCatherine1994: no... I reinstalled frontend, and run frontend, looks good...

frontend/src/services/ApiServiceFactory.ts
ianliuwk1019: Thanks Catherine!
Would you mind also remove the comment on line 11, but maybe we can still keep the "alias".

server/backend/testspg/authorization/test_jwt_validation.py
ianliuwk1019: The test on "jwt_validation" previously uses "/fam_application" as endpoint to test on.
I think it does not matter the endpoints, the purpose was just to use any simple endpoint to check on jwt token parsing does work from the "jwt_validation.py".
Changing to "{FAM_APPLICATION_ID}/user_role_assignment" I think still can use the "_unit" test client, just need to provide different mocks; could be difficult in this case, with also the Depends() mock. 

However, I am wondering why switching using the "test_client_fixture", why this assert works "assert data == []"? I thought it should contains something?

MCatherine1994: yeah, because currently we're using the get user role assignment endpoint, and it will check the permission of the requester, however if we mock the database, it won't find our requester from the database. But you're right, why the data is []... let me check...

MCatherine1994: oh Ian, because all admin records are in the application admin table, for FAM it doesn't have any regular user access

ianliuwk1019: I see, thanks for clarification.

client-code-gen/gen/admin-management-api/model/api-app-constants-user-type1.ts
ianliuwk1019: I tried on my end to run generation.
Same to me that it created the strange type.
This is quite strange....

frontend/src/services/ApiServiceFactory.ts
MCatherine1994: sure, I'll do that

client-code-gen/gen/admin-management-api/model/api-app-constants-user-type1.ts
ianliuwk1019: I tried several things, including reinstall python .venv but no difference.
It only happens in server/admin_management, not in server/backend for strange reason unknown.
Sorry, I couldn't tell the difference. Since frontend still runs fine (but swagger is having strange type), maybe we ignore this one for now but keep eyes on it for the next time.

ianliuwk1019 (approved):

Issue #1353, opened 2024-05-06
Update tests for UserNameInput with BCeID functionality


J0taFerreira:
Update tests for the BCeID functionality .

**Acceptance Criteria**
- [x] test if BCeID user can verify other BCeID users
- [x] test if the success message will display correctly when BCeID verify user
- [x] test if the error message will be shown in case that BCeID user tries to verify user from another organization
- [x] test the not found case

**Additional context**
- (https://github.com/orgs/bcgov/projects/65/views/1?pane=issue&itemId=52418494)

PR #1407, opened 2024-05-22
feat: #1353 update tests for user name input with bceid functionality

J0taFerreira:
- tested BCeID user can verify other BCeID users
- tested the success message will display correctly when BCeID verify user
- tested if the error message will be shown in case that BCeID user tries to verify user from another organization
- tested the not found case


frontend/src/tests/UserNameInput.spec.ts
```diff
@@ -179,12 +171,83 @@ describe('UserNameInput', () => {
...
+
+        const cardEl = wrapper.find('.custom-card').element as HTMLSpanElement;
+        expect(cardEl.textContent).toContain('Username');
+        expect(wrapper.find('#userId').element.textContent).toContain(USER_ID);
+        expect(wrapper.find('#userNotExist').element.textContent).toContain(
+            'User does not exist'
+        );
+
+        // triggers username and domain change to BCeID
+        await wrapper.setProps({ domain: newProps.domain, userId: USER_ID });
```
MCatherine1994, line 207: Hi Jota, sorry I forgot that our test file already have the test cases for searching BCeID, and thanks so much for adding more. I wonder if we could rename this newProps to "bceidProps"? and let's create a variable for the "User doest not exist"

frontend/src/tests/UserNameInput.spec.ts
```diff
@@ -299,92 +363,89 @@ describe('UserNameInput', () => {
...
+            LAST_NAME
         );
         expect(wrapper.find('#organizationName').element.textContent).toContain(
-            BUSINESSLEGALNAME
+            BUSINESS_LEGAL_NAME
         );
     });
 
-    it('Should show the user identity card with permission when BCEID user is not allowed to search', async () => {
+    it('should BCeID user able to verify other BCeID users and display info on card', async () => {
```
MCatherine1994, line 378: Hi Jota, I wonder if this case this is same as the one above it?

frontend/src/tests/UserNameInput.spec.ts
J0taFerreira: Hello! Sure can!

frontend/src/tests/UserNameInput.spec.ts
J0taFerreira: they are very similar, one spy on '`bceidSearch`' the other on '`idirSearch`' I think it can be put both on the same test case to be more concise.

MCatherine1994 (approved):

Issue #1395, opened 2024-05-16
Frontend is missing error messages


MCatherine1994:
**Describe the task**
**Case one:** If I'm the delegated admin that can only grant FOM_SUBMITTER with 00001011, and Ian already has the role FOM_SUBMITTER_0001011. When I tried to grant Ian access for FOM_SUBMITTER with 00001011, 00001012, 00001013. It returns me two types of errors, 409 for Conflicts, and 403 for Forbidden, because I don't have privilege to grant access with 00001012 and 00001013. But currently we only display one error message, which is incorrect. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/a64337dd-811d-4dcf-bddc-f12818b3013d)

**Case two:** If I'm the delegated admin that can only grant FOM_SUBMITTER with 00001011, and I'm trying to grant Ian access for FOM_SUBMITTER with 00001011, 00001012, 00001013. It will return success for 00001011, and 403 Forbidden error for 00001012 and 0001013 because I don't have privilege. We suppose to have one success notification and one error notification, but currently we only show the error notification.

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/22d7790f-9f11-4819-848e-9a8251408aa9)

 

**Acceptance Criteria**
- [x] Fix to allow display both success notification and error notification

**Additional context**
- We'll work on the design for case one, so delegated admin can only manage the forest client numbers they're allowed to, to avoid get those kind of errors, will leave it as for now. 
- Or here

PR #1408, opened 2024-05-22
fix: #1395 Frontend is not showing success notification and error notification simultaneously

J0taFerreira:
- fixed frontend not being able to show success and error notification at the same time

ref: #1395


frontend/src/store/NotificationState.ts
```diff
@@ -35,7 +35,6 @@ export const setNotificationMsg = (
     msg: string = '',
     fullMsg: string = ''
 ) => {
-    resetNotification()
```
ianliuwk1019, line 38: This change does work. However, some issues:
* The error message is not clear:
image
My understanding is this is the Case 1.

* ManagePermissions appears to have a bug for "delete" action.
Maybe I am not clear. So when the messages displayed already (previous case 1), and user still stays on the same component, and  user deletes a role assignment; then what do we expect to see?
Do we reset all previous messages and display current message (success/fail)?
image

J0taFerreira: Hello Ian! The case one will not be entering in this ticket I believe, Catherine and Olga will work on the design first if I'm not mistaken, and about the "delete" action, I think we would need to ask @OlgaLiber2. Sorry for the late reply!

MCatherine1994 (approved):

Issue #747, opened 2023-08-04
User duplication check for BCeID changed username


ConradBoydElliottGustafson:
**Describe the bug**
If a user has already logged into FAM with a BCeID, and then edits BCeID user name, the auth function will insert a new user record (with same GUID), since the 'on conflict' logic is done by user_type_code and user_name, not by user_guid.

**To Reproduce**
1. Log in through Cognito with a BCeID
2. Change BCeID username
3. Kill the Cognito session (through the console)
4. Log in through Cognito again
5. Observe 2 records in the DB for the same user

**Expected behavior**
A new record should only be created if it is a different BCeID. The auth function should compare the GUID first and do an update if the GUID already exists.

**Additional context**
This is legacy MVP behaviour. There is a case where we use the FAM console to assign a role to a user that has never logged in. Because we don't currently look up the user, we don't have the GUID at that time, so the check is done on username and usertype to handle that case. Until we have GUID reliably in every case, we will need to do both the GUID check and the existing one. The logic of the check likely has to change as I don't think we will be able to do the "on conflict" query in this case.

MCatherine1994:
This ticket we can wait a bit once we can search for bceid, so we will create the user with guid to solve the special cases in the additional content. 

- [ ] when user login through FAM, in the Auth Lambda function, update the auth function, to check conflict on user guid, and update the user name if needs
<img width="1721" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMFhEQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--2f7403e2ec8db1cbabd99091973f646303d9c9a1/image.png" alt="image.png" />

- [ ] when grant user access in FAM, verify the user name. Check user exists or not by user type and guid. If user does not exist, create the user with username, user type and guid. If exist, update the username if needs
 <img width="1656" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMGJEQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b43ad764ef7b92b7614833d0fe1992e327cbf930/Screen%20Shot%202024-01-12%20at%202.29.31%20PM.png" alt="Screen Shot 2024-01-12 at 2.29.31 PM.png" />

[...]

PR #1413, opened 2024-05-27
feat: #747 Store user guid when grant admins

MCatherine1994:
- Store user guid when grant application admin and when grant delegated admin
- Add user validator when grant admin access
- Fix tests 

Note: Still need to add some new tests, will do in following prs


server/admin_management/api/app/routers/router_guards.py
```diff
@@ -124,103 +140,66 @@ async def get_target_user_from_id(
...
-        user = user_service.get_user_by_domain_and_name(
-            rb_user_type_code,
-            rb_user_name,
+
+        target_new_user = TargetUser.model_validate(
+            {
+                "user_name": rbody["user_name"],
+                "user_type_code": rbody["user_type_code"],
+                "user_guid": rbody["user_guid"],
+            }
```
MCatherine1994, line 184: Most of the changes in this file are formatting. So here since we'll validate the user later, I don't feel we need to try to find the user from database anymore

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -124,103 +140,66 @@ async def get_target_user_from_id(
...
-                f"IDIM look up user guid:`{user_guid}` returns no result. " +
-                "business_guid will NOT be updated to target_user."
-            )
+    idim_validator_service = IdimValidator(requester, target_user)
+    search_result = idim_validator_service.verify_user_exist()
 
+    # add business_guid for BCeID user
+    if search_result.get("found") and search_result.get("businessGuid"):
+        target_user.business_guid = search_result.get("businessGuid")
     return target_user
```
MCatherine1994, line 203: Update the bceid search to a new method "get_verified_target_user", and we'll call the idim web service to find the user. If returns businessGuid, will add to the target user

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -234,10 +213,8 @@ async def enforce_self_grant_guard(
     """
     LOGGER.debug(f"enforce_self_grant_guard: requester - {requester}")
     LOGGER.debug(f"enforce_self_grant_guard: target_user - {target_user}")
-    is_same_user_name = requester.user_name == target_user.user_name
-    is_same_user_type_code = requester.user_type_code == target_user.user_type_code
 
-    if is_same_user_name and is_same_user_type_code:
+    if requester.user_guid == target_user.user_guid:
```
MCatherine1994, line 217: updated to check user_guid, because we can expect both requester and target_user will have the user_guid

server/admin_management/api/app/schemas.py
```diff
@@ -47,11 +47,11 @@ class Requester(BaseModel):
     # from datbase and the user record will contain this user_id; and also is the reference-id from other database
     # entities (e.g., app_fam.fam_application_admin, app_fam.fam_access_control_privilege)
     user_id: int
-    user_guid: Optional[Annotated[str, StringConstraints(max_length=32)]] = None
+    user_guid: Annotated[str, StringConstraints(min_length=32, max_length=32)] = None
```
MCatherine1994, line 50: I removed the "Optional",  cause we can expect requester and target_user should both have a user_guid when make an api call. And I updated the length of user_guid to be 32, so it won't happen the case that we got an empty string. Also checked the documentation, think the length is 32. 


Screen Shot 2024-05-23 at 4 40 11 PM
Screen Shot 2024-05-23 at 4 41 00 PM
Screen Shot 2024-05-23 at 4 41 16 PM

server/admin_management/tests/conftest.py
```diff
@@ -149,48 +152,47 @@ def setup_new_user(user_repo: UserRepository, db_pg_session: Session):
...
-            user_name,
-            cognito_user_id: Optional[str] = None) -> FamUser:
+        user_type: UserType, user_name, user_guid, cognito_user_id: Optional[str] = None
+    ) -> FamUser:
         new_user_create = FamUserDto(
             **{
-
                 "user_type_code": user_type,
                 "user_name": user_name,
+                "user_guid": user_guid,
```
MCatherine1994, line 163: Add user_guid when create new user

server/flyway/local_sql/V1000__update_test_user_cognito_id.sql
```diff
@@ -5,6 +5,10 @@ UPDATE app_fam.fam_user
 SET cognito_user_id = 'test-idir_b5ecdb094dfb4149a6a8445a0mangled@idir'
 WHERE user_name = 'COGUSTAF';
 
+UPDATE app_fam.fam_user
+SET user_guid = 'b5ecdb094dfb4149a6a8445a0mangled'
+WHERE user_name = 'COGUSTAF';
```
MCatherine1994, line 10: since now in the schema we make the requester's user_guid to be required, not optional anymore, add a fake user_guid for the test user

server/admin_management/api/app/routers/router_application_admin.py
```diff
@@ -58,6 +59,7 @@ def create_application_admin(
     request: Request,
     token_claims: dict = Depends(jwt_validation.authorize),
     requester: Requester = Depends(get_current_requester),
+    target_user: TargetUser = Depends(get_verified_target_user),  # validate target user
```
MCatherine1994, line 62: so we could verify the target user here. But I'm seeing we also try to get the target user from database for audit. But I feel for audit purpose, it's okay to use the target_user here directly? It will just tell from the given parameter, we're trying to grant access to which target user. In the current situation, if we want to grant access to a new user, it won't return from the user_service call, the user does not exist yet. But I feel it's okay to use the target_user here directly, for both grant app admin and grant delegated admin

Screen Shot 2024-05-27 at 12 37 19 PM

server/admin_management/api/app/routers/router_application_admin.py
```diff
@@ -97,6 +95,10 @@ def create_application_admin(
         raise e
 
     finally:
+        audit_event_log.target_user = user_service.get_user_by_domain_and_guid(
+            application_admin_request.user_type_code,
+            application_admin_request.user_guid,
+        )
```
MCatherine1994, line 101: get the target user after we created the user in the database

server/admin_management/api/app/services/user_service.py
```diff
@@ -19,32 +19,81 @@ def get_user_by_domain_and_name(self, user_type_code: str, user_name: str):
...
-                }
+            fam_user_by_domain_and_name = self.get_user_by_domain_and_name(
+                user_type_code, user_name
             )
+            # if not found user by user_guid, but found user by domain and name
+            # in the case of historical FAM user that has no user_guid stored
+            # add their user_guid
+            if (
+                fam_user_by_domain_and_name
+                and not fam_user_by_domain_and_name.user_guid
```
MCatherine1994, line 46: ideally, if we did a one-time load to add the user_guid for all existing users, we won't need this logic anymore. But it's also not hurt to have it

server/admin_management/api/app/services/validator/idim_validator.py
```diff
@@ -0,0 +1,85 @@
...
+            search_result = self.idim_proxy_service.search_idir(
+                IdimProxySearchParam(**{"userId": self.user_name})
+            )
+
+            # in edge case, the return guid from search doesn't match the guid given from request parameter
+            # this is unlikely to happen if the request comes from frontend because we also validate user in frontend
+            # but could happen if make backend api call directly
+            if (
+                search_result.get("found")
+                and search_result.get("guid").lower() != self.user_guid.lower()
```
ianliuwk1019, line 41: Hi Catherine, I probably didn't question guid format myself before, but now I have doubt....
Should the GUID format be case sensitive, so it needs to be exact encoding and not to compare UPPER() or lower()?
I think even IDIM doc isn't clear and even strange is guid from user pool might be case different....

ianliuwk1019 (changes requested):

server/admin_management/api/app/routers/router_guards.py
MCatherine1994: seems like we still need that to await request.json()

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -109,7 +122,7 @@ async def get_current_requester_without_access_check(
 # Very likely in future might have "cognito_user_id" case.
 async def get_target_user_from_id(
     request: Request,
-    user_service: UserService = Depends(user_service_instance),
+    # user_service: UserService = Depends(user_service_instance),
```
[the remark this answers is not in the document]
MCatherine1994: this is not in use anymore, I'll remove it

server/admin_management/api/app/services/validator/idim_validator.py
MCatherine1994: yeah, but I feel it's case insensitive, cause I saw both lower and upper cases before, forgot where.

server/admin_management/api/app/services/user_service.py
```diff
@@ -19,32 +19,81 @@ def get_user_by_domain_and_name(self, user_type_code: str, user_name: str):
...
 
         LOGGER.debug(f"User {fam_user.user_id} found.")
+        # update user_name if needs
+        fam_user = self.update_user_name(fam_user, user_name, requester)
         return fam_user
 
+    def update_user_name(
+        self,
+        user: models.FamUser,
+        user_name_from_request: str,
```
ianliuwk1019, line 82: Maybe "user_name_for_update" sounds better?

server/admin_management/tests/constants.py
```diff
@@ -119,19 +125,26 @@
...
 }
 
 TEST_ACCESS_CONTROL_PRIVILEGE_CREATE_REQUEST_CONCRETE = {
-    "user_name": TEST_USER_NAME,
+    "user_name": TEST_USER_NAME_IDIR,
     "user_guid": TEST_USER_GUID_IDIR,
     "user_type_code": famConstants.UserType.IDIR,
     "role_id": TEST_FOM_DEV_REVIEWER_ROLE_ID,
 }
+TEST_ACP_CREATE_CONCRETE_BCEID = {
```
ianliuwk1019, line 141: Since you use "_ACP_" here, not sure if you also like to rename other variable above related to "_ACP_" with short comment on what "_ACP_" is?

ianliuwk1019 (commented):
Just minor comments. Thank you Catherine for doing the improvement!!

server/admin_management/api/app/services/user_service.py
MCatherine1994: so here we try to compare the username in our database vs. the username from the request, if not the same, then we update. I feel "user_name_from_request" makes more sense?

server/admin_management/tests/constants.py
MCatherine1994: yeah I was thinking of that before, but if I made those changes in this pr, there are too many tests need to change, and I was only focus on the broken tests. But I'll refactor them when adding new tests later, in following prs

server/admin_management/api/app/services/validator/user_validator.py
```diff
@@ -0,0 +1,85 @@
...
+                )
+
+        if not search_result or not search_result.get("found"):
+            error_msg = f"Invalid request, cannot find user {self.user_name} {self.user_guid} with user type {self.user_type_code}"
+            utils.raise_http_exception(
+                error_code=ERROR_CODE_INVALID_REQUEST_PARAMETER,
+                error_msg=error_msg,
+            )
+
+        return search_result
```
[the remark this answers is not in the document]
MCatherine1994: Hi Ian, sorry I didn't get this, so what I need to do here? add a return type to the method?

server/admin_management/api/app/services/user_service.py
ianliuwk1019: I feel the variable should not be context aware because the method itself provide the logic to do the job and does not need to care who passes the value (from request, or from other method). The calling method needs to know what it needs to pass. Also this probably is a private method that's not meant for use from router/request (although the inputs most often are from the request).

server/admin_management/api/app/services/validator/user_validator.py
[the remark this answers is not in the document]
MCatherine1994: oh I see what you mean...make the return to be the type of TargetUser directly, and in the router we can just return the result from this method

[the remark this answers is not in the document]
MCatherine1994: um... but the return result schema is quite different than the TargetUser schema, I feel the logic is more clean currently, how you feel?

server/admin_management/tests/constants.py
ianliuwk1019: I see, thanks, I didn't notice that, sure, we can update them later time.

ianliuwk1019 (approved):

PR #1414, opened 2024-05-28
feat: #747 Store user guid when grant user access and update Auth Lambda function to check conflict on user guid

MCatherine1994:
- Store user guid when grant regular user access
- Add user validator when grant user access
- Fix backend tests
- Update unique constraint in the fam_user table to be on user_type_code and user_guid
- Update Auth Lambda logic to check conflict on user_guid
- Fix auth function tests

Note: Still need to add some new tests, will do in following prs


server/backend/api/app/crud/validator/user_validator.py
```diff
@@ -0,0 +1,101 @@
...
+            #     IdimProxyBceidSearchParam(
+            #         **{
+            #             "searchUserBy": IdimSearchUserParamType.USER_GUID,
+            #             "searchValue": self.verified_target_user.user_guid,
+            #         }
+            #     )
+            # )
+
+            # currently bceid cannot search bceid by user guid in TEST env, we just search by user name for now
+            search_result = self.idim_proxy_service.search_business_bceid(
```
MCatherine1994, line 63: Currently the bceid user cannot verify bceid user by user_guid, even in test env, use username for now. Will update to use user_guid after we enabled it.

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -50,7 +46,9 @@ def create_user_role_assignment(
     db: Session = Depends(database.get_db),
     token_claims: dict = Depends(jwt_validation.validate_token),
     requester: Requester = Depends(get_current_requester),
-    target_user: TargetUser = Depends(target_user_bceid_search)
+    target_user: TargetUser = Depends(
+        enforce_bceid_by_same_org_guard
```
MCatherine1994, line 50: so I let the enforce_bceid_by_same_org_guard to return the target user, and move it from router dependency to here, so we don't need to call the idim service twice

server/auth_function/test/login_event.json
```diff
@@ -16,15 +16,15 @@
             "cognito:user_status": "EXTERNAL_PROVIDER",
             "custom:idp_display_name": "Gustafson, Conrad CITZ:IN",
             "custom:idp_name": "idir",
-            "custom:idp_user_id": "B5ECDB094DFB4149A6A8445A0MANGLED",
+            "custom:idp_user_id": "B5ECDB094DFB4149A6A8445AMANGLED0",
```
MCatherine1994, line 19: make this guid different from the one we use for backend test, so we won't get conflict on user_guid

server/backend/api/app/crud/validator/user_validator.py
MCatherine1994: actually this is enabled in PROD env, we just need it to be enabled in test env as well

server/flyway/local_sql/V1000__update_test_user_cognito_id.sql
```diff
@@ -5,10 +5,17 @@ UPDATE app_fam.fam_user
 SET cognito_user_id = 'test-idir_b5ecdb094dfb4149a6a8445a0mangled@idir'
 WHERE user_name = 'COGUSTAF';
 
+-- add user_guid for COGUSTAF, as we use it as requester for tests, and requester should have user_guid
+-- please note that this user is different from the test user we use in auth_funtion tests
 UPDATE app_fam.fam_user
-SET user_guid = 'b5ecdb094dfb4149a6a8445a0mangled'
+SET user_guid = 'B5ECDB094DFB4149A6A8445A0MANGLED'
```
MCatherine1994, line 11: I feel maybe we switch to use all upper case for tests, cause I think from Cognito we'll get all upper case

server/auth_function/lambda_function.py
```diff
@@ -163,13 +163,32 @@ def populate_user_if_necessary(db_connection, event) -> None:
     else:
         user_name = event["request"]["userAttributes"]["custom:idp_username"]
 
+    cursor = db_connection.cursor()
+
+    # in the case of historical FAM user that has no user_guid stored, add their user_guid
+    # if user does not exist or user has a user_guid already, this update query will do nothing
+    query_add_guid= """
```
MCatherine1994, line 170: ideally, if we do a one-time load to add user_guid for all existing users, there is no need for this logic

MCatherine1994: fyi, there are 594 users in FAM PROD db that has no user_guid

server/backend/api/app/crud/crud_user.py
```diff
@@ -76,40 +89,59 @@ def create_user(fam_user: schemas.FamUser, db: Session):
...
+                fam_user_by_domain_and_name.user_id,
+                {models.FamUser.user_guid: user_guid},
+                requester,
+            )
+            LOGGER.debug(
+                f"User {fam_user_by_domain_and_name.user_id} found, updated to store their user_guid."
+            )
+            return fam_user_by_domain_and_name
+        # if not found user by user_guid, and not found user by domain and username,
+        # or found user by domain and username, but user guid does not match (this is the edge case that could happen when username changed from IDP provider)
```
ianliuwk1019, line 119: I wonder probably no need to add line 119 comment. It will be a new record in database to be inserted anyway (not found by user_guid and not found by user_name).

When IDP change "the same user's" username, is the user_guid also get changed? My assumption is no, but that might be a question to IDP...

server/backend/api/app/crud/validator/user_validator.py
ianliuwk1019: Thank you Catherine for the information. 
Since this is commented out, should we add comment in ticket 1181 at your TODO to remind that also this part of code need to be changed after test env is enabled as a follow-up change?

server/backend/api/app/routers/router_user_role_assignment.py
ianliuwk1019: It's the magic from "Depends", if the same depends being referenced more than once, it will not evaluate again.
So you don't need to move that validation to the parameters of the function. 
It simply make sense to be on "@router.post(dependencies=)" before any function execution. If moving to the function parameters as one of the dependency, it probably does not matter in this case, but that's already go beyond the life cycle of route and step into the route handler (which is the function). 

At the parameter you can, as it was, depends on "get_verified_target_user" and it will not call again externally (since it is a Depends).

ianliuwk1019 (changes requested):

server/backend/api/app/crud/validator/user_validator.py
MCatherine1994: Hi Ian, I already added the comment in 1181 for all the scenarios that are not working and I already sent the question to IDIM.

server/backend/api/app/routers/router_user_role_assignment.py
MCatherine1994: oh yes Ian, I remember you explained to me before, sorry I forgot, I'll change it back, thanks!!

server/auth_function/lambda_function.py
```diff
@@ -163,13 +163,32 @@ def populate_user_if_necessary(db_connection, event) -> None:
...
+
+    # insert new user, or update user information
     raw_query = """INSERT INTO app_fam.fam_user
         (user_type_code, user_guid, cognito_user_id, user_name, business_guid,
         create_user, create_date, update_user, update_date)
         VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid},
         CURRENT_USER, CURRENT_DATE, CURRENT_USER, CURRENT_DATE)
-        ON CONFLICT (user_type_code, lower(user_name)) DO
-        UPDATE SET user_guid = {user_guid},  cognito_user_id = {cognito_user_id}, business_guid = {business_guid};"""
+        ON CONFLICT (user_type_code, lower(user_guid)) DO
```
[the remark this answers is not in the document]
MCatherine1994: you're right, I verified yesterday, in our production data, all the users that have user_guid are using all upper cases, I'll remove the lower here.

server/backend/api/app/crud/crud_user.py
MCatherine1994: user_guid will never change I believe

MCatherine1994: so the comment on 119, I mean we found the user by domain and username (USER A) in our database, however this username is now used by another person (USER B), not the original person anymore, that's why the user_guid is not the same. This could happen that we initially grant USER A some access, so we have record for A. Later on, A changed his username, but never login to FAM, so our record is old. And then USER B changed his name to A's old name. When we grant B access, we got A's old name with B's user_guid, in this case, we should add a new user record for USER B. And once USER A login to FAM or any apps integrated with FAM, we'll update A's username.

server/flyway/sql/V50__update_fam_user_constraint_use_guid.sql
```diff
@@ -0,0 +1,5 @@
+-- Update unique constraint in fam user table to be user type code and user guid
+
+DROP INDEX app_fam.fam_usr_uk;
+
+CREATE UNIQUE INDEX fam_usr_uk ON app_fam.fam_user(user_type_code, lower(user_guid));
```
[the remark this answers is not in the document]
MCatherine1994: I feel it doesn't hurt to add lower here just to avoid human error, when someone try to add user using api call directly (unlikely happen though). I agree to remove the lower in lambda function

server/auth_function/lambda_function.py
MCatherine1994: I agree, that will become a historical record that never get update in the future. But I agree it's a very rare case.

server/flyway/sql/V50__update_fam_user_constraint_use_guid.sql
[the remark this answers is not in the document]
ianliuwk1019: Hi Catherine,
I am in doubt of IDIM GUID format, and feel uncertain what it could be.
If we are speaking from real GUID format, it is a hexdecimal format thus it is case-insensitive(if I understand correctly).
But, **IDIM GUID format** from doc says **32 alphanumeric characters**, not the real GUID format.
It says "**Case preserving**", so I might guess it does not just generate from `0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ` characters, but also `abcdefghijklmnopqrstuvwxyz` characters.
However, it then says "**By convention, this is upper case**", this is really **confusing statement for
developers**. Does IDIM just for now generate that GUID using `0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ` but reserve using `abcdefghijklmnopqrstuvwxyz` in future?

So, at the end, treating it as case-sensitive or case-insensitive (using lower/upper) **is the same (if their
"By convention statement" stays true**; and thus it might not hurt in our **ON CONFLICT** statement we use
**(user_type_code, lower(user_guid))**.

But still, I am really curious if user A has (GUID = '**123IAMUNIQUE456**') and suddenly in IDIM system it has
user B (GUID='**123IAMUNIQUe456**') or user C (GUID='**123iAMUNIQUE456**') or user xyz (GUID='**123iamunique456**').... and **is allowed** because it says it is"**Case preserving**", then we are going to treating them as the same user with big hard to spot problem.

Although I favor not to compare GUID with lower/upper as we don't know if it is still possible the GUID is case-sensitive, but I think I am ok for what you have here.

Wonder if @basilv  has more understanding of this IDIM GUID field and has other opinion.

server/flyway/local_sql/V1000__update_test_user_cognito_id.sql
ianliuwk1019: Thanks. Good change I think!

server/auth_function/lambda_function.py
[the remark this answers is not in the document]
basilv: Yes, the spec from IDIM seems pretty clear that the GUID values are all upper case. I would go with accepting the GUID values as supplied (don't do lower/upper), and compare them without lower/upper.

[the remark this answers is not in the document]
MCatherine1994: I feel it doesn't hurt to add constraint by checking the lower case `CREATE UNIQUE INDEX fam_usr_uk ON app_fam.fam_user(user_type_code, lower(user_guid));`, just to avoid human errors if create user directly by calling api (unlikely to happen in prod, but could happen in dev or test).

ianliuwk1019 (approved):
Thank you for complicated change!

PR #1418, opened 2024-05-31
fix: #747 Support nullable user guid for existing FAM user

MCatherine1994:


server/admin_management/api/app/schemas.py
```diff
@@ -47,7 +47,7 @@ class Requester(BaseModel):
     # from datbase and the user record will contain this user_id; and also is the reference-id from other database
     # entities (e.g., app_fam.fam_application_admin, app_fam.fam_access_control_privilege)
     user_id: int
-    user_guid: Annotated[str, StringConstraints(min_length=32, max_length=32)] = None
+    user_guid: Annotated[str, StringConstraints(min_length=32, max_length=32)]
```
MCatherine1994, line 50: requester user_guid should never be None

server/backend/api/app/schemas.py
```diff
@@ -37,7 +37,7 @@ class FamUser(BaseModel):
         None  # temporarily optional
     )
     user_name: Annotated[str, StringConstraints(max_length=20)]
-    user_guid: Annotated[str, StringConstraints(min_length=32, max_length=32)] = None
+    user_guid: Union[Annotated[str, StringConstraints(min_length=32, max_length=32)], None]
```
MCatherine1994, line 40: for existing FAM user, user_guid is nullable

ianliuwk1019 (approved):

PR #1419, opened 2024-05-31
feat: #747 Add tests for user validator and user service 

MCatherine1994:
- Add test for new user service method find_or_update
- Add test for new user service method update_user_name, update_user_business_guid
- Add test for user validator in backend, skip the test in admin management because of duplication


server/backend/testspg/crud/test_user_validator.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+LOGGER = logging.getLogger(__name__)
+
+# when run test locally, we can use our own idir username and user guid
+TEST_IDIR_USER_GUID = os.environ.get("TEST_IDIR_USER_GUID")
+TEST_IDIR_USER_NAME = os.environ.get(
+    "TEST_IDIR_USER_NAME"
+) or TEST_IDIR_REQUESTER_DICT.get("user_name")
+
```
MCatherine1994, line 19: I only added the test for user validator in backend, cause backend already setup with real IDIR GUID. I didn't add the test in admin management because of duplication, if we feel we need, I can add it there as well.

server/backend/testspg/crud/test_user_validator.py
```diff
@@ -0,0 +1,138 @@
...
+        assert verified_target_user.user_guid == target_user.user_guid
+        assert verified_target_user.user_type_code == target_user.user_type_code
+        assert verified_target_user.user_name == target_user.user_name
+        assert verified_target_user.business_guid is not None
+
+    # def test_verify_user_exist_bceid_not_found(self):
+    #     target_user = TargetUser(
+    #         **{
+    #             **TEST_TARGET_USER_BCEID_LOAD_2,
+    #             "user_guid": "USERGUIDNOTEXISTSPOJHSLEJFNSEKSL",
```
MCatherine1994, line 102: Skip the two tests for now, cause we suppose to search bceid on user_guid, but because that's not enabled in TEST env, we have to search on user_name. Once that's enabled, we can add the test back

server/admin_management/tests/services/test_user_service.py
```diff
@@ -44,3 +50,101 @@ def test_find_or_create(user_service: UserService):
...
+    updated_user = user_service.update_user_business_guid(
+        found_user.user_id,
+        TEST_USER_BUSINESS_GUID_BCEID,
+        TEST_NEW_IDIR_USER.create_user,
+    )
+    assert updated_user.business_guid == TEST_USER_BUSINESS_GUID_BCEID
+    # verify the business_guid is updated
+    found_user = user_service.get_user_by_domain_and_name(
+        TEST_NEW_BCEID_USER.user_type_code, TEST_NEW_BCEID_USER.user_name
+    )
```
ianliuwk1019, line 123: Not sure if I that's what I thought, maybe you don't even need to do `found_user = user_service.get_user_by_domain_and_name` because of line 108 is the same query to the database...
So if you remove line 121 to 123, probably the below assert (124/125) might still be working.
However, just wondering, no need to change any code....

MCatherine1994: yeah you're right, it's the same, just to make sure if we try to find again, it returns the same

server/backend/testspg/crud/test_user_validator.py
ianliuwk1019: Cool! thanks Catherine.
Since we already had `IdimProxyService` test which uses real TEST_IDIR_USER_GUID, this one is a validator which uses `IdimProxyService`, then ideally we don't need to use real ID anymore here with real connection to IDIM-Proxy.
A proper mocking the `IdimProxyService` returns on (`search_idir` and `search_business_bceid`) for several test scenarios would be better. We probably don't have this mocking example yet, but might not be difficult (just need to use python mock package and mock the `IdimProxyService` import I guess), only the mock data is a little hard to come up with.
Would you want to give it a try?

MCatherine1994: I was thinking of that before, but because of the edge case for information mismatch I want to use the real one for testing. I could do the mocked one in the admin management.

ianliuwk1019: Sure, but I am not sure what difference it would make for mismtach cases.
I can later see if I can do this improvement, because I feel the real IDIR ideally should only be used in real tests on external connection.

ianliuwk1019 (dismissed):

server/admin_management/tests/services/test_user_validator.py
```diff
@@ -0,0 +1,177 @@
...
+    "found": True,
+    "guid": TEST_NEW_TARGET_USER_BCEID_LOAD_2.get("user_guid"),
+    "userId": TEST_NEW_TARGET_USER_BCEID_LOAD_2.get("user_name"),
+    "businessGuid": TEST_USER_BUSINESS_GUID_BCEID,
+}
+
+
+class TestUserValidatorClass(object):
+    """
+    Testing UserValidator class with real remote API calls (TEST environment).
```
ianliuwk1019, line 56: Hi Catherine, we can probably now remove this comment.

server/admin_management/tests/services/test_user_validator.py
```diff
@@ -0,0 +1,177 @@
+import logging
+import pytest
+from fastapi import HTTPException
+from mock import patch
+
+from api.app.services.validator.user_validator import UserValidator, IdimProxyService
```
ianliuwk1019, line 6: Just curious to me, you can directly import "IdimProxyService" from `api.app.services.validator.user_validator`??, and use it for mock, wow!!

server/admin_management/tests/services/test_user_validator.py
MCatherine1994: yeah updated, thanks!

server/admin_management/tests/services/test_user_validator.py
MCatherine1994: oh yeah, I was surprised as well! but I changed the import to from the integration, feel that makes more sense

server/backend/api/app/crud/validator/user_validator.py
```diff
@@ -49,21 +49,11 @@ def verify_user_exist(self) -> TargetUser:
...
-            # )
-
-            # currently bceid cannot search bceid by user guid in TEST env, we just search by user name for now
             search_result = self.idim_proxy_service.search_business_bceid(
                 IdimProxyBceidSearchParam(
                     **{
-                        "searchUserBy": IdimSearchUserParamType.USER_ID,
-                        "searchValue": self.verified_target_user.user_name,
+                        "searchUserBy": IdimSearchUserParamType.USER_GUID,
+                        "searchValue": self.verified_target_user.user_guid,
```
MCatherine1994, line 56: IDIM enabled the bceid search bceid by user_guid, so enable it here as well

server/admin_management/tests/services/test_user_validator.py
```diff
@@ -0,0 +1,178 @@
...
+class TestUserValidatorClass(object):
+    """
+    Testing UserValidator class with mocked IDIM API calls.
+    """
+
+    def setup_class(self):
+        # local valid mock requester
+        self.requester_idir = Requester(**TEST_REQUESTER)
+
+    @patch.object(IdimProxyService, "search_idir")
```
ianliuwk1019, line 64: Only wondering. Do you know if this "patch" to mock the service need to be reset before each test case begins?
Also, do you know if we can verify if the mocked service is being called?
Thanks!

ianliuwk1019 (approved):

server/admin_management/tests/services/test_user_validator.py
MCatherine1994: I'm actually not sure, but if you check the `server/backend/testspg/crud/test_UserRoleValidator.py` file, we used patch before the mock forest client api return. And we can verify if the service is being called
Screen Shot 2024-06-04 at 2 45 06 PM

ianliuwk1019: What?! Oh.... man.... I used patch recently and didn't even remember I did.... I am getting so bad in memory.... so sorry, if I remembered I should point to you this file....
I approve... but thank you for reminding me!!

Issue #1421, opened 2024-06-03
Disable adding forest client temporarily


MCatherine1994:
**Describe the task**
Forest client verification is not working as the Forest Client API is down. So for now:
- when they select concrete role, no forest client number needs, work as normal
- when they select abstract role, instead of asking them enter a forest client number, will just display a message says "The client API is down and this role cannot be assigned. We are working on fixing the issue. Please try again tomorrow."

**Acceptance Criteria**
- [x] Disable adding forest client numbers
- [x] Hide Delegated admin tab from dashboard, hide add delegated admin option from side nav (this is not typically related to the first client api down, just because our delegated admin features are not fully completed, hide them to less confuse our users)

**Additional context**
- Will do a hotfix release this time, just because the new features from last deployment v1.16.0 (mainly bceid verification) are not fully working in all environments. Will apply the fixes on top of the last release v1.16.0. 
- Or here

MCatherine1994:
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/30ebcc79-2a96-478d-81bc-766e9f76063a)

MCatherine1994:
The forest client api issues was solved on June 4, enabled the forest client verification back into Production.

PR #1422, opened 2024-06-03
fix: #1421 Disable adding forest client, and hide delegated admin features

MCatherine1994:
refs: #1421 

- Disable adding forest client number for abstract role
- Hide creating/managing delegated admin functionalities 

Please note: the tests will fail in the pipeline because Forest Client API is down


ianliuwk1019 (approved):
Thanks Catherine!!

PR #1426, opened 2024-06-04
fix: #1421 Enable verifying forest client number as API problem solved

MCatherine1994:
refs: #1421


ianliuwk1019 (approved):
Thanks Catherine!!

Issue #1334, opened 2024-04-24
Improve user table behaviour


OlgaLiber2:
**Describe the task**
Improve user table behaviour (including delegated admin table)

**Acceptance Criteria**
- [x] add hover over to the user table
- [x] when a new user is added, insert the entry at the top of the table, highlight the row and add "new" tag next to it.
- [x] make the notification faded out on next action

**Additional context**
- Until the next action (next page, new tab), same as when we remove the notifications
- For user table, delegated admin table and application admin table
- when add multiple access, highlighted all 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/870918b3-f361-4307-8cb6-5555a1c62e78)

PR #1430, opened 2024-06-05
fix: #1334 improve user table behaviour

J0taFerreira:
- add hover over to the table
- when a new user is added, insert the entry at the top of the table, highlight the row and add "new" tag next to it.
- make the notification faded out on next action
ref: #1334 

styling will be broken before the common repo is approved


J0taFerreira:
SonarClaud is complaining about the the lack of testing for the added code

frontend/src/store/newUserComparatorState.ts
```diff
@@ -0,0 +1,180 @@
...
+import type { TabKey } from '@/enum/TabEnum';
+
+interface NewUsersValue {
+    userAccess: FamUserRoleAssignmentCreate[];
+    adminAccess: FamAppAdminCreateRequest[];
+    delegatedAdminAccess: FamAccessControlPrivilegeCreateRequest[];
+}
+
+// Initialize an array to store new users
+const newUsers = ref<NewUsersValue>({
```
MCatherine1994, line 21: Hi Jota, I wonder is there any reason that we need to track the new users for all three cases at the same time? I feel each time, there will only have one type of new user access, not sure if we really need an object to store three arrays.

frontend/src/store/newUserComparatorState.ts
```diff
@@ -0,0 +1,180 @@
+import { ref } from 'vue';
```
MCatherine1994, line 1: Hi Jota, we need to rename this file.  If you check other store files, it needs to start with upper case letter. And this is actually not new user, it's the new user access. Maybe it's better to call it `GrantUserAccessState.ts`. And you might want to rename your variables in this file, instead of calling them new users, change to new user access.

frontend/src/store/newUserComparatorState.ts
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+/**
+ * Compares a list of user role assignments and identifies new users.
+ * @param userRoleAssignments An array of user role assignments.
+ * @returns A sorted array of user role assignments with an additional property indicating if the user is new.
+ */
+export const compareUserTable = (
+    userRoleAssignments: FamApplicationUserRoleAssignmentGet[] = []
```
MCatherine1994, line 63: I wonder instead of creating these compare method, maybe we can modify the fetch methods. So before entering each route, we'll fetch the user access data. And when do the sorting here, found the new user access record, and put it on the top. 

Screen Shot 2024-06-06 at 11 58 17 AM

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
```diff
@@ -107,7 +109,7 @@ const deleteDelegatedAdmin = (
             />
             <DataTable
                 v-model:filters="delegatedAdminFilters"
-                :value="props.delegatedAdmins"
+                :value="compareDelegatedAdminTable(props.delegatedAdmins)"
```
MCatherine1994, line 112: so instead of comparing the values, could we set the style for the rows, where row index is less or equal to the length of the new user access? like for example the top 3 rows are the new access with the specific style

frontend/src/components/common/NotificationStack.vue
```diff
@@ -1,26 +1,71 @@
...
 
 <template>
-    <template v-for="(value, key) in notifications" :key="key">
-        <NotificationMessage
-            v-if="value.msg !== ''"
-            :severity="key.toString()"
-            :msg-text="value.msg"
-            :hasFullMsg="value.fullMsg !== ''"
-            class="notification-stack"
+    <Message
```
MCatherine1994, line 11: Hi Jota, just curious, do we need to change the notification message for this ticket?

frontend/src/store/newUserComparatorState.ts
MCatherine1994: and instead of comparing user_name and etc, we could compare the primary key that get from the return of the POST request, so you should store the result of the POST request, and only do that for successful cases. 
Screen Shot 2024-06-06 at 12 36 12 PM
Screen Shot 2024-06-06 at 12 29 48 PM
Screen Shot 2024-06-06 at 12 36 44 PM
Screen Shot 2024-06-06 at 12 31 20 PM

Screen Shot 2024-06-06 at 12 37 28 PM
Screen Shot 2024-06-06 at 12 30 51 PM

frontend/src/store/newUserComparatorState.ts
J0taFerreira: It sure can be just one, I'll refactor it

frontend/src/store/newUserComparatorState.ts
J0taFerreira: sure, I'll make the change.

frontend/src/components/common/NotificationStack.vue
J0taFerreira: Refactoring and removing the need for the v-if condition from the notification message restores the PrimeVue animation.

frontend/src/store/newUserComparatorState.ts
J0taFerreira: sounds good, I'll follow your suggestion

ianliuwk1019 (changes requested):

ianliuwk1019:
I only look through few files, maybe there were challenges you faced, but feeling this implementation is too confusing.
Previous NotificationState was already very confusing, I feel adding another confusing 'newUserComparatorState' isn't good (unless I just don't get your point ). They will cause maintenance headache.
Seems to me there is fundamental issues here.
Perhaps we can discuss next week with @MCatherine1994 (when she has time, she is very busy), and gather some idea and improvements.

frontend/src/store/newUserComparatorState.ts
```diff
@@ -0,0 +1,180 @@
...
+    arrayToUpdate: TabKey
+) => {
+    newUsers.value = {
+        ...newUsers.value,
+        [arrayToUpdate]: [...newUsers.value[arrayToUpdate], newItem],
+    };
+};
+
+export const highlightNewUserRow = (rowData: any) => {
+    console.log('row', rowData)
```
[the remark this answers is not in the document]
J0taFerreira: Hello Ian! Already taken care off, thanks!

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -139,6 +140,7 @@ const handleSubmit = async () => {
             await AppActlApiService.userRoleAssignmentApi.createUserRoleAssignment(
                 data
             );
+            setNewUsers(data, TabKey.UserAccess);
```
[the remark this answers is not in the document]
J0taFerreira: I'm already refactoring to move the logic and use the response, about the naming I'll try to find a better name

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -110,11 +113,13 @@ const deleteAdmin = (admin: FamAppAdminGetResponse) => {
                 :paginatorTemplate="TABLE_PAGINATOR_TEMPLATE"
                 :currentPageReportTemplate="TABLE_CURRENT_PAGE_REPORT_TEMPLATE"
                 stripedRows
+                :rowStyle="highlightNewUserRow"
             >
                 <template #empty> No user found. </template>
                 <template #loading> Loading users data. Please wait. </template>
                 <Column header="User Name" sortable field="user.user_name">
                     <template #body="{ data }">
+                        <NewUserTag v-if="data.isNewUser" />
```
MCatherine1994, line 122: Hi  Jota, I don't think you need to add the isNewUser attribute to the data, you can just check v-if="data.application_admin_id in [your_saved_new_user_access]"

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
```diff
@@ -71,9 +72,13 @@ const handleSubmit = async () => {
         : '';
 
     try {
-        await AdminMgmtApiService.applicationAdminApi.createApplicationAdmin(
+        const returnResponse = await AdminMgmtApiService.applicationAdminApi.createApplicationAdmin(
             data
         );
+        await router.push({
+            name: routeItems.dashboard.name,
+            params: { newUserInTable: returnResponse.data.application_admin_id }
```
MCatherine1994, line 80: I still feel it's not actually the new user, it's the new user access, shall we rename this "newUserInTable" to "newUserAccess"

frontend/src/components/common/NotificationStack.vue
MCatherine1994: Hi Jota, sorry maybe I miss understood, which animation we're trying to get here?

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -78,9 +123,37 @@ onUnmounted(() => {
...
+    first: UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew,
+    second: UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew
+) {
+    // First, sort by isNewUser (true first)
+    if (first.isNewUser && !second.isNewUser) return -1;
+    if (!first.isNewUser && second.isNewUser) return 1;
+    return 0;
+}
+
+const resetNewAddedTag = () => {
```
MCatherine1994, line 136: so if we check if user is new or not in the table component, we don't need these anymore. So you just need to reset the newUserInTable to be ""

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
ianliuwk1019: Jota, as mentioned before, it is not a new user and 'newUserInTable' does not make sense.
Just give it a meaningful id name you are passing.
Like `newUserAccessId`, `newAppAdminId` etc...

frontend/src/router/index.ts
```diff
@@ -51,7 +51,7 @@ const routes = [
         component: LandingView,
     },
     {
-        path: routeItems.dashboard.path,
+        path: `${routeItems.dashboard.path}/:newUserInTable?`,
```
ianliuwk1019, line 54: The 'newUserInTable' name does not make sense.
However, this is very interesting, what do you mean by `/:newUserInTable?`? What do you expect it to happen? Is `:newUserInTable` a placeholder? Then what about the "?" mark at the end?

Does it show up on browser address bar (like /123?) ? I can see it didn't and that puzzles me, is it some feature from vue-router I didn't notice? But if you look closely the browser behavior, it actually did it twice for routing. Maybe I misunderstand something in vue-router...

ianliuwk1019 (changes requested):

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -111,6 +112,7 @@ const confirmSubmit = async () => {
     const username = formData.value.userId.toUpperCase();
     const role = getSelectedRole()?.name;
     const successList: string[] = [];
+    const newlyAddedList: number[] = [];
```
ianliuwk1019, line 115: Same here, maybe `newDelegatedAdminAccessIds` or something similar or at least add a comment for variable to explain what it holds.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -142,6 +145,13 @@ const confirmSubmit = async () => {
                 ? formData.value.verifiedForestClients
                 : [''];
     }
+
+    const paramListString = newlyAddedList.join(',');
+    await router.push({
+        name: routeItems.dashboard.name,
+        params: { newUserInTable: paramListString }
```
ianliuwk1019, line 152: You may just use line 149 here as "newlyAddedList.join(',')"

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -54,18 +54,63 @@ const props = defineProps({
         type: Array as PropType<FamAccessControlPrivilegeGetResponse[]>,
         default: [],
     },
+    newUserInTable: {
```
ianliuwk1019, line 57: Not a good name.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -54,18 +54,63 @@ const props = defineProps({
...
 
+type UserRoleOrAppAdminOrAccessControlPrivilege =
+    | FamApplicationUserRoleAssignmentGet
+    | FamAppAdminGetResponse
+    | FamAccessControlPrivilegeGetResponse;
+
+type UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew =
+    UserRoleOrAppAdminOrAccessControlPrivilege & {
+        isNewUser: boolean;
+    };
```
ianliuwk1019, line 71: Sorry, I am confused, what are these two **type**s and `isNewUser` for?
Also the name, do you mean `isNewUserAccess`?

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -78,9 +123,37 @@ onUnmounted(() => {
...
 
+function sortByNewUserAndUserName(
+    first: UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew,
+    second: UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew
+) {
+    // First, sort by isNewUser (true first)
+    if (first.isNewUser && !second.isNewUser) return -1;
+    if (!first.isNewUser && second.isNewUser) return 1;
+    return 0;
+}
```
ianliuwk1019, line 134: Sorry, maybe I didn't get it, where is the userName in the sorting? Could you add some function comment?

ianliuwk1019 (changes requested):

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -54,18 +54,63 @@ const props = defineProps({
...
+type UserRoleOrAppAdminOrAccessControlPrivilege =
+    | FamApplicationUserRoleAssignmentGet
+    | FamAppAdminGetResponse
+    | FamAccessControlPrivilegeGetResponse;
+
+type UserRoleOrAppAdminOrAccessControlPrivilegeWithIsNew =
+    UserRoleOrAppAdminOrAccessControlPrivilege & {
+        isNewUser: boolean;
+    };
+const newUserInTable = props.newUserInTable.split(',').map(Number);
```
ianliuwk1019, line 72: This `newUserInTable` being used among all `userRoleAssignments`, `applicationAdmins` and `delegatedAdmins` 3 ref will have bad consequences. See the below code that you are trying to identify the `isNewUser`.
Are you able to identify the `newUserInTable` is really for just `userRoleAssignments` or just for `applicationAdmins` or just for `delegatedAdmins`, or for ALL of them?

Even though we currently display ONE TAB at a time, this really will confuse other developers (that might be the least bad thing) or worse that other developers may not even notice (and having headache debugging what goes wrong later)  because this bug is very hidden.

ianliuwk1019 (changes requested):

ianliuwk1019:
@J0taFerreira 
I suggest passing whichever `[newly granted access id(s)]` into each "**target**" table/componet ( "Application Admin", "Delegated Admin" and "User") as a prop, and make your logic inside that component would be better. You seem trying to unify too much and instead making them very confusing. But you will need to know how to identify which table/component to pass your `[newly granted access id(s)]`. One suggestion is pass another route parameter from previous route, second option is name your route parameter specific (as what type of id: user, admin or delegated amin ID) from previous route.

Do we actually need to sort it by userName initially? If no requirement for that (which I don't remember we have), then when component is mounted, why not just sorted with "**identifier ID**" in **descending** order (very easy, and highest number is always the newest in database) and highlight with the first [n] rows on top where [n] are the count/length of your `[newly granted access id(s)]` list (if not null)? The users can sort it as they wish if by "userName" afterward.

I can see this "ManagePermissions.vue" with many tabs as child components probably should have been better if that was designed as child routes and passing loading to child components is redundant. No need to change them, just what I saw and note.

frontend/src/components/grantaccess/GrantApplicationAdmin.vue
J0taFerreira: sorry, I'll change the naming

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
J0taFerreira: Hello! ok, I can do that

frontend/src/router/index.ts
J0taFerreira: The ` /:newUserInTable?` is a placeholder and the "?" means its optional. About it doing twice I'm not sure, I'll see what I can find.

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: I should have removed the userName from the name after I refactor it.

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: that makes sense, I'll refactor it to avoid that

frontend/src/components/common/NotificationStack.vue
J0taFerreira: The animation on the message leave event, I restored the primevue animation but I noticed that was not working ok, so I added a transition-group. should be working as expected now

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: I removed then, as they are not needed anymore.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -54,8 +54,30 @@ const props = defineProps({
...
     },
+    newUserAccessIds: {
+        type: String,
+        default: '',
+    },
+    newAppAdminId: {
+        type: String,
+        default: '',
+    },
+    newDelegatedAdminIds: {
```
MCatherine1994, line 65: Hi Jota, these names look good, but we don't need three props, we can just use one name for all three cases. They are doing the same thing, and we should always keep the code simple as we can.  Let's just call it `newUserAccessIds`, default = [], so for app admin, pass [app_admin_id]

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -123,6 +147,7 @@ const deleteUserRoleAssignment = async (
 
 const deleteAppAdmin = async (admin: FamAppAdminGetResponse) => {
     resetNotification();
+    resetNewTag();
```
MCatherine1994, line 150: so even though we can to keep `resetNotification()` and `resetNewTag()` as separate method, but inside this component, you can create a method to include two, and call this method everytime

const resetNotificationAndTag = () = > {
     resetNotification();
     resetNewTag();
}

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -29,6 +31,26 @@ const props = defineProps({
...
         required: true,
     },
+    newAppAdminId: {
+        type: Number || undefined,
+        default: undefined,
+    },
+});
+
+// newAppAdminId goes to the top of the array
+const applicationAdmins = computed(() => {
```
MCatherine1994, line 41: I feel we don't need to sort each table data in the table component, so all the data initially was got from `frontend/src/services/fetchData.ts`, we should sort them in those fetch methods.

frontend/src/components/common/NotificationStack.vue
MCatherine1994: Hi Jota, could you please capture a screenshot for me about the animation? thanks!! or maybe you can show me when we chat today, I just want to avoid unnecessary changes for the notification part. thanks!!

frontend/src/router/index.ts
ianliuwk1019: Always add **clear comment** for this kind of odd looking string (if it was provided by vue-router doc, better give a reference to it), so other people would not be puzzled for that.

Please look into why it routed twice, if you see it triggers fetching twice, I would not feel good for it (I would not be surprised it is a bug from vue router if it was provided such kind of syntax but fetching twice).

frontend/src/components/managePermissions/ManagePermissions.vue
MCatherine1994: as discussed, we need to tell the new ids are for which table, so it's okay to keep three for now, thanks!!

frontend/src/components/managePermissions/ManagePermissions.vue
J0taFerreira: sure, I'll follow your suggestion

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
J0taFerreira: I will try to move it to the fetch file

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -73,6 +81,15 @@ const deleteAdmin = (admin: FamAppAdminGetResponse) => {
         },
     });
 };
+
+const highlightNewAppAdminAccesRow = (rowData: any) => {
```
MCatherine1994, line 85: Hi Jota, sorry I missed this one in the past, but shall we make this a reusable method as well for all three tables? I know we'll have a completely new design later and these might change as well, but maybe for now let's still try to make it reusable? thanks!

MCatherine1994, line 85: I was thinking if we want to make this method reusable for all three tables, but since our new design looks quite different, I feel it's okay to keep it as now. But let's make this style reusable, maybe move the return style into constant, thanks!

frontend/src/router/index.ts
```diff
@@ -51,7 +51,8 @@ const routes = [
         component: LandingView,
     },
     {
-        path: routeItems.dashboard.path,
+        // a param can be made optional by adding "?"
+        path: `${routeItems.dashboard.path}/:newUserAccessIds?/:newAppAdminId?/:newDelegatedAdminIds?`,
```
ianliuwk1019, line 55: I do not think this is the intended usage from vue-router, although the "?" means optional.
This looks like a path matching but you have used it as optional and combined with several of them, it has no meaning if it does work (such as `/dashboard/` or `/dashboard/1/`)
I think this still needs to be changed.

frontend/src/services/utils.ts
```diff
@@ -46,3 +48,21 @@ export const formValidationSchema = (isAbstractRoleSelected: boolean) => {
...
+export const resetNewTag = () => {
+    router.push({
+        name: routeItems.dashboard.name,
+        params: {
+            newUserAccessIds: '',
+            newAppAdminId: '',
+            newDelegatedAdminIds: '',
+        },
+    });
+};
```
ianliuwk1019, line 68: This seems to be very specific only for dashboard route, probably shouldn't be in /servics/utils.ts

ianliuwk1019 (changes requested):

frontend/src/router/index.ts
MCatherine1994: Hi Ian, please check the router again, but basically updated to use the query parameter. The only thing is that when I try to access to `/dashboard/`, it doesn't work. But `/dashboard?newUserAccessIds=36,37` works fine.  I can show you tomorrow

MCatherine1994:
Hi Jota, please check the update I pushed. We don't have to refactor the notification state, so I revert it back and transition animation seems work fine. Please check the common theme repo and remove unnecessary style change as well. Thanks so much! We can touch base tomorrow.

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -160,8 +164,19 @@ const handleSubmit = async () => {
...
     );
+
     setCurrentTabState(TabKey.UserAccess);
-    router.push('/dashboard');
+
+    if (newUserAccessIds.length > 0) {
+        router.push({
+            path: `/${routeItems.dashboard.name}`,
+            query: {
+                newUserAccessIds: newUserAccessIds.join(','),
```
MCatherine1994, line 174: we need to be really clear about the query parameter type, they will all be string, so for example '1' for single, or '1,2,3' for multiple

frontend/src/services/fetchData.ts
```diff
@@ -9,33 +9,43 @@ import {
...
 import FamLoginUserState from '@/store/FamLoginUserState';
 import { selectedApplicationId } from '@/store/ApplicationState';
+import { isNewAccess } from './utils';
 
 // --- Fetching data (from backend)
 
 export const fetchUserRoleAssignments = async (
-    applicationId: number | undefined
+    applicationId: number | undefined,
+    newUserAccessIds: string = ''
```
MCatherine1994, line 18: all the fetch method will have a default value for query parameters, so inside method, when we the split method, it will return an empty string if no query parameter value is provided

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -74,13 +83,25 @@ const applicationsUserAdministers = computed(() => {
 
 const tabViewRef = ref();
 
-onUnmounted(() => {
+const resetNewTag = () => {
+    router.push({ query: {} });
+    newAppAdminId.value = '';
+    newUserAccessIds.value = '';
+    newDelegatedAdminIds.value = '';
```
MCatherine1994, line 90: here I'm not so sure, we need to clean the query parameter for the router, but somehow I found we need to manually reset these local parameter values.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -56,6 +58,13 @@ const props = defineProps({
     },
 });
 
+// get user access ids from router query parameters
+const { query } = useRoute();
```
MCatherine1994, line 62: and ManagePermissions is the parent component of the dashboard router, so the router query parameter should be received at this level, and then pass to child components, instead of reading from child components directly, even though read from child components will also work, but that's not the correct logic.

ianliuwk1019: Using useRoute() is good too; just a comment, not to request for a change. 
Using Props, even on the this parent component feels the component would be more stand-alone itself and will be easy to plug and play and probably also will be easier for testing.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -74,13 +83,25 @@ const applicationsUserAdministers = computed(() => {
...
+    router.push({ query: {} });
+    newAppAdminId.value = '';
+    newUserAccessIds.value = '';
+    newDelegatedAdminIds.value = '';
+};
+
+const resetNotificationAndNewRowTag = () => {
     resetNotification();
+    resetNewTag();
+};
```
ianliuwk1019, line 96: This seems to be the main control of reset, maybe just call it `reset`? Just wondering...

frontend/src/components/managePermissions/ManagePermissions.vue
ianliuwk1019: This is interesting, I am curious, is a `push` actually fetch again?

frontend/src/components/grantaccess/GrantAccess.vue
```diff
@@ -160,8 +164,19 @@ const handleSubmit = async () => {
...
+            },
+        });
+    } else {
+        router.push(`/${routeItems.dashboard.name}`);
+    }
```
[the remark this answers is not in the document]
MCatherine1994: yes, when there is duplication

frontend/src/router/index.ts
MCatherine1994: checkout to the main branch, when run it local got the same error, but FAM-DEV which runs on main branch works fine. So will ignore the local case.

frontend/src/components/managePermissions/ManagePermissions.vue
MCatherine1994: push for just query will not trigger the fetch again by trying locally

frontend/src/components/managePermissions/ManagePermissions.vue
MCatherine1994: yeah I agree, updated to use Props

frontend/src/services/fetchData.ts
```diff
@@ -56,24 +66,34 @@ export const deleteAndRefreshUserRoleAssignments = async (
...
-        const userNameCompare = first.user.user_name.localeCompare(
-            second.user.user_name
-        );
-
-        return userNameCompare;
-    });
-
-    return applicationAdmins;
+    const sortedApplicationAdmins = applicationAdmins
+        .slice()
```
ianliuwk1019, line 79: No need to use slice().

frontend/src/services/fetchData.ts
```diff
@@ -92,30 +112,40 @@ export const deleteAndRefreshApplicationAdmin = async (
...
-            second.user.user_name
-        );
-
-        return userNameCompare;
-    });
-
-    return delegatedAdmins;
-
+    const sortedDelegatedAdmins = delegatedAdmins
+        .slice()
```
ianliuwk1019, line 131: No need to use slice().

ianliuwk1019 (approved):
Hi @MCatherine1994 , thank you for doing the changes in details and showing me that empty param does not trigger re-fetching!
Just one very minor comment, otherwise I approve for it. I remember you also want to adjust hover coloring. Let me know then I can approve again.

frontend/src/services/fetchData.ts
```diff
@@ -9,33 +9,43 @@ import {
...
-        );
-        // By role_name
-        const roleNameCompare = first.role.role_name.localeCompare(
-            second.role.role_name
-        );
-        return userNameCompare != 0 ? userNameCompare : roleNameCompare;
-    });
-    return userRoleAssignments;
+    const sortedUserAccesses = userRoleAssignments
+        .slice()
```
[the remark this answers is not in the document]
J0taFerreira: Hello, Ian! "slice()" is not needed indeed but "slice().sort()" is a good way of sorting without mutating the original array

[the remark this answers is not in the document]
MCatherine1994: I think the only difference is if we want to change the original userRoleAssignments array. `slice()` will not change the original array, I feel in our case we don't need to keep the original array but also doesn't matter.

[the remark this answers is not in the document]
ianliuwk1019: Hi Jota,
I think, no "good" when there is no real purpose; it only is good when you need the original array.
This would add more confusion and reduce performance.

MCatherine1994:
Also fyi, noticed 2 high severity vulnerabilities, run the audit fix, so the package-lock.json has some changes which are for fixing 
Screen Shot 2024-06-19 at 10 19 07 AM

ianliuwk1019 (approved):
Looks good, approved, thanks Catherine!

Issue #1382, opened 2024-05-13
Store terms and condition acceptance record in database


MCatherine1994:
**Describe the task**
Create a new data table to store the acceptance record of terms and conditions for fam delegated admin bceid users. 

**Acceptance Criteria**
- [x] Create a new table, include columns user_id, terms and conditions version, accept date
- [x] Create an api to check if user needs to accept the terms and conditions, add basic security check that user is app admin or delegated admin; or add the check to the existing get my access api 
- [x] Create an api to store when user accept the terms and conditions, add basic security check that user is app admin or delegated admin
- [x] add database user permissions to insert, read from new table (insert only, no edits)
- [x] Regenerate client-code-gen
- [x] Add tests as appropriate

**Additional context**
- API is created under access management Lambda (lower privileged)
- will hold a design session for data model changes  -table and column naming
-  API modifications to enforce T&C acceptance under issue #1383
- T&C acceptance only applies to external delegated admins (BCeID)
- T&C version is hardcoded

MCatherine1994:
Another idea Ian suggested that we can create an validation method to check if user needs to accept the terms and conditions, add the check to the get my access api, and return false in the json. But we need to think about the get my access api is under admin management apis, but we want to reuse that in regular apis.

MCatherine1994:
I was thinking about to do user's terms and condition check with the `get_admin_user_access` api. 

If we add a validation method to check that when calling `get_admin_user_access` , currently this method will be called every time after user login to FAM. However, no matter we call the validation as router guard, or in the api cal:
- If user already accepts terms and conditions, return user's access directly
- If user needs to accept, we have to return something different than the current return to indicate that we need to popup the terms and conditions for the user. And once the user accept the terms and conditions, we have to call the `get_admin_user_access` api again to get user's access. 
By that said, I feel this doesn't simplify the apis, also will cause duplicate. 
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/e78c4879-a823-459b-8307-0d4a55be6896)

If we create a new api to check if user needs to accept terms and conditions, this is nothing related to the admin management apis. So every time when user login:
- check if user needs to accept terms and conditions
    - if hasn't accept show popup, when accept, call get my admin access api
    - if accepted already, call get my admin access api

MCatherine1994:
[...]

PR #1447, opened 2024-06-11
feat: #1382 Store terms and conditions acceptance record

MCatherine1994:
refs: #1382 

- Create a new table to store terms and conditions acceptance record
- Create an api to check if a user needs to accept terms and conditions
- Create an api to store the terms and conditions acceptance record

Note: Will create a separate pr for adding tests and regenerate client code

Screen Shot 2024-06-11 at 2 56 30 PM

We start by creating new apis, but will keep an eye on the performance in frontend, if it takes too much time, we could consider put this check into the existing get my access api


server/backend/api/app/main.py
```diff
@@ -98,31 +100,46 @@ def main():
...
+    router_idim_proxy.router,
+    prefix=apiPrefix + "/identity_search",
+    dependencies=[Depends(router_guards.authorize)],
+    tags=["IDIR/BCeID Proxy"],
+)
+app.include_router(
+    router_user_terms_conditions.router,
+    prefix=apiPrefix + "/user_terms_conditions",
+    dependencies=[Depends(router_guards.authorize)],
+    tags=["FAM User Terms and Conditions"],
```
MCatherine1994, line 130: Most of the changes in this file is just formatting, added router with basic authorization check

server/backend/api/app/models/model.py
```diff
@@ -667,8 +674,80 @@ class FamAccessControlPrivilege(Base):
...
+        String(100),
+        comment="The user or proxy account that created or last updated the record.",
+    )
+    update_date = Column(
+        TIMESTAMP(timezone=True, precision=6),
+        default=datetime.datetime.utcnow,
+        comment="The date and time the record was created or last updated.",
+    )
+
+    user = relationship("FamUser", back_populates="fam_user_terms_conditions")
```
MCatherine1994, line 753: I feel maybe we don't need the detail user information, so didn't put the lazy joined option

server/backend/api/app/routers/router_user_terms_conditions.py
```diff
@@ -0,0 +1,56 @@
...
+from api.app.schemas import Requester, FamUserTermsConditionsGet
+from api.app.crud import crud_user_terms_conditions
+from api.app.routers.router_guards import get_current_requester
+
+LOGGER = logging.getLogger(__name__)
+router = APIRouter()
+
+
+@router.get(
+    "/check",
```
MCatherine1994, line 15: I'm not sure how to name the path, we can rename it

server/flyway/sql/V51__add_user_terms_conditions_table.sql
```diff
@@ -0,0 +1,47 @@
...
+ALTER TABLE app_fam.fam_user_terms_conditions ADD CONSTRAINT fam_user_terms_conditions_pk PRIMARY KEY (user_terms_conditions_id)
+;
+ALTER TABLE app_fam.fam_user_terms_conditions ADD CONSTRAINT Reffam_user_terms_conditions_user
+    FOREIGN KEY (user_id)
+    REFERENCES app_fam.fam_user(user_id)
+;
+
+-- -- Add permision on 'fam_user_terms_conditions' table for app access management api db user
+GRANT SELECT, INSERT ON app_fam.fam_user_terms_conditions TO ${api_db_username}
+;
```
MCatherine1994, line 47: I'm really bad at naming, feel free to see if we want any better name, thanks so much!!

server/backend/api/app/routers/router_guards.py
```diff
@@ -331,6 +329,29 @@ async def internal_only_action(requester: Requester = Depends(get_current_reques
         )
 
 
+async def requester_is_external_delegated_admin(
```
MCatherine1994, line 332: I'm not sure if this needs to be a async method

server/backend/api/app/routers/router_user_terms_conditions.py
ianliuwk1019: This is an interesting one. 
I will use "**POST**", that's what most commonly seen for this kind of not reqular REST resource requests.

But for naming, I might opt for "**validate**" instead.
However, it might be good to follow some [convention](https://restfulapi.net/resource-naming/) (maybe this case has exception). At the end of the doc it mentioned it can be (POST having action=validate in the body):
image

Or google has its convention using (":") for "**[custom methods](https://cloud.google.com/apis/design/custom_methods)**" which it would look like 
"**POST /user_terms_conditions/user:validate**"
But I am not sure how FastAPI will have no issue with that or not, or simply it is just another endpoint handler as (/user_terms_conditions/user:validate --> `@router.post("user:validate")`)

Perhaps, most commonly seen is using (requestBody:  action=validate)

server/flyway/sql/V51__add_user_terms_conditions_table.sql
```diff
@@ -0,0 +1,47 @@
+-- Create fam_user_terms_conditions table
+CREATE TABLE IF NOT EXISTS app_fam.fam_user_terms_conditions
+(
+    user_terms_conditions_id        bigint          GENERATED BY DEFAULT AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    user_id                         bigint          NOT NULL,
+    version_id                      int             NOT NULL,
```
ianliuwk1019, line 6: I am thinking maybe column name just give it as "**version**".
With "**id**" it makes me confused that it is a system generated `id` or it is a foreign key to a real "**terms_and_conditions**" table's id, but we are not modeling that currently to store terms and condition text in backend.

However, I am wondering if the "**version**" column should be an "int" or "varchar(30)" type as I am wondering the format of version is simply just "1, 2, 3" or "1.0.1" or "v1.0.1", "FAM:DAMIN:V2024-05-19"or  something else. We didn't discuss that...
Perhaps @basilv knows the T&C does have version number(or string) format?

MCatherine1994: yeah I agree Ian, I was thinking to keep it simple for now, but that's a good point, we can ask tomorrow. I'll change it to "version" first, or "terms_conditions_version"?

server/backend/api/app/routers/router_user_terms_conditions.py
MCatherine1994: so maybe POST with `/user_terms_conditions/{version_id}/verify-user` ?

server/flyway/sql/V51__add_user_terms_conditions_table.sql
basilv: I'd say use string (varchar) for version as data type. Agreed this should be called version instead of version_id.
We will just use version = "1" initially.

server/backend/api/app/routers/router_user_terms_conditions.py
MCatherine1994: Screen Shot 2024-06-12 at 1 58 07 PM

server/backend/api/app/models/model.py
ianliuwk1019: Yes, thanks. I think we want "lazy", the default is  lazy='select'. I am still confused with `lazy='joined'` for existing objects, but I think they are "eager" loading so removing that make sense to me.

server/backend/api/app/routers/router_user_terms_conditions.py
MCatherine1994: maybe this way? 

Screen Shot 2024-06-12 at 2 21 57 PM

ianliuwk1019 (changes requested):

server/backend/api/app/routers/router_user_terms_conditions.py
ianliuwk1019: This is good I feel.

server/backend/api/app/crud/crud_user_terms_conditions.py
```diff
@@ -0,0 +1,42 @@
...
+    db: Session, user_id: int, version: str
+) -> bool:
+    return (
+        False
+        if db.query(FamUserTermsConditions)
+        .filter(
+            FamUserTermsConditions.user_id == user_id,
+            FamUserTermsConditions.version == version,
+        )
+        .one_or_none()
```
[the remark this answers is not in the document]
MCatherine1994: if found one, it will return "False", means user has agreed previously

if found none, then it will return "True", means user needs to agree

this logic is doing what we want here?

server/flyway/sql/V51__add_user_terms_conditions_table.sql
```diff
@@ -0,0 +1,47 @@
+-- Create fam_user_terms_conditions table
+CREATE TABLE IF NOT EXISTS app_fam.fam_user_terms_conditions
+(
+    user_terms_conditions_id        bigint          GENERATED BY DEFAULT AS IDENTITY (START WITH 1 INCREMENT BY 1),
+    user_id                         bigint          NOT NULL,
+    version                         varchar(30)     NOT NULL,
+    create_user                     varchar(100)    NOT NULL,
+    create_date                     timestamp(6)    DEFAULT CURRENT_DATE NOT NULL,
```
MCatherine1994, line 8: I think we can use create date as acceptance date?

server/backend/api/app/crud/crud_user_terms_conditions.py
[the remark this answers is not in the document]
ianliuwk1019: Ah ha, I see what I was confused at.
I suggest in this case don't use compact form of the python condition statement with query, it looks confusing.

server/backend/api/app/routers/router_user_terms_conditions.py
```diff
@@ -0,0 +1,70 @@
...
+    LOGGER.debug(
+        f"Check if user {requester.user_id} needs to accept terms and conditions of version {version}"
+    )
+
+    if is_external_delegated_admin:
+        return crud_user_terms_conditions.require_accept_terms_and_conditions(
+            db, requester.user_id, version
+        )
+    else:
+        return False
```
ianliuwk1019, line 45: Hi Catherine, sorry, just realized that I think this business condition make more sense to be in crud method (`def require_accept_terms_and_conditon`). Maybe we can discuss that tomorrow.

server/flyway/sql/V51__add_user_terms_conditions_table.sql
ianliuwk1019: I feel the same, but probably there is no current need.

server/backend/api/app/routers/router_user_terms_conditions.py
MCatherine1994: yeah I was thinking of that before, like previously we'll put it into service not router, I'm confused with crud method hhh, will update it.

server/backend/api/app/crud/crud_user_terms_conditions.py
```diff
@@ -0,0 +1,64 @@
...
+def create_user_terms_conditions(
+    db: Session, user_id: int, version: str, requester: str
+) -> FamUserTermsConditions:
+    LOGGER.debug(
+        f"Creating user terms conditions acceptance record for user {user_id} and version {version}"
+    )
+
+    if get_user_terms_conditions_by_user_id_and_version(db, user_id, version):
+        error_msg = "User already accepted terms and conditions."
+        raise_http_exception(status_code=HTTPStatus.CONFLICT, error_msg=error_msg)
```
MCatherine1994, line 52: Hi Ian, SonarCloud is complaining the type of the status code, maybe the type of the status_code in the method should be a number instead of HTTPStatus? 
Screen Shot 2024-06-13 at 11 07 31 AM
image

server/backend/api/app/crud/crud_utils.py
```diff
@@ -87,12 +94,23 @@ def is_app_admin(
...
+def is_requester_external_delegated_admin(
+    db: Session,
+    requester: Requester,
+):
+    if (
+        requester.user_type_code == UserType.BCEID
+        and crud_access_control_privilege.is_delegated_admin(db, requester.user_id)
+    ):
+        return True
+    return False
```
ianliuwk1019, line 116: I was thinking this method could be in `crud_access_control_privilege` but it is fine to be here I think.

server/backend/api/app/crud/crud_user_terms_conditions.py
ianliuwk1019: As discussed, although I looked into it but don't find how it can be incorrect type. I am confused with the Sonar report as well. The one thing I am thinking of is maybe to do an experiment on adding third (default) argument in (with the order of the function) and see if that Sonar issue disappear... but I am good to ignore it if you like...

ianliuwk1019 (dismissed):

ianliuwk1019 (approved):

Issue #1454, opened 2024-06-17
Security Vulnerability ("python-jose") Fixing Part 2


ianliuwk1019:
**Describe the bug**
On ticket #1398 , we tried to address most dependabot security vulnerability in FAM code by replacing the "**python-jose**" library with "**PyJwt**" library and "**authlib**".
However, there is a file "**bcsc_decryption.py**" using "**python-jose**" has complexity for replacing it with other library thus decided to separate the fixing into different ticket. 
In short, what "**bcsc_decryption.py**" do is to decrypt the "**jwe**" token getting from BCSC (from the BCSC's userinfo url) by using the AWS "decrypted_key" (**symmetric public key**). 

In fact, the file "**bcsc_decryption.py**" seems to be a copy-paste to FAM from 'python-jose' library itself (due to some reason unknown) and not sure if there is any custom adjustment for FAM.
The intended attempt to fixing the security vulnerability is as below:
- remove the "bcsc_decryption.py".
- use library "**authlib**" to decrypt the jwe token instead.
![image](https://github.com/bcgov/nr-forests-access-management/assets/81595625/3102c1c1-5aa1-487f-83ff-507e70a053a5)
  
**Acceptance Criteria**
- [ ] Replace "**router_bcsc_proxy.py**" on decrypt jwe part with code shown above screenshot and remove dependencies from "bcsc_decryption.py".
- [ ] Remove the need for using **python-jose** library, instead, using "**authlib**" library (FAM already has this installed as python package).
- [ ] Understand [BCSC custom proxy auth flow works](https://github.com/bcgov/nr-forests-access-management/wiki/BCSC-ENcryption) and test new BCSC flow at **Tools/Dev** space to make sure it works (most likely requires changing Cognito configuration for **BCSC IDP provider** on several endpoints and also need to have IDIM Welsy's help on reconfigure test environment on their side for callback).
- [ ] Once testing work, re-configure and re-test again when code is deployed to **TEST** environment.

**Additional context**
Might need to contact Wesley wesley.gilbert@gov.bc.ca for dev configuration so we can test the BCSC login in DEV

ianliuwk1019:
Questions to Wesley regarding to the bcsc proxy integration testing with BCSC (use FAM DEV environment with BCSC TEST instance).
The reply from Wesley:
```
Hi Ian,
 
[...]

PR #1455, opened 2024-06-17
fix: 1398 resolve python jose security vulnerability

ianliuwk1019:
admin_management:
	- use pyjwt and remove python-jose dependency.
	- fix code impacted by the library replacement on jwt.
	- fix tests.

backend:
	- use pyjwt for auth token dependency, still keeping python-jose for BCSC part(deal with it in second ticket).
	- fix code impacted by the library replacement on jwt.
	- fix tests.

Created #1454 to part 2 on replacing "python-jose" for BCSC.


server/backend/api/app/routers/router_bcsc_proxy.py
```diff
@@ -113,8 +114,28 @@ def bcsc_userinfo(request: Request, bcsc_userinfo_uri):
...
+    Will use the new ticket to add below code and validate decryption correctness.
+
+    # from authlib.jose import JsonWebEncryption
+    #
+    # jwe = JsonWebEncryption()
+    # decrypted_id_token = jwe.deserialize_compact(
+    #     jwe_token, decrypted_key
+    # )['payload']
+    """
+
```
ianliuwk1019, line 136: Comment is added above due the complexity of replacing "**python-jose**" in `bcsc_decryption.decrypt()`.
Leave it for another ticket to address this part.

server/backend/api/app/routers/router_bcsc_proxy.py
```diff
@@ -113,8 +114,28 @@ def bcsc_userinfo(request: Request, bcsc_userinfo_uri):
...
     decoded_id_token = jwt.decode(
-        decrypted_id_token, None, options={"verify_signature": False, "verify_aud": False}
+        decrypted_id_token, None, options={"verify_signature": False}
```
ianliuwk1019, line 138: This is the only part that gets changed in this file as I do not feel it requires the other option.

basilv (approved):

server/admin_management/tests/authorization/test_jwt_validation.py
```diff
@@ -155,4 +157,4 @@ def test_get_application_invalid_signature_failure(
 
     response = test_client_fixture_unit.get(f"{endPoint}", headers=headers(token))
 
-    assert_error_response(response, 401, ERROR_VALIDATION)
+    assert_error_response(response, 401, ERROR_CLAIMS)
```
MCatherine1994, line 160: Hi Ian, just want to double check this change is what we expected?

ianliuwk1019: Glad you asked this one.
 
Before replacing with library and after replacing with new library I ran whole test sets several times, this is one of the tests that were failed and needed to be adjusted.
In fact, this is **the one** test that's giving me confusion as well.
If we checkout '**main**' branch, run the same "set" of tests (under admin_management), they all pass. If we run **individual** test (::test_get_application_invalid_signature_failure) it will fail. (the same running outcomes for after I replace to the new library).

**To minimize the impact,** the "**ERROR_CLAIMS**" is what's expected from test failure. But, actually **I would prefer removing this test** (I didn't do it), as the test result is not consistent and I feel the initial setup (in 'main' branch) probably was not even correct; but the JWT validation is complicated.

What I thought is, when the **"jwt_validation.py"** initially was written should not have used try-catch many exceptions, some of the validations in jwt_validation.py were(are) not necessary; it should just relies on the library's validation to throw its meaningful/detail exception. However, for library's throwing exception back to our backend FastAPI, then FastAPI probably won't be able to handle it and crush the backend. We should just had a general try-catch and return known "HTTPException" that FastAPI is able to handle and return the response.
However, to change this will change quite some codes (and tests) and will impact a lot. 
So at the end, for minimize the impact of this pr, the "**ERROR_CLAIMS**" is what expected for now.

Or do you feel you also prefer to remove this test, or change the code (in **"jwt_validation.py"**) altogether (it is probably better for future)?
@basilv @MCatherine1994

MCatherine1994: Hi Ian, thanks for the explanation!!

MCatherine1994 (approved):

Issue #1425, opened 2024-06-04
Backend add security check if user already accept the terms


MCatherine1994:
**Describe the task**
Add security check to backend apis that will be used by BCeID delegated admin, to check if the user already accept the terms and conditions

**Acceptance Criteria**
- [ ] Update backend apis (the ones bceid delegated admin will use), check if user already accept the terms and conditions
- [ ] Add or fix tests 

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
For apis to check for (BCeID) T&C the check will need to know which "**version**" of T&C the api requesting user currently accepts.
Last time during discussion it seems to me the T&C will be saved at frontend;
However, if we want every related endpoint to pass the "**version**" from **frontend**, it requires to include the T&C version as **new parameter in request** for every related endpoints. It will have a broader impact on these endpoints to adjust.
Do we want that?

Or, if we don't save the T&C (and its version at backend), we could define a T&C version in a "**business constant**" (**in code**, not in db), and that will be the **most recent "version"** to rely on. However, the drawback (small) is when there is T&C update at **frontend**, we also need to update this "constant" at the **backend**. 
Is it acceptable for this design question?
@basilv @MCatherine1994

basilv:
The backend can hardcode the current T&C version and check if the version the user accepted is the current version.

ianliuwk1019:
Hi @basilv , as discussed, for now I can enforce the T&C validation only on **POST/PUT** endpoints and temporarily ignore the GET (and the search [on user/client-number] endpoints.

For your concerns on every endpoint that (if applicable) goes extra database trip(s) to validate the T&C, you are correct, it does go extra trip(or trips) to validate T&C. (For a search, or multiple search, or **predictive search**, they are already very slow to the idim service, and we probably don't want **extra 1 or 2 db session validation checks** again each time)

[ Current implementation for validating user on T&C is: checks if user is "**external (BCeID) delegated admin**" (that's 1 db trip to `fam_access_control_privilege` table), if yes, then checks if user has T&C record (that's 2nd db trip to `fam_user_terms_conditions` table) ]

[...]

PR #1463, opened 2024-06-24
feat: #1425 add backend endpoint checks for bceid user accept terms and conditions

ianliuwk1019:
- Coded CURRENT_TERMS_AND_CONDITIONS_VERSION at backend as version "1" and use it in apis and checks.
- Added 'is_delegated_admin' and 'requires_accept_tc' as custom attributes in Requester schema.
- RouterGuard adjustment for Requester for joined loading information from 'fam_access_control_privilege' and 'fam_user_terms_conditions' 
- Added 'enforce_bceid_terms_conditions_guard' for backend api endpoints.
- Fixed affected tests due to new 'enforce_bceid_terms_conditions_guard' checks for endpoints. 
- Generated new api client.


server/backend/api/app/routers/router_guards.py
```diff
@@ -50,38 +41,63 @@ async def get_current_requester(
...
-    )
+) -> Requester:
     LOGGER.debug(
-        f"Debug router_guard get_current_requester, current db connection: {db}"
+        f"Retrieving current requester from: request_cognito_user_id: {request_cognito_user_id}"
     )
-    fam_user: FamUser = crud_user.get_user_by_cognito_user_id(
+    fam_user: FamUser = crud_user.fetch_initial_requester_info(
         db, request_cognito_user_id
     )
```
ianliuwk1019, line 50: The 'fetch_initial_requester_info' method uses programmatic SqlAlehcmy ORM joined to get other information (delegated_admin, T&C) for this user.

server/backend/api/app/routers/router_guards.py
```diff
@@ -50,38 +41,63 @@ async def get_current_requester(
...
+        and not has_current_terms_conditions_accepted
     )
-    requester.access_roles = access_roles
-    LOGGER.debug(f"Current request user (requester): {requester}")
-    return requester
+
+    return {
+        "is_delegated_admin": is_delegated_admin,
+        "requires_accept_tc": requires_accept_tc
+    }
```
ianliuwk1019, line 96: New custom attributes for "Requester" schema. The information is parsed from previous joined tables information from specific user.

server/backend/api/app/routers/router_guards.py
```diff
@@ -414,3 +427,13 @@ async def enforce_bceid_by_same_org_guard(
...
+
+
+def enforce_bceid_terms_conditions_guard(
+    requester: Requester = Depends(get_current_requester),
+):
+    if requester.requires_accept_tc:
+        utils.raise_http_exception(
+            error_code=ERROR_CODE_TERMS_CONDITIONS_REQUIRED,
+            error_msg="Requires to accept terms and conditions.",
+        )
```
ianliuwk1019, line 439: The router_guard checks for T&C for use in backend api endpoints for BCeID delegated admin user.
Exception has default http error status: 400.

server/backend/testspg/router/test_router_application.py
```diff
@@ -419,6 +417,9 @@ def test_get_user_role_assignments_filtering_for_bceid_delegated_admin(
         fom_dev_access_admin_token, access_grants_not_able_to_see
     )
 
+    # override router guard dependencies
+    override_enforce_bceid_terms_conditions_guard()
```
ianliuwk1019, line 421: Fix broken tests that is for BCeID delegated admin requester; `override_enforce_bceid_terms_conditions_guard` defaults to pass the check.

server/backend/api/app/models/model.py
```diff
@@ -192,6 +184,150 @@ class FamUserType(Base):
...
+        String(100),
+        comment="The user or proxy account that created or last updated the record.",
+    )
+    update_date = Column(
+        TIMESTAMP(timezone=True, precision=6),
+        default=datetime.datetime.utcnow,
+        comment="The date and time the record was created or last updated.",
+    )
+
+    user = relationship("FamUser", back_populates="fam_user_terms_conditions")
```
MCatherine1994, line 328: Hi Ian, so here it's actually no change for the model, just change the place in the file for these two models I believe?

server/backend/api/app/models/model.py
```diff
@@ -240,13 +376,13 @@ class FamUser(Base):
     )
 
     fam_user_role_xref = relationship("FamUserRoleXref", back_populates="user")
-    user_type_relation = relationship(
+    user_type_relation: Mapped[FamUserType] = relationship(
```
MCatherine1994, line 379: just curious, do we need to change the user_type_relation?

server/backend/api/app/models/model.py
ianliuwk1019: Ah, yes, no change on other places; some parts are just reorganized, I should put some comments on this one.

server/backend/api/app/models/model.py
ianliuwk1019: We don't need to change this one. I think for some reason before I touched it and forgot to change it back (but does not hurt to have it). I revert it back.

MCatherine1994:
Hi Ian, it looks good, thanks so much! Just one thing, I wonder since we already have the requester.is_delegated_admin, maybe we don't need this check anymore, to check if is external delegated admin, we could just use `requester.user_type_code` and  `requester.is_delegated_admin`? so we can delete that `is_requester_external_delegated_admin` method
Screen Shot 2024-06-25 at 3 04 01 PM

ianliuwk1019:
> Hi Ian, it looks good, thanks so much! Just one thing, I wonder since we already have the requester.is_delegated_admin, maybe we don't need this check anymore, to check if is external delegated admin, we could just use `requester.user_type_code` and `requester.is_delegated_admin`? so we can delete that `is_requester_external_delegated_admin` method Screen Shot 2024-06-25 at 3 04 01 PM

Thanks for the find! I missed that one. Will do the adjustment.

MCatherine1994 (approved):

client-code-gen/gen/app-access-control-api/api/famuser-terms-and-conditions-api.ts
```diff
@@ -92,10 +84,6 @@ export const FAMUserTermsAndConditionsApiAxiosParamCreator = function (configura
             // oauth required
             await setOAuthToObject(localVarHeaderParameter, "6jfveou69mgford233or30hmta", [], configuration)
```
basilv, line 85: Should this "6jfv..." value be a constant? What is this for?

ianliuwk1019: Hi Basil, this file is one of the generated frontend api-client code, it is internal to the generated code that we don't control it and don't directly use it frontend or backend. 
The value "**6jfv**..." by itself looks like is one of the environment variable ("**COGNITO_CLIENT_ID**"), the Cognito client for FAM. How it gets into generated code is unknown to me, but it seems the OpenAPI/Swagger is smart enough to grab that when server is being served up, and it seems Swagger uses this value to describe the "**securitySchemes**".
The value previously already exists and did not appear to cause any issue. This is the "openapi.json" that contains the value:
image

basilv: Ah, okay, I missed that this was a generated file when I was reviewing, that makes more sense.

Issue #1326, opened 2024-04-23
Load testing - Concurrent users - phase 3


gormless87:
**Describe the task**
Test concurrent logins for downstream apps (50)

**Acceptance Criteria**
- [ ] Determine tool to use to generate concurrent load
- [ ] Measure login performance under concurrent load, compare against performance requirements
- [ ] Evaluate failures/bottlenecks

**Additional context**
- Follow up to  #1321 
- Separate tickets to address any issues discovered

ianliuwk1019:
Hi @Basil, your initial load data script can be loaded successfully into my local postgres database.
As initially planned, we might want to load these **_LOAD_TESTER_** data into AWS TEST environment, but currently we don't have a simple way to run only the load-testing migration into specific environment (AWS TEST).
Current local work because the docker file copy migration scripts form two different locations for flyway.

I take a look and see that our flyway scripts are uploaded to AWS S3 Bucket, maybe, if we could find a way to load the script onto AWS S3 Bucket based on environment, then we can have a way to execute load-testing migration ONLY on AWS TEST environment.
![Image](https://github.com/bcgov/nr-forests-access-management/assets/81595625/7c7bbb9e-cd1c-4ffa-a294-03984ecd3f1f)

I also see Flyway documentation has "FLYWAY_LOCATIONS" environment option that can specify multiple locations of the scripts, its another possibility:
![Image](https://github.com/bcgov/nr-forests-access-management/assets/81595625/8607edfa-c5ca-4b77-a409-1eb9446ccb78)


For this ticket, do you want to do that or do you think it's better create another ticket to investigate the how-to and work on making flyway migration for specific environment possible for our FAM workspace, and leave this ticket to just have your load-testing sql script done and ready?
@MCatherine1994

OlgaLiber2:
Include @webgismd in our next meeting and when we are discussing tools to use.
@basilv @gormless87

OlgaLiber2:
no longer relevant

PR #1465, opened 2024-06-25
chore: #1326 Data generation script for load testing

basilv:
Generate ~1 year of fake data for load testing. Tested loading into database but haven't tested with running application yet.


ianliuwk1019:
Hi @basilv ,
Not a comment, just FYI.
I took your data.sql and placed them into `flyway/local_sql` folder to run migration. Despite initial warning, it was run successfully.
image
With fam_user table loaded with:
image

basilv:
Sounds like there were no performance issues running locally, so I think this PR is ready for review. Next step after review is likely to load this data into test environment.

MCatherine1994:
Hi Basil, added the Docker file for admin management, also added to the docker-compose file. I was running `docker-compose up` on my local, and get both backend api, and admin management api work fine. If we are still seeing issues with them, I'll take another look. Thanks!

ianliuwk1019 (approved):
Hi Basil, just wondering, after initial load-test migration in TEST environment all sequences restart at higher number from this `RESTART WITH 100000` which it should have no consequence I think. If later we need some other large load-test data (especially for role assignment xref) I guess it should be fine to insert data in between (< 100000).

basilv:
Yes, the 'design' of the data generation load allows us to change it in the future and re-run it without affecting manaully-created data.

basilv:
@ianliuwk1019 @MCatherine1994 This is ready for final review.

ianliuwk1019:
Thank you so much @basilv for the comprehensive load-testing and detail procedures at readme!!
Since you have the `.js` script to regenerate data, wondering do we still need the generated `data.sql` in repo, it seems the size is quite large to checkout.

basilv:
Yes, we need the data.sql in the repo for pulling it onto the ec2 instance for actually running the script against the database.

ianliuwk1019 (approved):
Thank you so much Basil!!

Issue #1464, opened 2024-06-25
Hide delegated admin features in TEST and PROD env


MCatherine1994:
**Describe the task**
Because we're not ready to deploy the delegated admin feature to production, but also don't want to block our regular releases and prod deployment. We want to disable the delegated admin features in the test and prod env. 
There is a "target_env" variable in frontend, and we will disable the bceid login, and hide the delegated tab, hide the delegated admin options from left navigation bar.

**Acceptance Criteria**
- [x] Disable the bceid login, hide the delegated admin tab, hide the delegated admin options from left navigation bar based on environment
- [x] Update tests if need

**Additional context**
- Add any other context about the task here.
- Or here

PR #1467, opened 2024-06-27
fix: # 1464 Hide delegated admin features for non-dev environments

MCatherine1994:
refs: #1464

Hide delegated admin features for TEST, TOOLS and PROD environments, as we're not ready to deploy that for user to see yet.


ianliuwk1019 (approved):
Looks good, thank you Catherine!!

Issue #1459, opened 2024-06-20
Support SILVA: add a redirect url for SILVA DEV


MCatherine1994:
**Describe the task**
Ricardo requests to add a new redirect url for SILVA DEV, and they want to keep all the existing ones

**Acceptance Criteria**
- [ ] Add a new redirect login url for SILVA DEV `http://localhost:4173/dashboard` in the terraform config for SILVA
- [ ] Add a new redirect logout url for SILVA DEV `http://localhost:4173/` with the logout chain in the terraform config for SILVA

**Additional context**
- Add manually for now, but still need to add in our code for next release
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/4f5de404-9704-4eea-91e1-7708fa602ed5)

Issue #1461, opened 2024-06-24
Support Spar: update redirect vanity urls for test and prod


MCatherine1994:
**Describe the task**
Update the redirect urls for SPAR TEST and SPAR PROD

**Acceptance Criteria**
- [x] Update the vanity redirect login url for test to be https://spar-tst.nrs.gov.bc.ca/, as well as the logout url with logout chain
- [x] Update the vanity redirect login url for prod to be https://spar.nrs.gov.bc.ca/, as well as the logout url with logout chain

**Additional context**
- Keep the existing ones as well
- Add manually for now, but still need to add in our code for next release
![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/71ce0866-fddf-4c4f-b9d2-76972876fe94)

PR #1468, opened 2024-06-28
fix: #1459 #1461 Add redirect urls for SPAR and SILVA

MCatherine1994:
refs: #1459, #1461

### DIFF
--- a/infrastructure/server/oidc_clients_spar.tf
+++ b/infrastructure/server/oidc_clients_spar.tf
@@ -43,9 +43,11 @@ resource "aws_cognito_user_pool_client" "test_spar_oidc_client" {
   callback_urls                                 = concat([
     "http://localhost:3000/",
     "https://oidcdebugggersecure-c6af30-dev.apps.gold.devops.gov.bc.ca/",
-    "https://nr-spar-test-frontend.apps.silver.devops.gov.bc.ca/"
+    "https://nr-spar-test-frontend.apps.silver.devops.gov.bc.ca/",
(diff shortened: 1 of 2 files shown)


infrastructure/server/oidc_clients_spar.tf
```diff
@@ -43,9 +43,11 @@ resource "aws_cognito_user_pool_client" "test_spar_oidc_client" {
   callback_urls                                 = concat([
     "http://localhost:3000/",
     "https://oidcdebugggersecure-c6af30-dev.apps.gold.devops.gov.bc.ca/",
-    "https://nr-spar-test-frontend.apps.silver.devops.gov.bc.ca/"
+    "https://nr-spar-test-frontend.apps.silver.devops.gov.bc.ca/",
```
ianliuwk1019, line 46: Wondering do they need to remove the old one?

ianliuwk1019: Never mind, seems the ticket mentioned to keep the old one for now.

ianliuwk1019 (approved):

infrastructure/server/oidc_clients_spar.tf
MCatherine1994: last time when they sent me the message they want to keep the old ones, cause they haven't finish transferring to the new urls. But I'll ask them again next week just to make sure they still need the old ones.

MCatherine1994:
confirmed with Craig, they want to keep the existing urls as well for a couple more weeks before their prod release

PR #1470, opened 2024-06-30
feat: #1425 backend check bceid user accept tc tests

ianliuwk1019:
ref:1425
- tests for validation checking terms and conditions acceptance.


server/backend/api/app/routers/router_guards.py
```diff
@@ -37,7 +37,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-async def get_current_requester(
+def get_current_requester(
```
ianliuwk1019, line 40: I remove the "async" since there is no need.

MCatherine1994 (approved):

Issue #1365, opened 2024-05-08
Add first name, last name and email address to user table


MCatherine1994:
**Describe the task**
Add first name, last name and email address columns to user table

**Acceptance Criteria**
- [x] Add new columns for first name, last name and email in the fam_user table
- [x] determine string lengths
- [x] Add string length restriction to the new columns 
- [x] Update data model for admin management and backend

**Additional context**
- Add any other context about the task here.
- BCSC out of scope

MCatherine1994:
From the latest IDIM web service documentation, first name and last name will have maximum 50 chars, and email will have maximum 250 chars. 

![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/e310c273-106e-4bfe-95b1-b7c5323a9167)



![Image](https://github.com/bcgov/nr-forests-access-management/assets/23131735/974e0654-805b-4d05-aff8-681c9c9ca3ee)

PR #1472, opened 2024-07-02
feat: #1365 Add first, last name and email columns to fam user table

MCatherine1994:
refs: #1365

image
image


ianliuwk1019 (dismissed):
Thanks Catherine!

MCatherine1994:
Hi Ian, I wonder should we add some description to these columns? I saw some of the columns in the table have description, but some don't. Maybe these three are self-explainable?

ianliuwk1019:
> Hi Ian, I wonder should we add some description to these columns? I saw some of the columns in the table have description, but some don't. Maybe these three are self-explainable?

Good point, I forgot about that, although they are very obvious. I think in the past the team starts adding comments as a best practice. Thanks.

ianliuwk1019 (approved):

Issue #1383, opened 2024-05-13
Use user terms acceptance api for frontend


MCatherine1994:
**Describe the task**
A clear and concise description of what the task is.

**Acceptance Criteria**
- [x] Frontend call api to check if user already accept the terms and conditions when bceid delegated admin login
- [x] Frontend call api to store when user accept the terms and conditions

**Additional context**
- Current 'version' is stored at backend, don't need to pass 'version' to backend to check or store user's terms and conditions.
- Or here

MCatherine1994:
When user has no role in FAM, should stay on the dashboard page with an error message indicates that the login user has no access in FAM, but currently it logs user out. I'll take a look on this one.

PR #1473, opened 2024-07-03
feat: #1383 use user terms acceptance api on frontend

J0taFerreira:
- Frontend call api to check if user already accept the terms and conditions when bceid delegated admin
- Store user when the terms and conditions is accept
ref: #1383


frontend/src/components/common/TermsAndConditions.vue
```diff
@@ -6,6 +6,19 @@ import {
...
+import { AppActlApiService } from '@/services/ApiServiceFactory';
+import { setRouteToastError } from '@/store/ToastState';
+
+const acceptTermsAndConditions = async () => {
+    try {
+        await AppActlApiService.userTermsAndConditionsApi.createUserTermsAndConditions();
+        hideTerms();
+    } catch (error: any) {
+        // temporary error handling
+        setRouteToastError(error.message);
```
ianliuwk1019, line 18: Looks correct to me but just wondering is the toast error able to show up right after 'logout', have you tried?

frontend/src/store/FamLoginUserState.ts
```diff
@@ -266,8 +267,12 @@ const cacheUserAccess = async () => {
     }
 };
 
-// TODO: check if login user needs to accept terms and conditions, if need, show terms here
-const requiresAcceptTermsCondition = async () => {};
+const requiresAcceptTermsCondition = async () => {
+    const needsToAccept = await AppActlApiService.userTermsAndConditionsApi.validateUserRequiresAcceptTermsAndConditions();
+    if (needsToAccept.data) {
+        showTermsForAcceptance();
+    }
```
ianliuwk1019, line 274: Can we just return a true/false and call "**showTermsForAcceptance**()" in the same "**handlePostLogin**" method and not here in state file?

ianliuwk1019 (changes requested):

frontend/src/store/FamLoginUserState.ts
J0taFerreira: Hello! Sure can!

frontend/src/components/common/TermsAndConditions.vue
J0taFerreira: You mean show up in the landing page?

ianliuwk1019: Then please don't put the code there for others wondering if what you intended to do isn't working (if you already verify and know it isn't working) due to some reason. 
For now, please remove the toast error, only just logout.
Thank you.

J0taFerreira: It's working, if there is an error while accepting it will show a  toast, but the toast is not on the landing page

ianliuwk1019: I see, the transition might be too fast that user does not even have time to know what error it is.
Sure, then leave it, it is a limitation of current toast handling.
Thanks.

ianliuwk1019 (approved):

Issue #1456, opened 2024-06-18
Update role selection to use radio in table


MCatherine1994:
**Describe the task**
Update the role selection dropdown in Grant Access and Grant Delegated Admin form to become a table with radio selection and role description as in the new design. Start with single selection for now. 

**Acceptance Criteria**
- [x] Backend get my access api return role description
- [x] Regenerate client code for frontend
- [x] Frontend create a role table component 
- [x] Update the two forms to use role table for role selection
- [x] Update test if needs

**Additional context**
- Add any other context about the task here.
- Or here

https://www.figma.com/design/VuQBlZZ1PmGUlEBJriyzo5/Manage-permissions?node-id=2149-30454&t=n7wOvIiT2h9I2AVw-4

OlgaLiber2:
@NickSaglioni here's the link to figma https://www.figma.com/design/VuQBlZZ1PmGUlEBJriyzo5/Manage-permissions?node-id=2149-30454&t=n7wOvIiT2h9I2AVw-4

MCatherine1994:
Hi Nick @NickSaglioni , just fyi, we still need to keep the current role selection dropdown component, cause we'll use the dropdown later as well. Thanks!

NickSaglioni:
The `Checkbox` component used in Figma is not being used for now, replaced by the `RadioButton` component. This is because `Checkboxes` in definition accept multiple selections, which the application is not prepared yet. `RadioButton` only accepts one value, so it was used in this case. It should be replaced with Checkbox once multiple values could be used.

PR #1478, opened 2024-07-04
feat: #1456 update role selection to use radio in table

NickSaglioni:
- Backend get my access api return role description;
- Regenerate client code for frontend;
- Frontend create a role table component;
- Update the two forms to use role table for role selection;


MCatherine1994:
Did a little bit cleanup and styling, should be good now. Primevue does have built-in single select and multiple select in the data table component, but since our selection needs to trigger the forest client field, I feel it's okay to not use the built-in one, but write our own radio column. 

Screen Shot 2024-07-05 at 2 39 11 PM

MCatherine1994:
Screen Shot 2024-07-05 at 2 45 31 PM
Screen Shot 2024-07-05 at 2 45 51 PM

MCatherine1994:
Also we keep the role selection dropdown for now, cause I think we'll use it as well later for the full new design.

ianliuwk1019 (approved):
Fixed some code and styling. Approve this PR.

PR #1479, opened 2024-07-05
fix: #1383 Only check if user needs accept terms and conditions if is FAM user

MCatherine1994:
refs: #1383

Our backend protects our endpoints that only FAM user (has at least one delegated admin or admin role in FAM) can call, otherwise will return an authorization error. So update frontend to only check if user needs accept terms and conditions if the login user is FAM user (has access in FAM).

### DIFF
--- a/frontend/src/services/AuthService.ts
+++ b/frontend/src/services/AuthService.ts
@@ -45,7 +45,15 @@ const handlePostLogin = async () => {
         // This is to update the FamLoginUser for FamLoginUser.accesses.
         // For now team decided to grab user's access only when user login and may change later.
         await LoginUserState.cacheUserAccess();
-        await LoginUserState.requiresAcceptTermsCondition() && showTermsForAcceptance()
+
+        if (
+            LoginUserState.state.value.famLoginUser?.accesses &&
+            LoginUserState.state.value.famLoginUser.accesses.length > 0
+        ) {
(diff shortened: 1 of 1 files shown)


frontend/src/services/AuthService.ts
```diff
@@ -45,7 +45,15 @@ const handlePostLogin = async () => {
         // This is to update the FamLoginUser for FamLoginUser.accesses.
         // For now team decided to grab user's access only when user login and may change later.
         await LoginUserState.cacheUserAccess();
-        await LoginUserState.requiresAcceptTermsCondition() && showTermsForAcceptance()
+
+        if (
+            LoginUserState.state.value.famLoginUser?.accesses &&
+            LoginUserState.state.value.famLoginUser.accesses.length > 0
+        ) {
```
ianliuwk1019, line 52: Hi Catherine, just curious, what endpoint call and what error was it from backend before this condition is added?
Is it the "validate_user_requires_accept_terms_and_conditions" endpoint?

MCatherine1994: yes, and that's just because of a general security check, that only FAM user (at least one role in FAM) can call them  Screen Shot 2024-07-05 at 2 50 28 PM

ianliuwk1019 (approved):

Issue #1366, opened 2024-05-08
Store first name, last name, and email when create user


MCatherine1994:
**Describe the task**
When grant application admin, delegated admin, and user access, verify user by user_guid (by calling IDIM web service), and get first name, last name and email and store them. 

**Acceptance Criteria**
- [x] modify backend logic where user_guid is verified
- [x] store in pg-db
- [x] tests updated

**Additional context**
- Add any other context about the task here.
- IDIR and Business BCeID in scope

PR #1483, opened 2024-07-15
feat: #1366 Store first name, last name, and email when create user

ianliuwk1019:
* Add first_name, last_name, email fields to Schema TargetUser class to store searched result.
* Update router, crud and service code to use TargetUser to update first_name, last_name and email
  when creating application admin, delegated admin and user permission.
* Rename UserValidator to TargetUserValidator and add storing first_name, last_name and email after validating.
* Update tests and add new tests.


server/admin_management/api/app/routers/router_guards.py
```diff
@@ -190,8 +175,8 @@ async def get_verified_target_user(
     """
     Validate the target user by calling IDIM web service, and update business Guid for the found BCeID user
     """
-    user_validator = UserValidator(requester, target_user)
-    return user_validator.verify_user_exist()
+    target_user_validator = TargetUserValidator(requester, target_user)
+    return target_user_validator.verify_user_exist()
```
ianliuwk1019, line 179: No change. Just rename the class.

server/admin_management/api/app/services/user_service.py
```diff
@@ -94,28 +94,50 @@ def update_user_name(
             return self.user_repo.get_user(user.user_id)
         return user
 
-    def update_user_business_guid(
-        self, user_id: int, business_guid: str, requester: str  # cognito_user_id
+    def update_user_properties_from_verified_target_user(
+        self, user_id: int, target_user: TargetUser, requester: str  # cognito_user_id
     ):
```
ianliuwk1019, line 99: Hope the naming is good.
Previous "business_guid" update is combined.

server/admin_management/api/app/services/validator/target_user_validator.py
```diff
@@ -75,13 +67,16 @@ def verify_user_exist(self) -> TargetUser:
...
-            if search_result.get("found") and search_result.get("businessGuid"):
-                self.verified_target_user.business_guid = search_result.get(
-                    "businessGuid"
-                )
+        # Update various target_user fields from idim search if exists
+        if search_result and search_result.get("found"):
+            self.verified_target_user.business_guid = search_result.get("businessGuid")
+            self.verified_target_user.email = search_result.get("email")
+            self.verified_target_user.first_name = search_result.get("firstName")
+            self.verified_target_user.last_name = search_result.get("lastName")
```
ianliuwk1019, line 75: "firstName", "lastName" and "email" are set to target_user after search.

server/admin_management/tests/services/test_user_validator.py
```diff
@@ -65,8 +61,8 @@ def setup_class(self):
     def test_verify_user_exist_idir(self, mock_search_idir):
         mock_search_idir.return_value = MOCK_SERACH_IDIR_RETURN
         target_user = TargetUser(**TEST_NEW_TARGET_USER_IDIR)
-        user_validaor = UserValidator(self.requester_idir, target_user)
-        verified_target_user = user_validaor.verify_user_exist()
+        target_user_validaor = TargetUserValidator(self.requester_idir, target_user)
+        verified_target_user = target_user_validaor.verify_user_exist()
```
ianliuwk1019, line 65: No change on this file, just rename some variables.

server/backend/api/app/routers/router_guards.py
```diff
@@ -386,8 +386,8 @@ async def get_verified_target_user(
     """
     Validate the target user by calling IDIM web service, and update business Guid for the found BCeID user
     """
-    user_validator = UserValidator(requester, target_user)
-    return user_validator.verify_user_exist()
+    target_user_validator = TargetUserValidator(requester, target_user)
+    return target_user_validator.verify_user_exist()
```
ianliuwk1019, line 390: No change, just some variable renamed.

server/backend/testspg/crud/test_user_validator.py
```diff
@@ -51,8 +48,8 @@ def setup_class(self):
     def test_verify_user_exist_idir(self, mock_search_idir):
         mock_search_idir.return_value = MOCK_SERACH_IDIR_RETURN
         target_user = TargetUser(**TEST_IDIR_REQUESTER_DICT)
-        user_validaor = UserValidator(self.requester_idir, target_user)
-        verified_target_user = user_validaor.verify_user_exist()
+        target_user_validaor = TargetUserValidator(self.requester_idir, target_user)
+        verified_target_user = target_user_validaor.verify_user_exist()
```
ianliuwk1019, line 52: No change to this file, just some variable renamed.

server/admin_management/api/app/services/application_admin_service.py
```diff
@@ -40,6 +42,9 @@ def create_application_admin(
         fam_user = self.user_service.find_or_create(
             request.user_type_code, request.user_name, request.user_guid, requester
         )
+        fam_user = self.user_service.update_user_properties_from_verified_target_user(
+            fam_user.user_id, target_user, requester
```
MCatherine1994, line 46: Hi Ian, I wonder since we'll always add first name, last name and email when create user, should we add those three properties in the find_or_create method? and leave the update business_guid method?

server/admin_management/api/app/schemas.py
```diff
@@ -69,6 +70,9 @@ class TargetUser(Requester):
     checking "is_new_user()".
     """
     user_id: Optional[int] = None
+    first_name: Optional[Annotated[str, StringConstraints(max_length=50)]] = None
```
MCatherine1994, line 73: shall we add a comment there? once we finish the one-time load, the first, last name and email shouldn't be optional anymore?

ianliuwk1019: Hi Catherine, this TargetUser schema object can be more flexible since it is only internal that does not return to frontend; and it needs to deal with information gathering for few fields from the search result at separate stage after it is initially initialized. 
Before saving, the ORM model can check for entity fields' correctness, so should be good.
But, right now our db has "first_name", "last_name", "email" as **not** nullable, maybe they also can be nullable as they are from external and cannot always be sure they exist in external system... that's later topic when there is issue...

server/admin_management/api/app/services/application_admin_service.py
ianliuwk1019: Hi Catherine, both could work. 
I feel the "find_or_create" already has complex logic not only conditionally trying to find by "user_name" or "user_guild" and dealing with user_name update, so I feel doing a separate function for a single purpose for just update all other fields (except user_name) is a bit clear to me.

MCatherine1994: right Ian, cause if found user, we should do an update, thanks so much for explanation!!

MCatherine1994 (approved):

Issue #1481, opened 2024-07-15
Allow search test user accounts in FAM production for APPs in dev and test env


MCatherine1994:
**Describe the task**
Currently our FAM production only allows searching production user (cause it's only connected with production  IDIM web service). However, for the application integrated with FAM, their DEV and TEST environments are also integrated with our FAM production, we need to allow them to verify and add test business bceid to their APP in test environment. For example, we should allow Sabina to verify a test business bceid user and grant access for SPAR TEST using FAM PROD. 

**Acceptance Criteria**
- [ ] Pass the selected application's environment to backend api call when verifying user from frontend (if selected application is FAM, then that can be treated as prod application. FAM can only grant IDIR as app admin, so shouldn't be impacted)
- [ ] IDIM proxy url will be depended on the application environment
- [ ] Add both IDIM proxy TEST and PROD urls to terraform as environment variables

**Additional context**
- Add any other context about the task here.
- Or here

PR #1489, opened 2024-07-16
fix: #1481 Fix application env parameter undefined error

MCatherine1994:
refs: #1481


frontend/src/store/ApplicationState.ts
```diff
@@ -30,20 +30,10 @@ export const selectedApplicationId = computed(() => {
 
 // --- Getter
 
-export const selectedApplicationShortDisplayText = computed(() => {
```
MCatherine1994, line 33: this is not in use

ianliuwk1019 (approved):
Interesting is has to be cast to AppEnv but I approve.

PR #1490, opened 2024-07-17
fix: #1481 add new configuration for idim api token to support test/prod environments for FAM PROD

ianliuwk1019:
ref #1481 
* Add "IDIM_PROXY_API_KEY_TEST" and "IDIM_PROXY_API_KEY_PROD" environment vars to support IDIM Proxy search on TEST and PROD.
* Add terraform configuration to setup the new two keys for fam_api.tf backend api.
* Add same switching logic at backend config.py file to obtain different API key based on app_env.
* TODO: need to setup Github secrets for DEV/TEST/PROD(2 new keys)


ianliuwk1019:
Close this PR as confirmed with @MCatherine1994 , the IDIM proxy key is the same for all environments, so no need to add this.

Issue #1487, opened 2024-07-16
Allow admin management backend verifying test bceid users in FAM prod


MCatherine1994:
**Describe the task**
Currently our FAM production only allows searching production user (cause it's only connected with production  IDIM web service). However, for the application integrated with FAM, their DEV and TEST environments are also integrated with our FAM production, we need to allow them to verify and add test business bceid to their APP in test environment. For example, we should allow Sabina to verify a test business bceid user and grant access for SPAR TEST using FAM PROD. 

**Acceptance Criteria**
- [ ] Update admin management router guards to allow verifying users to use the IDIM proxy url based on app env 
- [ ] Follow the same fix as we did for app access management backend

**Additional context**
- follow up on #1481

PR #1493, opened 2024-07-18
fix: #1487 Fix IDIM user verification for admin management router guards

MCatherine1994:
refs: #1487

Pass application_id from frontend to reduce confusion. The app_environment is null for FAM.


server/backend/api/app/crud/crud_utils.py
```diff
@@ -100,3 +101,31 @@ def is_app_admin(
...
+def get_aws_target_env() -> AwsTargetEnv:
+    # target_env is assigned from gov's AWS platform, does not exist in local (None).
+    return os.environ.get("target_env")
+
+
+def is_on_aws_prod() -> bool:
+    return get_aws_target_env() == AwsTargetEnv.PROD
+
+
+def use_api_instance_by_app(application: models.FamApplication) -> ApiInstanceEnv:
```
MCatherine1994, line 115: I was thinking if put these methods in config.py or utils, maybe better to be in the utils?

server/backend/api/app/crud/crud_utils.py
```diff
@@ -100,3 +101,31 @@ def is_app_admin(
...
+    FAM PROD environment supports (DEV/TET/PROD) integrated applications.
+    Only PROD application at FAM PROD uses API instance in PROD.
+    Lower FAM environment uses only TEST instance.
+    Ref @FAM Wiki: https://github.com/bcgov/nr-forests-access-management/wiki/Environment-Management
+    """
+    api_instance_env = ApiInstanceEnv.TEST  # API TEST instance as default.
+    if (is_on_aws_prod() and (
+        # either PROD app or app is FAM
+        application.app_environment == AppEnv.APP_ENV_TYPE_PROD or
+        application.application_name == APPLICATION_FAM
```
MCatherine1994, line 126: so I pass application instead of the app_env, so we can check by application_name

server/admin_management/api/app/services/utils_service.py
```diff
@@ -0,0 +1,35 @@
...
+    FAM PROD environment supports (DEV/TET/PROD) integrated applications.
+    Only PROD application at FAM PROD uses API instance in PROD.
+    Lower FAM environment uses only TEST instance.
+    Ref @FAM Wiki: https://github.com/bcgov/nr-forests-access-management/wiki/Environment-Management
+    """
+    api_instance_env = ApiInstanceEnv.TEST  # API TEST instance as default.
+    if is_on_aws_prod() and (
+        # either PROD app or no app_env (which is FAM, FAM has no app_env)
+        app_env == AppEnv.APP_ENV_TYPE_PROD
+        or not app_env
```
MCatherine1994, line 30: In the case of granting application admin, there is no role information to get the application information. So if app_env is None, that's FAM

server/admin_management/api/app/routers/router_guards.py
```diff
@@ -168,14 +179,55 @@ async def get_target_user_from_id(
...
     target_user: TargetUser = Depends(get_target_user_from_id),
+    role: FamRole = Depends(get_request_role_from_id),
 ) -> TargetUser:
     """
     Validate the target user by calling IDIM web service, and update business Guid for the found BCeID user
     """
-    target_user_validator = TargetUserValidator(requester, target_user)
+    # ignore the role validation here, cause we have other router guards to do that when grant/delete delegated admin
+    # in the case that role is none, that's for granting/deleting application admin with FAM, which has no app_env
+    app_env = role.application.app_environment if role and role.application else None
```
MCatherine1994, line 225: we will either get application from role while granting/deleting delegated admin, or set app_env to be None in the case of granting application admin

server/admin_management/local-dev.env
```diff
@@ -8,5 +8,6 @@ COGNITO_USER_POOL_ID=ca-central-1_p8X8GdjKW
 COGNITO_CLIENT_ID=6jfveou69mgford233or30hmta
 COGNITO_USER_POOL_DOMAIN=dev-fam-user-pool-domain
 FC_API_TOKEN=thisisasecret
-IDIM_PROXY_BASE_URL=https://nr-fam-idim-lookup-proxy-test-backend.apps.silver.devops.gov.bc.ca
+IDIM_PROXY_BASE_URL_TEST=https://nr-fam-idim-lookup-proxy-test-backend.apps.silver.devops.gov.bc.ca
+IDIM_PROXY_BASE_URL_PROD=https://nr-fam-idim-lookup-proxy-prod-backend.apps.silver.devops.gov.bc.ca
```
ianliuwk1019, line 12: Hi Catherine, maybe this _PROD can be removed?

server/backend/api/app/crud/crud_utils.py
ianliuwk1019: Having it at utils is great!! I didn't think about that. Thanks.
Maybe `get_aws_target_env()` can be in config file.

Also I didn't think about passing 'application' instead of app_environment and then check for `application_name`, but I think it is better, I told you you have sharp eyes. Do you think it is reusable in all cases? Probably yes. I would change that on my end.

And, in `server/admin_management` you didn't do the same logic and didn't pass application. Maybe it can be consistent?

server/backend/api/app/routers/router_guards.py
```diff
@@ -438,3 +441,10 @@ def enforce_bceid_terms_conditions_guard(
             error_code=ERROR_CODE_TERMS_CONDITIONS_REQUIRED,
             error_msg="Requires to accept terms and conditions.",
         )
+
+
+def get_api_instance_env(
+    application_id: int, db: Session = Depends(database.get_db)
+) -> ApiInstanceEnv:
+    application = crud_application.get_application(db, application_id)
+    return crud_utils.use_api_instance_by_app(application)
```
ianliuwk1019, line 450: Hi Catherine, can this be in `router_utils.py`?

server/backend/api/config/config.py
```diff
@@ -162,13 +162,9 @@ def get_forest_client_api_baseurl():
     return forest_client_api_baseurl
 
 
-def get_idim_proxy_api_baseurl(app_env: AppEnv):
+def get_idim_proxy_api_baseurl(api_instance_env: ApiInstanceEnv):
     idim_proxy_api_baseurl = get_env_var("IDIM_PROXY_BASE_URL_TEST")
```
ianliuwk1019, line 166: Nothing wrong here.
Just wondering, previously for "forest_client" the baseurl is hardcode as default value and does not need to be set in local_dev as one of the environment variable.
Do you prefer to get it for local or also hardcode the value as default?
For higher environment this will be a get().

server/backend/testspg/router/test_router_idim_proxy.py
```diff
@@ -23,6 +24,7 @@
 
 endPoint_search_idir = f"{apiPrefix}/identity_search/idir"
 endPoint_search_bceid = f"{apiPrefix}/identity_search/bceid"
+endPoint_search_param_application_id = f"&application_id={FOM_DEV_APPLICATION_ID}"
```
ianliuwk1019, line 27: Didn't check each test case, but I assume `&application_id={FOM_DEV_APPLICATION_ID}"` because it all use FOM_DEV?

ianliuwk1019 (commented):
Hi Catherine, look very good to me, just minor questions.

server/backend/api/app/routers/router_guards.py
MCatherine1994: um... I was thinking of that before, but then feel it can be a router dependency, so the idim router apis can get the api instance directly
Screen Shot 2024-07-18 at 3 52 24 PM

server/backend/testspg/router/test_router_idim_proxy.py
MCatherine1994: I use that because the router tests doesn't care about the application, and we pass the application_id just to let the idim uses test url. In the real case, user will select an application first, and then verify the user, the application_id will be passed to the api call as a parameter.

server/admin_management/local-dev.env
MCatherine1994: I put it here just because I was testing the prod situation previously on local. I'll remove it.

server/backend/api/app/crud/crud_utils.py
MCatherine1994: Hi Ian, that's because it's impossible to get application information from role (no role) when we grant app admin. And I don't want to make another database call to grab application information by application_admin_id or application from request body. Cause we're pretty sure in the case of granting app admin, we'll use the prod url for idim apis. So if you check the similar method I had for admin management use_api_instance_by_app_env, it uses app_env, and in the case of app_env is None, that means it's FAM.

ianliuwk1019: No worry, I just noticed it is not consistent and didn't recall your case.
It's good, just I need to see how to reuse yours (should be no problem), I will just wait on yours since I am having strange error on my workspace.

server/admin_management/api/app/services/utils_service.py
```diff
@@ -0,0 +1,35 @@
...
+
+from api.app.constants import AppEnv, ApiInstanceEnv, AwsTargetEnv
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_aws_target_env() -> AwsTargetEnv:
+    # TARGET_ENV is assigned from gov's AWS platform, does not exist in local (None).
+    return os.environ.get("TARGET_ENV")
```
MCatherine1994, line 12: Hi Ian, I just realized this needs to be upper case "TARGET_ENV" instead of "target_env". Terraform has it in upper case.

server/backend/api/app/routers/router_guards.py
ianliuwk1019: Oh, sorry, you probably misunderstood.
I meant, putting "get_api_instance_env()" in "router_utils.py" can still be used in the router as dependency.

server/backend/api/app/crud/crud_utils.py
MCatherine1994: yeah, oh yeah, it should be fine. We verify forest client number when grant delegated admin, and we should have the role.application.app_environment information.

server/admin_management/api/app/services/utils_service.py
ianliuwk1019: Yes, I think I noticed the other day but forgot again to mention... thanks!

server/backend/api/config/config.py
MCatherine1994: Hi Ian, I think about this again, and I agree we could remove it, I also did a cleanup of the terraform, almost forgot, thanks so much!

ianliuwk1019 (approved):
Looks good to me, thank you so much Catherine!!

Issue #1367, opened 2024-05-08
Update GET methods to return first name, last name and email address


MCatherine1994:
**Describe the task**
Update get application admin, get delegated admin and get user role assignment methods to return user information with first name, last name and email address

**Acceptance Criteria**
- [x] verify the return results included first/last/email (based on schema we expect this to be true))
- [x] update tests
- [x] review and update audit log

**Additional context**
- Found a bug while testing locally, the bug is related to the audit log, add to the acceptance criteria to fix the audit log, because it could impact production
- Or here

PR #1497, opened 2024-07-22
feat: #1367 Update get method to return user information

MCatherine1994:
refs: #1367


ianliuwk1019 (approved):
looks good, thank you Catherine!

PR #1498, opened 2024-07-22
fix: #1367 Fix and simplify audit log

MCatherine1994:
refs: #1367


server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -76,35 +76,32 @@ def create_access_control_privilege_many(
...
             access_control_privilege_request.role_id
         )
         audit_event_log.application = audit_event_log.role.application
 
-        return access_control_privilege_service.create_access_control_privilege_many(
+        new_delegated_admin = access_control_privilege_service.create_access_control_privilege_many(
             access_control_privilege_request, requester.cognito_user_id, target_user
         )
+        # get target user from database, so for existing user, we'll have get the cognito user id
+        audit_event_log.target_user = user_service.get_user_by_domain_and_guid(
```
MCatherine1994, line 89: The only difference is that, the target_user returns from router_guard won't have the cognito_user_id. So in the case the target_user is an existing user, by find user from database, we'll get the cognito_user_id.

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -67,20 +67,25 @@ def create_user_role_assignment(
...
-        return crud_user_role.create_user_role(
+        new_user_role_assignment =  crud_user_role.create_user_role(
             db,
             role_assignment_request,
             target_user,
-            requesting_user.cognito_user_id
+            requester.cognito_user_id
+        )
+        # get target user from database, so for existing user, we'll have get the cognito user id
+        audit_event_log.target_user = crud_user.get_user_by_domain_and_guid(
```
MCatherine1994, line 83: instead of get user by domain and user name, we get user by domain and user guid now.

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -76,35 +76,32 @@ def create_access_control_privilege_many(
     )
 
     try:
-        audit_event_log.requesting_user = user_service.get_user_by_cognito_user_id(
-            requester.cognito_user_id
-        )
+        audit_event_log.requesting_user = requester
```
ianliuwk1019, line 79: Hi Catherine, I am wondering if this direct assign work (maybe work at runtime) as it seems they are different types.

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -76,35 +76,32 @@ def create_access_control_privilege_many(
...
         audit_event_log.role = role_service.get_role_by_id(
             access_control_privilege_request.role_id
         )
         audit_event_log.application = audit_event_log.role.application
 
-        return access_control_privilege_service.create_access_control_privilege_many(
+        new_delegated_admin = access_control_privilege_service.create_access_control_privilege_many(
```
ianliuwk1019, line 85: I feel maybe just use "response" as variable... this service method returns a list of response objects.

server/backend/api/app/routers/router_user_role_assignment.py
ianliuwk1019: Just wondering, the "target_user" and the crud_user.get_user_by_domain_and_guid() are having different types, does it work for variable assignment?

server/admin_management/api/app/routers/router_application_admin.py
```diff
@@ -75,34 +77,32 @@ def create_application_admin(
...
         audit_event_log.application = application_service.get_application(
             application_admin_request.application_id
         )
 
-        return application_admin_service.create_application_admin(
+        new_application_admin = application_admin_service.create_application_admin(
             application_admin_request, target_user, requester.cognito_user_id
         )
 
+        # get target user from database, so for existing user, we'll have get the cognito user id
```
ianliuwk1019, line 89: Sorry Catherine, do you mean "we can get"?

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: yeah, for "requester" I updated the type to be the Requester type from schema in audit config

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: yeah thanks Ian I'll change it

server/admin_management/api/app/routers/router_application_admin.py
MCatherine1994: yes, sorry hh

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: but I'll double check the type of "target_user"

ianliuwk1019 (dismissed):

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -76,34 +76,36 @@ def create_access_control_privilege_many(
...
-            audit_event_log.target_user = models.FamUser(
-                user_type_code=access_control_privilege_request.user_type_code,
-                user_name=access_control_privilege_request.user_name,
-                user_guid="unknown",
-                cognito_user_id="unknown",
+            audit_event_log.target_user = FamUser(
+                user_type_code=target_user.user_type_code,
+                user_name=target_user.user_name,
+                user_guid=target_user.user_guid,
+                cognito_user_id=target_user.cognito_user_id,
```
ianliuwk1019, line 108: I see, so you tried to use `**target_user` but the schema object it not something that can be destructed.
I wonder if `**target_user.model_dump_json()` would work... but it does not matter for now.

ianliuwk1019 (approved):

server/admin_management/api/app/routers/router_access_control_privilege.py
MCatherine1994: yeah I actually tried, but the target_user schema is based on the requester schema, and it has a field "access_roles". I got an error while using the model_dump. "TypeError: 'access_roles' is an invalid keyword argument for FamUser".

Issue #1115, opened 2024-01-09
Display user’s first name, last name and email in the table


MCatherine1994:
**Describe the task**
As mentioned in #911, display user info

**Acceptance Criteria**
- [x] Add new columns first name, last name and email in app admin, delegate admin and user access tables
- [x] Make sure new columns are sortable and searchable

**Additional context**
- Add any other context about the task here.
- IDIR and BCeID in scope
- depends on #1366, #1367

OlgaLiber2:
@MCatherine1994  I think we can close this ticket now.

PR #1499, opened 2024-07-23
feat: #1115 Display user information in table

MCatherine1994:
refs: #1115

### DIFF
--- a/frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
+++ b/frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
@@ -47,19 +47,6 @@ const newAppAdminIds = computed(() => {
 
 const adminFilters = ref({
     global: { value: '', matchMode: FilterMatchMode.CONTAINS },
-    'user.user_name': { value: null, matchMode: FilterMatchMode.STARTS_WITH },
-    'user.user_type.description': {
-        value: null,
-        matchMode: FilterMatchMode.CONTAINS,
-    },
-    'role.role_name': {
-        value: null,
-        matchMode: FilterMatchMode.CONTAINS,
-    },
-    'application.app_environment': {
-        value: null,
-        matchMode: FilterMatchMode.CONTAINS,
-    },
 });
 
 const adminSearchChange = (newvalue: string) => {
@@ -120,6 +107,9 @@ const highlightNewAppAdminAccesRow = (rowData: any) => {
                     'user.user_name',
                     'application.application_name',
                     'user.user_type.description',
+                    'user.first_name',
+                    'user.last_name',
+                    'user.email',
                     'role.role_name',
                     'application.app_environment',
                 ]"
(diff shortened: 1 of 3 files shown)


frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -47,19 +47,6 @@ const newAppAdminIds = computed(() => {
 
 const adminFilters = ref({
     global: { value: '', matchMode: FilterMatchMode.CONTAINS },
-    'user.user_name': { value: null, matchMode: FilterMatchMode.STARTS_WITH },
```
MCatherine1994, line 50: Realized we don't need these filters, those are for individual column filter. But in our case, we just need to global filter.

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
```diff
@@ -152,6 +142,18 @@ const highlightNewAppAdminAccesRow = (rowData: any) => {
                     header="Domain"
                     sortable
                 ></Column>
+                <Column field="user.first_name" header="Full Name" sortable>
```
MCatherine1994, line 145: Seems like we can only pass one field data, so the sorting will be based on the first name, which I feel is fine. Previously the design will show both first_name and last_name column, but the new design we saw will just be full name.

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
```diff
@@ -172,6 +158,18 @@ const highlightNewDelegatedAdminAccessRow = (rowData: any) => {
                     header="Domain"
                     sortable
                 ></Column>
+                <Column field="user.first_name" header="Full Name" sortable>
+                    <template #body="{ data }">
+                        {{
+                            data.user.first_name && data.user.last_name
```
MCatherine1994, line 164: since we haven't done the one time load, the first and last name could be none, in this case will display nothing.

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
ianliuwk1019: I am curious (I may not be familiar with detail of this dataTable) that what this "**field**" actually do.
Can it be pass with full name as `:field="getUserFullName"`, or simply it is useless like this:
image

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
ianliuwk1019: That's great, thanks Catherine!

frontend/src/components/managePermissions/table/DelegatedAdminTable.vue
ianliuwk1019: Sure, make sense.

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
MCatherine1994: I think it links to the data we pass to the user table component, so it will be the fields inside the "userRoleAssingment" data
Screen Shot 2024-07-23 at 2 01 36 PM

ianliuwk1019: I did try both for what I was curious at, before I asked, and the datatable sorting and filtering seems still working fine.  So I was wondering maybe it is not necessary in our datatable.
But you can leave it there if you know it does affect something.

ianliuwk1019 (approved):

frontend/src/components/managePermissions/table/ApplicationAdminTable.vue
MCatherine1994: oh really? I know the filtering should work. This field might impact the sorting. We can review it again when switch to the new design.

PR #1501, opened 2024-07-24
fix: #1487 Fix typo of AWS target environment variable

MCatherine1994:
refs: #1487


ianliuwk1019 (approved):
Thanks for the quick fix, Catherine!!

Issue #1386, opened 2024-05-14
One time load to add first name, last name and email for existing users


MCatherine1994:
**Describe the task**
Write a script for a one time load to add the first name, last name and email for existing FAM users.

**Acceptance Criteria**
- [x] add the first name, last name, email and user_guid for existing FAM users
- [x] Create a new endpoint in the existing backend (protected by a Bearer token) to update first name, last name and email

**Additional context**
- Get a list of not found users
- IDIR and BCeID in scope
- load information for the user not having first and last name and email user guid
- api logic can refer to the spike ticket #1387

MCatherine1994:
Note that in our production database, there could have some test business bceid users that were created before we enabled the user validation for bceid. When we run the one time load, we want to track how many of the users in production database that cannot be found by IDIM web service. Later on we might want to do a cleanup for those old test users.

MCatherine1994:
One-time load response from FAM DEV:

[fam-dev-one-time-load-user-info-response.json](https://github.com/user-attachments/files/16419402/fam-dev-one-time-load-user-info-response.json)

MCatherine1994:
Found that API Gateway has a maximum 29s timeout limit. 

![Image](https://github.com/user-attachments/assets/3d178988-859d-437d-98b9-4bca50597c41)

Lambda also has a time out config, currently is set to be 15s, this can be changed. 

![Image](https://github.com/user-attachments/assets/7c7c39d0-d1b8-40e5-99c0-fe8d01433d51)


In the case that API Gateway has a maximum 29s limit that cannot changed, we could load around 100 users at a time.

PR #1504, opened 2024-07-25
feat: #1386 Add an api to load and update user information to match records in IDIM

MCatherine1994:
refs: #1386


server/backend/api/app/main.py
```diff
@@ -129,6 +130,13 @@ def main():
     dependencies=[Depends(router_guards.authorize)],
     tags=["FAM User Terms and Conditions"],
 )
+app.include_router(
+    router_user.router,
+    prefix=apiPrefix + "/user",
+    dependencies=[Depends(router_guards.verify_api_key_for_update_user_info)],
```
MCatherine1994, line 136: Protect the new api with an api key, because if we want to call this api by setting up a cron job in API Gateway in the future, it's hard to do a user authentication. Pass an api key will be easier.

server/backend/api/app/routers/router_user.py
```diff
@@ -0,0 +1,32 @@
...
+
+LOGGER = logging.getLogger(__name__)
+router = APIRouter()
+
+
+@router.post("", status_code=HTTPStatus.OK, response_model=FamUserUpdateResponse)
+def update_user_information_from_idim_source(
+    page: int = 1,
+    per_page: int = 100,
+    use_pagination: bool = False,
```
MCatherine1994, line 18: I added the pagination option, but it's optional, we can choose to use it or not. And I'm not sure how to setup cronjob with pagination, but we can discuss that the update ticket.

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,131 @@ def fetch_initial_requester_info(
...
+def update_user_info_from_idim_source(
+    db: Session, use_pagination: bool, page: int, per_page: int
+) -> schemas.FamUserUpdateResponse:
+    """
+    Go through each user record in the database, update the user information to match the record in IDIM web service
+    """
+    # get a requester from the database
+    requester = (
+        db.query(models.FamUser)
+        .filter(models.FamUser.user_name == "CMENG")
```
MCatherine1994, line 267: we need to setup a requester (valid IDIR user) in order to call IDIM web service, I just put my name here.

server/backend/api/app/schemas.py
```diff
@@ -73,6 +73,14 @@ class FamUserInfo(BaseModel):
     )
 
 
+class FamUserUpdateResponse(BaseModel):
+    total_users_count: int
+    current_page: int
+    users_count_on_page: int
+    success_user_id_list: List[int]
+    failed_user_id_list: List[int]
```
MCatherine1994, line 81: return the success and failed user id list, for example as below:

Screen Shot 2024-07-25 at 3 06 13 PM

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+                            }
+                        )
+                    )
+                    properties_to_update = {
+                        models.FamUser.user_guid: search_result.get("guid"),
+                    }
+            else:
+                # ignore bc service card users
+                ignored_user_list.append(user.user_id)
+                LOGGER.debug(f"Updating information for user {user.user_name} is ignored because we only focus on IDIR and Business BCeID")
```
MCatherine1994, line 353: ignore the users with bc service card types

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+    )
+    # setup IDIM web service
+    api_instance_env = (
+        ApiInstanceEnv.PROD if crud_utils.is_on_aws_prod() else ApiInstanceEnv.TEST
+    )
+    idim_proxy_service = IdimProxyService(requester, api_instance_env)
+
+    # grab fam users from user table
+    fam_users = get_users(db)
+    total_users_count = len(fam_users)
```
ianliuwk1019, line 280: I wonder if this can be named "total_db_users_count"

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+                failed_user_list.append(user.user_id)
+
+        except Exception as e:
+            LOGGER.debug(f"Failed to update user info: {e}")
+            failed_user_list.append(user.user_id)
+
+    return schemas.FamUserUpdateResponse(
+        **{
+            "total_users_count": total_users_count,
+            "current_page": page,
```
ianliuwk1019, line 382: if `not use_pagination`, do you still intend to give it a value for "current_page" as 1?

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+                # IDIM web service doesn't support search IDIR by user_guid, so we search by userID
+                search_result = idim_proxy_service.search_idir(
+                    schemas.IdimProxySearchParam(**{"userId": user.user_name})
+                )
+
+                if not user.user_guid:
+                    # if user has no user_guid in our database, add it
+                    properties_to_update = {
+                        models.FamUser.user_guid: search_result.get("guid"),
+                    }
```
ianliuwk1019, line 316: Hi Catherine, wonder maybe we don't need to add this "guid" to properties_to_update just here.

When you later have the `if search_result.get("found")` when search_result is **found**, just add this "guild" to update there, same as "userId" and all other fields?

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+                        schemas.IdimProxyBceidSearchParam(
+                            **{
+                                "searchUserBy": IdimSearchUserParamType.USER_GUID,
+                                "searchValue": user.user_guid,
+                            }
+                        )
+                    )
+                    properties_to_update = {
+                        models.FamUser.user_name: search_result.get("userId"),
+                    }
```
ianliuwk1019, line 335: Same as previous comment:

When you later have the if search_result.get("found") when search_result is **found**, just add this "userId" to update there, same as "guid" and all other fields?

server/backend/api/app/crud/crud_user.py
```diff
@@ -246,9 +247,142 @@ def fetch_initial_requester_info(
...
+                    models.FamUser.email: search_result.get("email"),
+                    models.FamUser.business_guid: search_result.get("businessGuid"),
+                }
+
+                update(db, user.user_id, properties_to_update, requester.cognito_user_id)
+                LOGGER.debug(f"Updating information for user {user.user_name} is done")
+                success_user_list.append(user.user_id)
+            else:
+                LOGGER.debug(
+                    f"Invalid request, cannot find user {user.user_name} {user.user_guid} with user type {user.user_type_code}"
```
ianliuwk1019, line 371: Probably can remove "Invalid request," as there is no request in this context.

server/backend/api/app/crud/crud_user.py
MCatherine1994: Hi Ian, that's because we shouldn't update user guid if the user already has that value in our database. 

So basically two cases:

- We have user's user_guid in the database, find user by user_guid, if the information mismatch, we update the username and etc. We should never update user_guid, we rely on that
- If we don't have user's user_guid, then we search by user_name, and add the user_guid into the database

For idir user, since we can't search user by user_guid (IDIM doesn't support), we'll search by user_name. The we only add user_guid, not making changes. Cause we'll always rely on user_guid, not user_name

server/backend/api/app/crud/crud_user.py
MCatherine1994: I'm actually fine with it... do you feel confuse?

server/backend/api/app/crud/crud_user.py
MCatherine1994: thanks Ian, I added a method in the config.py

server/backend/api/app/routers/router_user.py
```diff
@@ -0,0 +1,32 @@
...
+from api.app.crud import crud_user
+from fastapi import APIRouter, Depends
+from sqlalchemy.orm import Session
+from api.app.schemas import FamUserUpdateResponse
+
+LOGGER = logging.getLogger(__name__)
+router = APIRouter()
+
+
+@router.post("", status_code=HTTPStatus.OK, response_model=FamUserUpdateResponse)
```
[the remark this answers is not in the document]
MCatherine1994: How about this?
Screen Shot 2024-07-26 at 2 19 23 PM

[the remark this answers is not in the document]
ianliuwk1019: Thanks, I can't come up with a good path name. Seems acceptable.

server/backend/api/app/crud/crud_user.py
ianliuwk1019: I only wonder your intention since at one hand use_pagination is False, but on the other hand it returns a current_page. I think it is fine. No big deal.

server/backend/api/app/crud/crud_user.py
ianliuwk1019: Maybe I am confused,
Isn't that if it finds the user by the user_guid, then it does not matter to add this field to the update list or not since it is the same; but if it finds the user by the user_id, then updating to the user_guid is correct since this is the only user that has the user_guid (when empty in our db)?
Unless there could be user(s) that they all have the same user_guid (with the same type)? Then in that case, db probably throw error and your code does log the error for what's wrong with the user update?

ianliuwk1019: Resolved confusion with Catherine.

ianliuwk1019 (approved):
Thank you very much Catherine!

Issue #1511, opened 2024-07-29
Refactor POST user role assignment api to support multiple forest client numbers


MCatherine1994:
**Describe the task**
Currently when we grant user access with multiple forest clients numbers, we call the Backend API multiple times. But we don't want to send notification email multiple times, update the Backend API to support multiple forest client number in one api.

**Acceptance Criteria**
- [x] Backend update post user role assignment api to support grant access with multiple forest clients numbers
- [ ] Update/add backend tests if need
- [x] Frontend update form data to support multiple forest client numbers
- [x] When submit the form, make one api call with the updated api

**Additional context**
- Add any other context about the task here.
- Or here

OlgaLiber2:
option to send email is at the bottom of the screen 
@MCatherine1994 @ianliuwk1019 

![Image](https://github.com/user-attachments/assets/a9d02738-f8d9-4b63-aad1-165a27696645)

PR #1515, opened 2024-08-01
feat: #1511 Support grant user access with multiple forest client numbers in one api call

MCatherine1994:
refs: #1511

- Backend api accepts multiple forest client numbers for granting access user role assignments
- Frontend call the granting access user role assignments only once


server/backend/api/app/routers/router_guards.py
```diff
@@ -235,35 +235,43 @@ async def authorize_by_privilege(
...
             if (
-                forest_client_number
+                forest_client_numbers
                 and role.role_type_code == RoleType.ROLE_TYPE_ABSTRACT
             ):
                 parent_role = role
-                forest_client_role_name = (
-                    crud_user_role.construct_forest_client_role_name(
-                        parent_role.role_name, forest_client_number
+                for forest_client_number in forest_client_numbers:
```
MCatherine1994, line 248: Hi Ian, when we accept multiple forest client numbers from frontend, for delegated admin, we need to check the privilege for each forest client number. So before making too many changes, for now, as long as the delegated admin miss the privilege for one of the forest client numbers, will throw permission error. We could discuss that in our scrum next week. I'm thinking in the future, once we only allow delegated admin to select forest client numbers from allowed list, this won't be a big problem.

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -72,78 +74,118 @@ def create_user_role(
...
+                    error_code=famConstants.ERROR_CODE_INVALID_REQUEST_PARAMETER,
+                    error_msg=error_msg,
+                )
+
+            # Check if child role exists or add a new child role
+            child_role = find_or_create_forest_client_child_role(
+                db, forest_client_number, fam_role, requester
             )
+            # Role for associating with user
+            associate_role = child_role if require_child_role else fam_role
```
ianliuwk1019, line 134: This condition seems not necessary anymore, maybe can use `child_role` directly for `create_user_role_assignment`.

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -72,78 +74,118 @@ def create_user_role(
...
+    xref_dict = {"application_id": role.application_id, "user": user, "role": role}
 
     if fam_user_role_xref:
-        LOGGER.debug(
-            "FamUserRoleXref already exists with id: " +
-            f"{fam_user_role_xref.user_role_xref_id}."
+        xref_dict = {**fam_user_role_xref.__dict__, **xref_dict}
+        error_msg = (
+            f"Role {fam_user_role_xref.role.role_name} already assigned to user {fam_user_role_xref.user.user_name}. "
+            + f"FamUserRoleXref already exists with id: {fam_user_role_xref.user_role_xref_id}"
```
ianliuwk1019, line 165: I feel maybe we can remove the second line from error_msg, probably useless to return to frontend since first line already tells all.

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -258,57 +297,10 @@ def find_or_create_forest_client_child_role(
...
-        .filter(
-            models.FamUserRoleXref.user_role_xref_id == user_role_xref_id
-        )
+        .filter(models.FamUserRoleXref.user_role_xref_id == user_role_xref_id)
         .one_or_none()
     )
     return user_role
-
-
-class UserRoleValidator:
```
ianliuwk1019, line 274: This was very early code. The initial intention was to have a central validator to check for all business validations for user_role_assignment create request, not just for "forest_client_number". It was not in the validator folder initially and for some reason the role_assignment creation didn't use it for other validations anymore... 
I think it is ok that you create a new validator in the validator folder only for forest_client_number.

server/backend/testspg/crud/validator/test_forest_client_validator.py
```diff
@@ -76,57 +87,23 @@ def test_forest_client_number_exists(
...
-    mock_forest_client,
-):
-    request_dict = {**ACCESS_GRANT_FOM_DEV_AR_00000001_IDIR, "forest_client_number": client_id_to_test}
-    mock_forest_client.find_by_client_number.return_value = (
-        __to_mock_forest_client_return(
-            request_dict["forest_client_number"],
-            expcted_result["api_client_status_code"],
-        )
+    # find forest client number
+    forest_client_validator_return = (
```
ianliuwk1019, line 100: Since this is not a return from validator, maybe the variable can be "forest_client_search_return"?

ianliuwk1019 (commented):
Thank you so much Catherine, just have some small comments.

server/backend/api/app/schemas.py
```diff
@@ -266,6 +274,14 @@ class GCNotifyGrantAccessEmailParam(BaseModel):
     send_to_email: EmailStr
 
 
+# ------------------------------------- Forest Client API Integraion ---------------------------------------- #
+class ForestClientIntegrationFindResponse(BaseModel):
```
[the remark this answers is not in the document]
MCatherine1994: Hi Ian, this one is copied from admin management, and I remember we decided to include the "integration" in the name, so it's different than the FAM forest client number methods. If you feel it's better to remove the integration, I'm fine with it, I'll change both folders then.

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -72,78 +74,118 @@ def create_user_role(
...
-                "Invalid role assignment request. " +
-                f"Forest Client Number {request.forest_client_number} does not exist.")
-            raise_http_exception(
-                error_code=famConstants.ERROR_CODE_INVALID_REQUEST_PARAMETER,
-                error_msg=error_msg
+        forest_client_integration_service = ForestClientService(api_instance_env)
+
+        for forest_client_number in request.forest_client_numbers:
+            # validate the forest client number
+            forest_client_validator_return = (
```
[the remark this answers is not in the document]
MCatherine1994: yeah I agree with you. This one is also copied from admin management, I would prefer to update both if we need to. However, since forest client api now support search multiple forest client numbers in one api call, do you mind if we refactor this when we switch to use that new api?

server/backend/api/app/crud/crud_user_role.py
MCatherine1994: I think frontend is using different error messages, not directly using the messages from backend. But I agree, I should start to write error messages that is good enough that can be used by frontend directly for the future.

server/backend/api/app/crud/crud_user_role.py
MCatherine1994: yeah but if we check the code, it's actually just checking the forest client numbers... I feel it's okay to remove it

server/backend/api/app/schemas.py
```diff
@@ -226,6 +226,14 @@ class FamApplicationUserRoleAssignmentGet(FamUserRoleAssignmentGet):
     )
 
 
+class FamUserRoleAssignmentCreateResponse(BaseModel):
+    status_code: int
+    detail: FamApplicationUserRoleAssignmentGet
```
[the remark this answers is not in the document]
MCatherine1994: yeah, FamUserRoleAssignmentGet is actually not used by other places anymore, I'll combine the two

server/backend/api/app/schemas.py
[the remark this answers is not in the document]
ianliuwk1019: Sorry Catherine, I didn't recall previous conversation, but if you now make them consistent I think it is ok. Thanks!!

ianliuwk1019 (approved):
Thank you so much Catherine for the great refactoring!!

PR #1517, opened 2024-08-06
fix: Upgrade test container version to fix pipeline issue

MCatherine1994:
Got a pipeline error when working on the pr #1515, all backend tests failed because of the test container cannot start the container from the docker compose file anymore. https://github.com/bcgov/nr-forests-access-management/actions/runs/10271029988/job/28426203492
Screen Shot 2024-08-06 at 3 15 05 PM
Screen Shot 2024-08-06 at 3 14 38 PM
The cause of the issue is not very obvious. Everything works last week, and started to get this error from this morning. But could related to the way how test container starts the docker, the `docker-compose` command is no longer supported in GitHub Actions. I didn't find any documentation to explain the issue. But the issue could be solved by updating the `testcontainers` version to the latest v4.7.2, and that requires python > 3.9. 
Screen Shot 2024-08-06 at 3 07 57 PM

So updated `testcontainers` version to v4.7.2 in backend, admin management and auth lambda folder, updated python to be v3.12 in pipeline workflow yamls.

### DIFF
--- a/.github/workflows/reusable_terraform_server.yml
+++ b/.github/workflows/reusable_terraform_server.yml
@@ -64,10 +64,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up Python 3.8
+      - name: Set up Python 3.12
         uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.12"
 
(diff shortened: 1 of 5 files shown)


ianliuwk1019 (approved):
Thanks Catherine!

PR #1518, opened 2024-08-07
fix: Update python version in terraform to keep consistent

MCatherine1994:
Upgrade python to v3.12 in terraform to be the same version in the GitHub pipeline workflow yamls. 

We got a deployment error when merged pr #1517 https://github.com/bcgov/nr-forests-access-management/actions/runs/10275482241, the Api Gateway is not reachable. It's caused by a python version mismatch issue. After manually set the lambda python version to 3.12, the Api Gateway works again. 
Screen Shot 2024-08-06 at 5 08 38 PM

### DIFF
--- a/.github/workflows/reusable_terraform_server.yml
+++ b/.github/workflows/reusable_terraform_server.yml
@@ -64,10 +64,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Set up Python 3.8
+      - name: Set up Python 3.12
         uses: actions/setup-python@v4
         with:
-          python-version: "3.8"
+          python-version: "3.12"
 
       - name: Auth function zip
         run: |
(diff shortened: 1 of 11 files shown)


MCatherine1994:
Hi Ian, I wonder if you know what this "version: 3.8" means here? It's in all three docker compose files @ianliuwk1019 
image

ianliuwk1019:
> Hi Ian, I wonder if you know what this "version: 3.8" means here? It's in all three docker compose files @ianliuwk1019 image

Hm... I don't think I know the reason. They were probably inherit from (copy-paste) the first docker-compose file, but I don't know why it was 3.8. I think we could remove this "version" to avoid confusion, the docker docs says it is not being used.

ianliuwk1019 (dismissed):
Thanks Catherine!
I see the sonar template also has this "3.8" but not sure if it is actually used...
image

ianliuwk1019 (approved):

Issue #799, opened 2023-08-16
Enable notification email after grant user access


MCatherine1994:
**Describe the task**
Send email notification after granting a user access in FAM. We already have our backend integrated with GC Notify. 

**Acceptance Criteria**
- [x] Update IDIM proxy return schema to include user email
- [x] Update target user schema to include email
- [x] Review the GC Notify api, update parameters based on the template from ticket #1073  if needs 
- [x] At the end of the create user role assignment method, call the gc notify email api to send the email
- [x] Only call the email api when frontend check box is checked
- [x] Verify the parameter pass to the send email method, make sure the email address is a validate one

Additional information
- In the case that email sending failed, what we need to do?  Can we trust the gc notify server and our server enough?
- Email for granting delegated admin is in ticket #1469

MCatherine1994:
The first two acceptance criteria are done

MCatherine1994:
Currently when we grant app access with multiple forest client numbers, we call the api multiple times. But I feel we don't want to send multiple email in this case, maybe I should create another ticket to allow the current POST method support multiple forest client numbers? So from frontend we just call the api once, and send one email with multiple forest client numbers? @ianliuwk1019 

Delegated admin is fine.

ianliuwk1019:
> Currently when we grant app access with multiple forest client numbers, we call the api multiple times. But I feel we don't want to send multiple email in this case, maybe I should create another ticket to allow the current POST method support multiple forest client numbers? So from frontend we just call the api once, and send one email with multiple forest client numbers? [@ianliuwk1019](https://github.com/ianliuwk1019)
> 
> Delegated admin is fine.

I feel ideally it should have been like that previously; but due to we were reusing existing endpoint for quick implementation. I think we should, we probably don't want to send email for each different forest client number. It will have a major refactor at frontend I feel.

MCatherine1994:
> > Currently when we grant app access with multiple forest client numbers, we call the api multiple times. But I feel we don't want to send multiple email in this case, maybe I should create another ticket to allow the current POST method support multiple forest client numbers? So from frontend we just call the api once, and send one email with multiple forest client numbers? [@ianliuwk1019](https://github.com/ianliuwk1019)
> > Delegated admin is fine.
> 
> I feel ideally it should have been like that previously; but due to we were reusing existing endpoint for quick implementation. I think we should, we probably don't want to send email for each different forest client number individually. It will have a major refactor at frontend I feel but should be a better frontend after change.

[...]

PR #1527, opened 2024-08-13
feat: #799 sending grant user email

ianliuwk1019:
ref: #799 
- minor adjustment on schema for get application role assignment to be easier to get application name.
- adjust GC Notify template to replace initial application_team_contact_email to contact_message in order
  to conditionally rendering contact message easily.
- add/adjust email params to GC Notify service.
- send email when granting user access depends on requires to send user notification checkbox from frontend.
- catch 400/500 error so event does not rollback when email sending fails.
- generated api-client for schema change.
- adjust frontend getting application_id due to backend schema change.
- adjust backend tests for getting application_id due to backend schema change.

image


server/backend/api/app/crud/crud_application.py
```diff
@@ -23,7 +24,7 @@ def get_application(db: Session, application_id: int):
 
 def get_application_role_assignments(
     db: Session, application_id: int, requester: schemas.Requester
-) -> List[models.FamUserRoleXref]:
+) -> List[schemas.FamApplicationUserRoleAssignmentGet]:
```
ianliuwk1019, line 27: Nothing change, only change return type from 'model' to 'schema' (still the same type return at router)

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -167,3 +171,36 @@ def delete_user_role_assignment(
                 user_role.role.client_number.forest_client_number
             )
         audit_event_log.log_event()
+
+
+def __send_user_access_granted_email(
```
ianliuwk1019, line 176: In "admin_management", this method probably can be in a service; here in server/backend, I placed it here in router (don't really like it to be in CRUD) as a private method.

server/backend/api/app/schemas.py
```diff
@@ -11,21 +10,10 @@
 
 
 # --------------------------------- FAM Application --------------------------------- #
-class FamApplicationCreate(BaseModel):
```
ianliuwk1019, line 14: This is not being used, removed.

server/backend/api/app/routers/router_user_role_assignment.py
ianliuwk1019: As discussed, moved send_user_access_granted_email from private method for router into "crud_user_role" file as public method. Let router dispatches to send email as it should.

frontend/src/components/managePermissions/ManagePermissions.vue
```diff
@@ -134,7 +134,7 @@ const deleteUserRoleAssignment = async (
     try {
         userRoleAssignments.value = await deleteAndRefreshUserRoleAssignments(
             assignment.user_role_xref_id,
-            assignment.role.application_id
+            assignment.role.application.application_id
```
MCatherine1994, line 137: Hi Ian, I'm still wondering if we need to include the "application", cause the role table already has the application_id information... Screen Shot 2024-08-13 at 1 04 39 PM

server/backend/api/app/crud/crud_user_role.py
```diff
@@ -298,3 +297,37 @@ def find_by_id(db: Session, user_role_xref_id: int) -> models.FamUserRoleXref:
...
+        - FAM currently is not concerned with checking status from GC Notify (callback) to verify
+            if email is really sent from GC Notify.
+    """
+    granted_roles = ", ".join(item.detail.role.role_name for item in roles_assigned)
+    email_service = GCNotifyEmailService()
+    email_params = schemas.GCNotifyGrantAccessEmailParam(**{
+        "first_name": target_user.first_name,
+        "last_name": target_user.last_name,
+        "application_name": roles_assigned[0].detail.role.application.application_description,
+        "role_list_string": granted_roles,
```
MCatherine1994, line 324: Hi Ian, just remind that, how about in the case the status of granting one role is failed, but the rest are successful? should the email only include the successful roles?

ianliuwk1019: Oh yes, good catch, I forgot this method return multiple responses. Thanks, I will adjust that to send email for successful roles granted.

frontend/src/components/managePermissions/ManagePermissions.vue
ianliuwk1019: Initially it had the application id; but the schema or the return to frontend does not necessary (better if not in some cases) need to match exact from the model. I only changed it to grab the application_name for convenience; we could include both but I think it is not necessary.

MCatherine1994: oh I see... or we could pass the application_name as a parameter? so it won't impact the existing schema, and will not include additional application information in the return to frontend?

image

ianliuwk1019: I would not take application_name from audit_event_log's property; either getting from the main event response or from db again (which is what schema change to take advantage of ORM model conversion to schema).

It's been a long time not refactoring this, but this audit_log can actually be a stand-alone call rather than in router; and router may not necessary know audit_log has such property.

MCatherine1994: sure, then I'll follow the same for admin management so it's consistent, thanks!

MCatherine1994 (approved):

Issue #1471, opened 2024-07-02
Handle error case when email sending failed


MCatherine1994:
**Describe the task**
Add error handling for failed email

**Acceptance Criteria**
- [x] When email fails to send, display notification in frontend

**Additional context**
- error could be from GC Notify bad email address
- relates to #799 and #1469
- ticket #1532 is created to follow up for frontend to be able to display multiple notification messages for a type

MCatherine1994:
![Image](https://github.com/user-attachments/assets/ffe9998b-442a-4704-b54c-fbe84068e83c)



![Image](https://github.com/user-attachments/assets/c45c8bd7-1558-4f76-bc4e-78c8d4301b57)

MCatherine1994:
We are using Trail version and team key, so it can only send email to our team members or the safelist people. So we have to request go live and setup the limits.

ianliuwk1019:
Hi @OlgaLiber2 would you mind add the frontend general email sending failure error message text here from the mockup or the link to it? Thank you very much!

ianliuwk1019:
Also will change email templates to reflect:
* not show word "**for the following organizations**" when granting concrete role such as fom_reviewer.
@OlgaLiber2

OlgaLiber2:
@marina-carvalho can you help Ian with the error mockup please?

marina-carvalho:
Sure, @ianliuwk1019 Here's the link to the mockup with the error message: https://www.figma.com/design/WhDujL6e0lkgpno740BZBx/Manage-and-add-permissions%3A-Application-admin?node-id=3061-126007&t=YXh6qnJSubTHpIP6-4

ianliuwk1019:
> Sure, [@ianliuwk1019](https://github.com/ianliuwk1019) Here's the link to the mockup with the error message: https://www.figma.com/design/WhDujL6e0lkgpno740BZBx/Manage-and-add-permissions%3A-Application-admin?node-id=3061-126007&t=YXh6qnJSubTHpIP6-4

Hi @marina-carvalho Thank you very much! But I seem to have no permission to the link. It asks me to send you request and I did.

marina-carvalho:
@ianliuwk1019 can you try again, please?

ianliuwk1019:
Hi @marina-carvalho I can view them now, thanks for granting me access!!

ianliuwk1019:
Although the mockup text: "failed to send email to [[email@address.com]]:
[...]

PR #1531, opened 2024-08-15
feat: #1471 Handle email sending error

ianliuwk1019:
ref: #1471 
- email sending status enum defined to be used for email sending to external service.
- granting user and granting delegated admin endpoints return response changed to incorporate email sending status.
- adjust router and service due to response schema change.
- add 'with_client_number' email param to schema and its conditonal expression for email service
- set email sending status depending on sending result to the response.
- generate api client.
- adjust frontend to show email sending failure message when email sending failed.


server/admin_management/api/app/schemas.py
```diff
@@ -339,4 +344,5 @@ class GCNotifyGrantDelegatedAdminEmailParam(BaseModel):
     last_name: Annotated[str, StringConstraints(max_length=20)]
     role_list_string: Annotated[str, StringConstraints(max_length=200)]
     application_team_contact_email: Optional[EmailStr] = None
+    with_client_number: Literal['yes', 'no']
```
ianliuwk1019, line 347: This is new GC Notify's variable for conditional display (it needs value as 'yes' or 'no').

frontend/src/components/grantaccess/form/UserNameInput.vue
```diff
@@ -72,7 +72,7 @@ const verifyUserId = async () => {
         }
     } finally {
         if (verifiedUserIdentity.value?.found) {
-            emit('setVerifyResult', true, verifiedUserIdentity.value.guid);
+            emit('setVerifyResult', true, verifiedUserIdentity.value.guid, verifiedUserIdentity.value.email);
```
ianliuwk1019, line 75: emit searched user 'email'. This is then used for displaying "failed to send email to" error message based on mockup.

server/admin_management/api/app/routers/router_access_control_privilege.py
```diff
@@ -32,7 +27,7 @@
 
 @router.post(
     "",
-    response_model=List[schemas.FamAccessControlPrivilegeCreateResponse],
+    response_model=schemas.FamAccessControlPrivilegeResponse,
```
MCatherine1994, line 30: Hi Ian, I think you already imported the FamAccessControlPrivilegeResponse from schemas, so maybe no need to call it schemas.FamAccessControlPrivilegeResponse

server/admin_management/api/app/schemas.py
MCatherine1994: thanks Ian, later on once we support grant multiple roles might need to look at this again, good for now

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
```diff
@@ -136,6 +139,12 @@ const confirmSubmit = async () => {
                 errorList.push(forestClientNumber ?? '');
             }
         });
+        if (returnResponse.data.email_sending_status == EmailSendingStatus.SentToEmailServiceFailure) {
+            setNotificationMsg(
```
MCatherine1994, line 143: Did we test the situation when email failed to sent, and there is also an error when grant duplication access? I wonder if the error message will override each other...

server/admin_management/api/app/services/access_control_privilege_service.py
```diff
@@ -234,6 +232,7 @@ def send_email_notification(
                 return
```
MCatherine1994, line 232: do we want to return the status not required in this case?

server/backend/api/app/routers/router_user_role_assignment.py
```diff
@@ -24,7 +25,7 @@
 
 @router.post(
     "",
-    response_model=List[schemas.FamUserRoleAssignmentCreateResponse],
+    response_model=schemas.FamUserRoleAssignmentResponse,
```
MCatherine1994, line 28: Hi Ian, same here, maybe you can use this schema directly as it's imported

server/admin_management/api/app/routers/router_access_control_privilege.py
ianliuwk1019: Right, maybe I was confused that time from IDE, will remove it.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
ianliuwk1019: Hm... I think you are right, I didn't corelated the case. For too long, I think I am losing frontend sense even when I saw previous lines "errorList" and was confused. Need to look into it. Thank you so much!!

server/admin_management/api/app/services/access_control_privilege_service.py
ianliuwk1019: That should be good, the default is not_required in the variable.

server/backend/api/app/routers/router_user_role_assignment.py
ianliuwk1019: Removed.

frontend/src/components/grantaccess/GrantDelegatedAdmin.vue
ianliuwk1019: Ticket #1532 is added for dealing with this message overriden issue.

MCatherine1994:
thanks so much Ian!

MCatherine1994 (approved):

Issue #1559, opened 2024-08-21
SchemaSpy workflow permission error when triggered by dev or prod deployment


MCatherine1994:
**Describe the task**
We're getting a permission error when schema spy workflow was triggered by the dev and prod deployment.

![Image](https://github.com/user-attachments/assets/0e2317e4-5ca6-4c0a-8a04-47cc18ce6f0e)

It runs successfully when trigged manually

**Acceptance Criteria**
- [ ] Fix the pipeline issue to allow run schema spy when do prod deployment

**Additional context**
- Add any other context about the task here.
- Or here

PR #1560, opened 2024-08-22
fix: #1559 Fix SchemaSpy trigger in deployment pipeline

MCatherine1994:
refs: #1559


.github/workflows/dev_deployment.yml
```diff
@@ -10,7 +10,7 @@ on:
 # permission can be added at job level or workflow level. Ref: https://github.com/aws-actions/configure-aws-credentials#OIDC
 permissions:
   id-token: write # This is required for requesting the JWT
-  contents: read # This is required for actions/checkout
+  contents: write # This is required for actions/checkout
```
MCatherine1994, line 13: Since we need to update this permission to be "write", I feel more safe to do this in dev deployment pipeline instead of the prod one. So the data model generator will be trigger every time when we merge to main.

ianliuwk1019: Sounds good. 
The error seems to be  from the last step of `reusable_data_model_gen.yml`. Is it possible this `permissions: contents: write` be inside the `reusable_data_model_gen.yml` before `jobs:`?

MCatherine1994: I tried that, but it returns me the error, I think if the dev_deployment config permission to be read, I can't do permission write in the data model gen yaml

MCatherine1994: Screen Shot 2024-08-22 at 11 27 31 AM

ianliuwk1019: Is it because when you use ci_infrastructure.yml to tested it out, it already specifically set to `contents: read` but for dev_deployment.yml, this `content` has not been set yet so can be set in reusable yml jobs?

MCatherine1994: dev_deployment also has content set to be read already

ianliuwk1019: I see, I must be looking at the wrong one that did not have it.
Thanks!

ianliuwk1019 (approved):

Issue #1242, opened 2024-03-14
Add display Name for roles


gormless87:
**Describe the task**
Add a database column to store a role display name

**Acceptance Criteria**
- [x] Add db column (100 character limit) for role display name - nullable
- [x] Populate the new column for existing roles (to be added to this ticket)
- [x] Ensure child roles use parent abstract role display name as their display name
- [x] Update ORM model(s) for fam_role for new column
- [x] Update the return results of return-my-access-api to include role display name and role purpose
- [x] Update get-user-role-assignment, ~~get-app-admin~~, get-delegate-admin to include role display name 
- [x] Add/update test if need

**Additional context**
- Proper English role name to display in UI
-  #1128 is for display
- column needs to be nullable

mockup:

![Image](https://github.com/bcgov/nr-forests-access-management/assets/121979429/0c3626e6-4648-46c5-9792-2f668aa9be09)

ianliuwk1019:
From database we have these roles across applications.
We don't have role display names for them.
Would these look fine (column two below):

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/po-yi.liu/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/po-yi.liu/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">



</head>

<body link="#467886" vlink="#96607D">


role_name |  
-- | --
Admin | Admin
Approver | Approver
CLIENT_ADMIN | Client Admin
CLIENT_EDITOR | Client Editor
CLIENT_VIEWER | Client Viewer
FOM_REVIEWER | Reviewer
FOM_SUBMITTER | Submitter
Planner | Planner
SPAR_MINISTRY_ORCHARD | Ministry Orchard
SPAR_NONMINISTRY_ORCHARD | Non-ministry Orchard
SPAR_TSC_ADMIN | TSC Admin
Submitter | Submitter
Viewer | Viewer



</body>

</html>

@OlgaLiber2

OlgaLiber2:
[...]

PR #1564, opened 2024-08-23
feat: #1242 Add role display name

ianliuwk1019:
ref #1242 
* Add flyway V54 for new role's display_name column.
* V54 flyway migration script to update existing roles for display_name column.
* Add new role display_name column to model at server/backend and server/admin_management.
* Adjust schema to include display_name and role_purpose for role creation and role retrieval information.
* Set display_name value when creating child_role.
* Return display_name value and role_purpose value for endpoints.

image
image
image


server/flyway/sql/V54__add_role_display_name.sql
```diff
@@ -0,0 +1,112 @@
...
+
+-- Migrate existing roles to set display_name below; for all 'C' and 'A' role types and all child roles.
+
+-- CLIENT application roles: ('CLIENT_ADMIN%', 'CLIENT_EDITOR%', 'CLIENT_VIEWER%');
+-- 'CLIENT_ADMIN' (role_type = 'C')
+UPDATE app_fam.fam_role SET display_name = 'Admin'
+    WHERE role_id in (
+    SELECT role.role_id
+    FROM app_fam.fam_role role
+    WHERE role.role_name like 'CLIENT_ADMIN%');
```
MCatherine1994, line 14: Hi Ian, wondering do we need "%" at the end for concrete role? thanks!

ianliuwk1019: Ha, no, but it does not matter actually; if concrete role, it will just find the concrete role id (only 1).  
I give every script the same to be easier to write.

MCatherine1994 (approved):

PR #1578, opened 2024-08-29
fix: ignore env files

craigyu:
Removed env files, it's too much of a risk to have it in the repo to have things accidentally committed.

Updated realted readme and added a doc so new devs can find the local-dev.env template [there](https://github.com/bcgov/nr-forests-access-management/wiki/Local-Env-Vars-Template)
image

Removed the default setting that requires vscode users to double click to open a file


DerekRoberts (approved):
Big yes!  So much better.  :)

DerekRoberts (approved):
Lots of improvements here!  Keep it up.

Issue #1580, opened 2024-08-30
de-spaghetti-fying backend schema


craigyu:
### Description:
As a developer, I find it cumbersome to scroll through hundreds of lines of code to locate a specific model or schema that needs editing.

Currently, all our models are defined in a single file, `model.py`, which exceeds 700 lines. Similarly, all schemas are consolidated into a single file, `schemas.py`, with over 300 lines.

As our project continues to grow, this approach will become increasingly difficult to maintain and navigate. Adding new definitions to these large files will further complicate the structure, making it challenging to manage, debug, and understand the codebase.

### Action:
- Refactor the server/backend structure so that each model and schema is placed in its own file.

### Update:
This issue will focus on schema files only

### Acceptance Criteria:
- The application should function as expected with all tests passing after the refactoring.
- The codebase should be easier to navigate and maintain, with clearly organized and modular files for models and schemas.

craigyu:
A decision has been made to leave `models/model.py` unchanged due to our plan to use `sqlacodegen`, which can only output models into a single file.

We decided to split only the schemas into individual files, as this has no downstream effects.

PR #1581, opened 2024-08-31
fix: #1580 models refactoring

craigyu:


craigyu:
single model files wins

PR #1582, opened 2024-08-31
fix: deconstructing backend models

craigyu:


DerekRoberts:
@craigyu Holy crap!  This is a huge improvement.  I look forward to seeing where you take this app.  :)

PR #1586, opened 2024-09-05
fix: #1580 de-spaghetti-fying backend schema

craigyu:
Separate schemas into individual files


server/backend/api/app/schemas/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+    "FamUserRoleAssignmentResponseSchema",
+    "IdimProxySearchParamSchema",
+    "IdimProxyBceidSearchParamSchema",
+    "IdimProxyIdirInfoSchema",
+    "IdimProxyBceidInfoSchema",
+    "GCNotifyGrantAccessEmailParamSchema",
+    "ForestClientIntegrationFindResponseSchema",
+    "RequesterSchema",
+    "TargetUserSchema",
+]
```
ianliuwk1019, line 84: Hi Craig, no sure if there is a concerns later on for housekeeping these imports in __init__, but do feel cool to learn new things.
Thanks. As discussed, would you mind to remove the "__all__" if you also feel not necessary?

craigyu: image

removed, although pylance isn't happy, but the file is smaller to look at

ianliuwk1019 (approved):
Thanks a lot Craig!! Although feel seems a bit too many individual files (1 class per file) in the schema folder, but we can go from here for now.

Issue #1593, opened 2024-09-17
(PROD) Email sending fail when granting user permission.


ianliuwk1019:
**Describe the bug**
Client team Maria reported an issue when granting user permission with  “Send email to notify user” checkbox checked. When email checked is not check, the granting flow ran successfully.

![image](https://github.com/user-attachments/assets/5eae9a41-1e59-4c63-8dd5-c0505d88e4f4)

**Acceptance Criteria**
- [x] hotfix to fix issue.
- [x] deploy to production and verified.
- [x] notify client.
- [ ] merge hotfix to main branch.

**Additional context**
- Tried FOM (DEV) on production with email checkbox checked to assign other user and it worked.

PR #1599, opened 2024-09-18
fix: #1593 gc notify app name max length

craigyu:
Adjusted app_name's max len from 35 to 100


craigyu:
will resolve the conflicts later

ianliuwk1019 (approved):
Thanks for merging back!

Issue #1596, opened 2024-09-17
Implement Email Styling Enhancements


craigyu:
**Describe the task**
Spike done in #1552 
Make the email template more scannable

**Acceptance Criteria**
we want to:
- [x] show roles as the display name (not db names) for example, "Submitter", "Reviewer", etc 
- [x] add organization names associated with client numbers 
- [x] add a bullet for each role + client number 

mockup for the desired changes: 
![Image](https://github.com/user-attachments/assets/c11fdde1-07a4-4bd8-a8b3-346eaec69bc2)

![Image](https://github.com/user-attachments/assets/c7086df1-42d7-4ed0-b5df-27ccd7d8ec02)

[Context](https://github.com/bcgov/nr-forests-access-management/issues/1552#issuecomment-2334604061):
GC Notify's reply (Their respond speed is impressive!).

```
Hi Ian,

Unfortunately, we don't natively support what you are trying to achieve. Also, we don't support HTML content in templates or variables. 

However, if you are using the API, you could dynamically create the content of a variable in your own code to fit the what you need. You would put in a variable the desired content using the same formatting than you would in a template. 

In your code, you build something like this variable = "* Item 1\n* Item 2\n* Item 3" (dynamically built from the number of items you have). Use the \n new line character between your items. 

And in the template, you would put ((variable)) which will take the content you pass it from the API. 

JSON payload:

{
  "email_address": "recipient@email.com",
  "template_id": "aaaaaa-bbbb-ccc-dddd-11111111111",
  "personalisation": {
    "variable": "* Item 1\n* Item 2\n* Item 3"
    }
}

Let us know if this works. 

Regards,
Jimmy P.
The GC Notify Team
```

ianliuwk1019:
Hi @OlgaLiber2 , do you have mockup for the FOM Submitter role with organization?
Thanks.

craigyu:
Possible client info formats

CanFor Inc (0074124214)

CanFor Inc [0074124214]

0074124214 - CanFor Inc

CanFor Inc - 0074124214

OlgaLiber2:
@ianliuwk1019 I updated the mockup. Let me know if you have any questions.

ianliuwk1019:
Hi @OlgaLiber2 , in delegated admin email template enhancement, do you intend to remove this statement (it was in first version)?
`As a delegated admin, you are required to comply with our terms and conditions.`

ianliuwk1019:
[...]

PR #1608, opened 2024-09-26
feat: #1596 Implement email styling enhancements

ianliuwk1019:
Closes: #1596 
* update frontend T&C component to add some note.
* on GC Notify, create _V2 templates for both regular user template and delegated admin template.
* update both "server/backend" and "server/admin_management" on GCNotifyEmailService to 
  reflect the template id change and params change.
* update dynamic wording in code for templates parameters.
* save latest version to project "public" folder to be accessible as a link, and use it for delegated admin template.
* update granting user and granting delegated admin method to provide data for email params.
* adjust email template to show scoped client as a point lists.
* minor other email template adjustments on wording and spacing.


server/admin_management/api/app/utils/utils.py
```diff
@@ -28,3 +29,10 @@ def remove_app_env_suffix(name: str):
         if name.endswith(suffix):
             return name[: -len(suffix)]
     return name
+
+
+def is_success_response(response: Response):
+    SUCCESS_LIST = [
+        HTTPStatus.OK, HTTPStatus.CREATED, HTTPStatus.ACCEPTED, HTTPStatus.NO_CONTENT
+    ]
```
craigyu, line 37: I prefer to have this in the constants file, but this is fine too

craigyu (approved):
Good work, i didn't test it locally since it was tested already

server/admin_management/api/app/utils/utils.py
ianliuwk1019: Thanks. This I don't feel it will be used by other, for now just here.

ianliuwk1019:
> Good work, i didn't test it locally since it was tested already

No problem, definitely need to be tested once it is merged to dev.

Issue #1614, opened 2024-10-03
Add new role for CLIENT


OlgaLiber2:
**Describe the task**
Add a suspend role for CLIENT

**Acceptance Criteria**
- [ ] role is added in TEST and DEV environments
- [ ] display name should be "Suspend"

**Additional context**
![Image](https://github.com/user-attachments/assets/d92c47b1-b0fc-4bdd-9e24-4537f59ddb20)

ianliuwk1019:
@OlgaLiber2 Is this a concrete role (C) or abstract role (A)?

OlgaLiber2:
@ianliuwk1019 it would be a concrete role that does not require client numbers.

PR #1618, opened 2024-10-03
fix: #1614 add client suspend role

craigyu:
Added a new role for Forest Client. tested manually on local


ianliuwk1019 (approved):
Nice insert query for multiple!!

Issue #1545, opened 2024-08-16
Store audit records when granting/removing delegated admin access


MCatherine1994:
**Describe the task**
Store the audit records when granting or removing delegated admin access. 

**Acceptance Criteria**
- [x] Update the existing create access control privilege api to add the record in audit table when creating a delegated admin
- [x] Update the existing delete access control privilege api to add the record in audit table when removing some access for a delegated admin
- [x] Add the forest client name when create the audit record, for creating access control privilege, we already verify forest client number, so we can get the forest client name. For deleting, we need to search the forest client number as well, if we can't find, then either return unknown forest client name or fail the transaction
- [x] Add tests

**Additional context**
- Epic story 2 #1549
- Create one audit record when create delegated admin with multiple organizations

ianliuwk1019:
At this ticket, the frontend user permission history page is also minor adjusted to show delegated admin content for role change.
After this ticket is done, we can also close #1546 , #1548.

PR #1625, opened 2024-10-07
feat: #1545 store audit records when grantingremoving delegated admin access

ianliuwk1019:
- move server/admin_management schema file into folder schemas and refactor some imports.
- add permission change audit relatled ORM mapping to model at server/admin_management.
- add permission_audit_history.py to hold permission aduit history related schema objects.
- add permission audit 'repository' and 'service' and the logic to store delegated admin permission change operation history.
- add tests
- adjust frontend permission history page to properly display role details for delegated admin based on mockup.

---------------------------------------
- The server/admin_management "schema" file is git moved to be under "schemas" folder. It is intended in the future to further refactor into groups of smaller schema files. Quite some files are only changed for the imports but the functionalities were not touched. These files can be ignored.


frontend/src/components/UserPermissionHistoryTable/PermissionDetailsCol.vue
```diff
@@ -1,62 +1,77 @@
...
 
 const props = defineProps<{
     permissionDetails: PrivilegeDetailsSchema;
+    permissionChangeType: String;
 }>();
 
+enum PermissionChangeType {
+    GRANT = 'GRANT',
+    REVOKE = 'REVOKE'
+};
```
craigyu, line 15: Let's get client-code-gen to give us this value instead of defining here. It might not change often but it's a good idea to make the front-end somewhat independent 

We might need to redefine some schemas

image

craigyu (changes requested):
Very nice work! Tested and demo-ed! Left one change request for maintainability

frontend/src/components/UserPermissionHistoryTable/PermissionDetailsCol.vue
ianliuwk1019: No problem, initially I thought about the same thing too but didn't get to it at the backend, this is now adjusted. Thanks.

craigyu (approved):
Good stuff!

Issue #1626, opened 2024-10-08
Paginate GET user_role_assignment endpoint


craigyu:
**Description**

Paginate the results displayed on the Users Table
![Image](https://github.com/user-attachments/assets/fe402c5d-17b4-43ef-b813-0498d4021e7e)

Endpoint design:
- Updating the endpoint name to use dashes instead of underscores
    - `/fam_applications/{application_id}/user_role_assignment` to `/fam-applications/{application_id}/user-role-assignment`
    - `/fam_applications/{application_id}/users/{user_id}` to `/fam-applications/{application_id}/users/{user_id}` 
    - Refactor areas that use these endpoints, e.g. tests, front-end
- Default: `/fam_applications/{application_id}/user_role_assignment`
    - Will sort by user name, page = 1, page size = 50
- with pagination: `/fam_applications/{application_id}/user_role_assignment?page={pageNumber}&size={pageSize}`
    - if one of them is missing, then use the default values, e.g. if page number is 1 and no pageSize is provided, then use the default 50 for pageSize
- with Sort `/fam_applications/{application_id}/user_role_assignment?sortBy={columnName}`
    - should reject a call if an invalid column name is provided, 400 Bad request
    - default page number and page size should apply if not present
- with Search `/fam_applications/{application_id}/user_role_assignment?search={keyWord}`
    - should be able to encode and decode search string, e.g. a space would be %20
    - default page number and page size should apply if not present
- maintain the current privileges 
- should return an `x-total-count` header
- return an Empty list `[]` if there's no hit

**Acceptance Criteria**
- [ ] An `x-total-count` header 
- [x] A page number that starts with 1 defaulted to 1, a page size defaulted to 50, with a max limit of 100 items per page.
- [x] Sort by columns, default sort by username, should be able to sort by all columns
- [x] Search by keyword on columns [ username, full name, email, client number, Role display_name]
- [x] Make a PR to a feature branch

**Additional Context**
We will be adding forest client name to this endpoint later. 
This will break frontend's user table if deployed, so it should be deployed to a feature branch, the feature branch will be merged to `main` once frontend is ready

basilv:
This story should result in addressing defect #1486

craigyu:
decided to remove pagination metadata from header and put them into response body

PR #1634, opened 2024-10-22
feat: #1626 Get user role assignment endpoint backend pagination

ianliuwk1019:
Closes: #1626 
* Create page params schema for pagination, sorting, filtering; and paged schema for returing results and meta information.
* Update model.py to utilize SqlAlchemy to get full_name as hybrid column for sorting.
* Add simple paginate_service to filtering, ordering and pagination from base query and return paginated result with meta information.
* Adjust "fam_application" router path to "fam-application".
* Config`/{application_id}/user-role-assignment` endpoint to receive page_params with `PagedResultsSchema` response model return.
* Apply paginate_service to GET application-user/role query to get paged result. 
* Do client-code gen for user/role assignment endpoint change.
* Fix tests due to endpoint change.

<hr>

image
image

* With default pagination on application large volume data:
image

* With total pages = 204, but paging on page=205:
image

* Starts on second page:
image

* Sort by user_name descending:
image

 * Search keyword="111" on second page:
 image

* While FOM_DEV having 3 user-role assignments, search on keyword "submitter"
image
image

* **[Verify as a BCEID delegated admin]**:
   With BCEID user "TEST-3-LOAD-CHILD-1" is FOM_DEV delegated admin (FOM_REVIEWER, FORM_SUBMITTER_00001018), and with the same user-role assignment setup previously:
image
image

-- search for "submitter" should see nothing:
image

-- search for "reviewer" should see 1 record:
image


server/backend/api/app/constants.py
```diff
@@ -103,6 +104,19 @@ class EmailSendingStatus(str, Enum):
...
 APPLICATION_DESC_MAX_LEN = 200
 CREATE_USER_MAX_LEN = 100
+MIN_PAGE = 1
+DEFAULT_PAGE_SIZE = 50
+MIN_PAGE_SIZE = 10
+
+# The intent is  to have "max=100" per page, however frontend is not ready, so if need to return "all records" found
+# and let frontend do the pagination, we could set it to 100000.
+# >> MAX_PAGE_SIZE = 100
+MAX_PAGE_SIZE = 100000
```
ianliuwk1019, line 114: This is temporary for 100000/per page since frontend isn't ready for backend pagination, so frontend can still get all records for now.

server/backend/api/app/schemas/pagination.py
```diff
@@ -0,0 +1,66 @@
...
+    page_number: int | None = Field(Query(default=1, ge=1, description="page number to get the paged data"))
+
+More over, the endpoint also need to use "Depends()" for function argument, like below exampe:
+    def get_fam_application_user_role_assignment(
+        application_id: int,
+        page_params: PageParamsSchema = Depends(),
+        ...
+    )
+
+And, in this case, the validation error default from Pydantic will not be 400, it will be 422 (from Pydantic)
```
ianliuwk1019, line 29: Note, currently built in validation errors from Pydantic is 422. Hope that's acceptable for frontend.

craigyu (changes requested):
Impressive work! The endpoint works as intended.

However, could you provide the available options on the swagger doc for the sortBy and sortOrder? 
Something like this
image

It's easier for devs when they need to use this endpoint. I know sortBy is very trivial but it's easier without having to guess whether it's `desc` or `dsc` or something else

ianliuwk1019:
> Impressive work! The endpoint works as intended.
> 
> However, could you provide the available options on the swagger doc for the sortBy and sortOrder? Something like this image
> 
> It's easier for devs when they need to use this endpoint. I know sortBy is very trivial but it's easier without having to guess whether it's `desc` or `dsc` or something else

Ah, good point! I will let you know if I can add them easily.

ianliuwk1019:
> 9a87f9d



> Impressive work! The endpoint works as intended.
> 
> However, could you provide the available options on the swagger doc for the sortBy and sortOrder? Something like this image
> 
> It's easier for devs when they need to use this endpoint. I know sortBy is very trivial but it's easier without having to guess whether it's `desc` or `dsc` or something else

I can't find an easy way from FastAPI to add additional information in a nice way. I managed to do that at "description" field using <br> instead:
image
See if you think this is good.

craigyu (approved):
thanks for adding them!

Issue #1641, opened 2024-10-25
Delete users that are deleted from IDPs


craigyu:
If a user is deactivated, allow their user role assignments to be deleted

![Image](https://github.com/user-attachments/assets/92bb2ce9-a9ed-4e82-916c-b93d5b947d18)

Acceptance Criteria:
- [x] Inactive user's role assignments can be deleted 
- [ ] Test with BCeID (if possible) and IDIR 
- [x] Should fix all 3 DELETE endpoints

ianliuwk1019:
Ref log (prod):
![Image](https://github.com/user-attachments/assets/bd63b266-8d7d-45a8-baf1-dbb29c8d246f)

ianliuwk1019:
Looks like it is due to a router guard is using a validator which is calling external IDIM service to get back user's information for checking (on target_user, in this case 'RDECAMPO') .
![Image](https://github.com/user-attachments/assets/4021b0d8-fc08-4e72-a527-7deb18974189)

This router_guard "**enforce_bceid_by_same_org_guard**" is needed for both CREATE and DELETE user_role_assignment endpoint to enforce if a BCeID user is trying to delete another user, it can only delete the user belongs to the same organization.

PR #1661, opened 2024-11-19
fix: #1641 Allow delete inactive users

ianliuwk1019:
Closes: #1641 
* Adjust `get_verified_target_user` router guard so that only BCeID requester
  needs to retrieve target user information from IDIM servic eto verify this user.
  This allows IDIR requester to be able to delete inactive users and BCeID requester
  to have appropriate error message when not able to delete inactive user.
  
* Add tests for the scenarios.


server/backend/api/app/routers/router_guards.py
```diff
@@ -412,21 +412,46 @@ async def enforce_bceid_by_same_org_guard(
...
+            target_user = get_verified_target_user(requester=requester, target_user=target_user, role=role)
+        except Exception as e:
+            LOGGER.error(f"Target user could not be verified: {e}.")
+            app_name = role.application.application_name
+            error_msg = f"Unable to verify user {target_user.user_name}. Please contact {app_name} administrator for the action."
+            utils.raise_http_exception(
+                status_code=HTTPStatus.INTERNAL_SERVER_ERROR,
+                error_code=ERROR_CODE_UNKNOWN_STATE,
+                error_msg=error_msg,
+            )
```
craigyu, line 453: I prefer the `424 Failed Dependency` status code since I associate 500 with the server bugging out unexpectedly, here the doc from mozilla.org

> The HTTP 500 Internal Server Error [server error response](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) status code indicates that the server encountered an unexpected condition that prevented it from fulfilling the request. This error is a generic "catch-all" response to server issues, indicating that the server cannot find a more appropriate [5XX error](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses) to respond with.

craigyu (commented):
All 3 delete actions tested working locally with `FAKE_LOAD_TEST_11498`, but I see there's no changes in the /admin_management folder, it has a `async def get_verified_target_user`, does it need to be updated as well?

ianliuwk1019:
> All 3 delete actions tested working locally with `FAKE_LOAD_TEST_11498`, but I see there's no changes in the /admin_management folder, it has a `async def get_verified_target_user`, does it need to be updated as well?

Hi Craig,
For deletion from /admin_management, there is no scenarios to fix like the fixed one that requires verifying user externally, so there is no change. Thanks for testing them all as well! Good observation, I did remove one `async` but didn't notice others (lazy..). They were probably copy-paste initially left over there. I remove them, thanks for mentioning that!

craigyu (approved):

Issue #1665, opened 2024-11-20
FAM prod release


ianliuwk1019:
Release FAM Prod:
- Frontend refactoring.
- Delegated admin + BCeID enabling.
- Reminder: Recover SILVA allowed callback URLs manually at Cognito after prod deployment (ref: #1653)
- Scroll bar fix

**Additional context**
- Verify on FAM TEST before prod release.

PR #1666, opened 2024-11-20
fix: #1665 prod release fixes

craigyu:
- Scroll bar visual issue

### DIFF
--- a/frontend/src/router/index.ts
+++ b/frontend/src/router/index.ts
@@ -13,15 +13,16 @@ export const router = createRouter({
         }
     }),
     scrollBehavior(to) {
-        // Check if the route has the `protectedLayoutMeta`
         if (to.meta?.layout === "ProtectedLayout") {
-            return {
-                top: 0,
-                behavior: "smooth",
-                el: "#proctected-layout-container",
-            };
+            const mainElement = document.querySelector(".main");
+            if (mainElement) {
+                mainElement.scrollTo({
+                    top: 0,
+                    behavior: "smooth",
+                });
+            }
+            return false; // Prevent default Vue Router scrolling behavior
         }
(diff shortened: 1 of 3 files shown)


frontend/src/router/index.ts
```diff
@@ -13,15 +13,16 @@ export const router = createRouter({
...
-                el: "#proctected-layout-container",
-            };
+            const mainElement = document.querySelector(".main");
+            if (mainElement) {
+                mainElement.scrollTo({
+                    top: 0,
+                    behavior: "smooth",
+                });
+            }
+            return false; // Prevent default Vue Router scrolling behavior
```
ianliuwk1019, line 24: This is interesting, new to me.
Not sure if we need to do all of these element.scrollTo.
I saw it provides this "position" to this `scrollBehavior()` callback:
image
image

ianliuwk1019 (approved):
Approved, thanks Craig! But just a small comment that I saw the doc can specify pos.

frontend/src/router/index.ts
craigyu: yeah we could utilize the savedPosition in the future if we need to go back to the same position on a page, currently we don't have this requirement. 
Here's the reason for using `scrollTo` 

There's a issue of this not working if we just do
```
return {
                top: 0,
                behavior: "smooth",
                el: ".main",
            };
```

because overflow-y: auto or scroll sets the scrolling behavior to the .main container, and Vue Router’s scrollBehavior typically works with the window’s scroll position, not custom scrollable elements. To make Vue Router’s scrollBehavior work with .main, we need to manually scroll the .main element instead of relying on the default window scrolling.

ianliuwk1019: No problem. Thanks for clarify this for my understanding!

PR #1671, opened 2024-11-22
fix: #1657 move the add permission button

### LINKED ISSUE #1657: Move the add permission button
## Description
Move the add button for user and delegated admin table from the table toolbar to the top right of the screen

## Acceptance criteria
- [ ] Move the add button for user and delegated admin table from the table toolbar to the top right of the screen
    - Image
- [ ] Add subtitle to permission history page
- [ ] Add "Added On" column

craigyu:
Closes #1657

- Added subtitle to the Permission History page
- Moved the add button from table tool bar to the app dropdown row
- Added the Added On column
    - default sort on FAM App Admin table is by username and ascending
    - default sort on other tables is by create_date and descending


frontend/src/views/ManagePermissionsView/index.vue
```diff
@@ -215,6 +215,36 @@ const addNotifications = (newNotifications: PermissionNotificationType[]) => {
...
+    else if (activeTabIndex.value === 1) {
+        router.push({
+            name: AddAppPermissionRoute.name,
+            query: {
+                requestType: "addDelegatedAdmin",
+                applicationId: selectedApp.value?.id,
+            },
+        });
+    }
+};
```
ianliuwk1019, line 246: I guess there are not many tabs, so "0", "1" might be clear, but wondering if you like to associated the index "0" -> "App users table" and index "1" -> "Delegated admin table". such as:
```
export enum MPermissionTableTabsIndex {
  AppUsersTab,   // <-- 0 index
  DelegatedAdminTab  // <-- 1
}
```
And use the enum for the `activeTabIndex.value` state.

I guess also two `else if` can be combined as (however, separated is easier to read too):
```
    else {
        router.push({
            name: AddAppPermissionRoute.name,
            query: {
                requestType: activeTabIndex.value === 0? "addUserPermission" : "addDelegatedAdmin",
                applicationId: selectedApp.value?.id,
            },
        });
    }
```

frontend/src/views/UserDetails/index.vue
```diff
@@ -9,6 +9,10 @@ import UserPermissionHistoryTable from "@/components/UserPermissionHistoryTable"
 import BreadCrumbs from "@/components/UI/BreadCrumbs.vue";
 import type { BreadCrumbType } from "@/types/BreadCrumbTypes";
 import { ManagePermissionsRoute } from "@/router/routes";
+import { useQuery } from "@tanstack/vue-query";
+import { AdminMgmtApiService } from "@/services/ApiServiceFactory";
+import { AdminRoleAuthGroup } from "fam-admin-mgmt-api/model";
```
ianliuwk1019, line 14: Seems not used.

frontend/src/views/UserDetails/index.vue
```diff
@@ -28,6 +32,19 @@ const navigateBack = () => {
...
+ * it is used to display the app name in subtitle.
+ */
+const adminUserAccessQuery = useQuery({
+    queryKey: ["admin-user-access"],
+    queryFn: () =>
+        AdminMgmtApiService.adminUserAccessesApi
+            .adminUserAccessPrivilege()
+            .then((res) => res.data),
+    select: (data) => getApplicationById(Number(applicationId), data),
+});
```
ianliuwk1019, line 46: No change, Just wonder, do we fetch this every time since I see it at several places or do we intend to fetch every time? Or is this cached?

ianliuwk1019 (approved):
Just few comments but looks good and approved. Thanks!

frontend/src/views/ManagePermissionsView/index.vue
craigyu: https://github.com/bcgov/nr-forests-access-management/pull/1671/commits/058d51c8c0bd75bf16b7653698644368ba7a1e22

frontend/src/views/UserDetails/index.vue
craigyu: if there's a cached value it will be using the cached data, if there isn't a cached value, e.g. after a refresh, then the data will be automatically fetched again. 

However, in manage permission view, this data is refetched on every mount with 
```javascript
refetchOnMount: true,
```

This refetch (when there's a cached data already) will not trigger the skeleton of the dropdown to show, since it's the skeleton will only show when `query.isLoading === true`, there's a small difference between `query.isFetching` and `query.isLoading`, loading indicates that this the first data load (no cached value), fetching is true whenever the data is fetching, meaning if loading is true then fetching is true, but fetching is true doesn't mean loading is true. This allows us to silently refresh the data

frontend/src/views/UserDetails/index.vue
craigyu: https://github.com/bcgov/nr-forests-access-management/pull/1671/commits/274eccd8df4387a6030070e578ba5b6f86518c56

ianliuwk1019 (approved):

Issue #1673, opened 2024-11-26
Display IDIM error in grant access pages


craigyu:
display an warning notification in the following grant access pages
- [ ] Grant app admin
- [ ] Grant app user 
- [ ] Grant d-admin

ianliuwk1019:
This is a temporary solution to deal with immediate production issue due to not able to search users through external IDIM Web Service (incident happen on or starting few days before 2024-11-25).

PR #1674, opened 2024-11-26
fix: #1673 add IDIM warn notification message

craigyu:
Screenshot 2024-11-26 at 11 07 42 AM

image

image


ianliuwk1019 (approved):
That was really good quick one, thanks Craig!!

Issue #1675, opened 2024-11-27
[Pending-Until IDIM Confirmed] Fix FAM search IDIR user scenario


ianliuwk1019:
**Describe the bug**
Since some time last week (2024-11-21) and starting 2024-11-25, FAM IDIR search stop working.
This is caused by IDIM proxy IDIR search user issue reported on this ticket: https://github.com/bcgov/nr-fam-idim-lookup-proxy/issues/132
The ticket above will adjust request param to IDIM proxy for "requester" to use "User GUID instead of "User ID".

FAM project will need to adjust this accordingly to pass requester's "User GUID" for IDIR search to work.

**Affect:**
- All environments.

**Acceptance Criteria:**
- Adjust both backends idim_proxy.py `**search_idir**` method to pass requester's "**User GUID**".
- Make sure existing tests pass.
- Deploy and verify 

**Additional context**
- Dependency to ticket: https://github.com/bcgov/nr-fam-idim-lookup-proxy/issues/132. Wait until at least it is deployed to TEST to be able to verify.

PR #1676, opened 2024-11-27
fix: #1675 (b) Temporary fix FAM search IDIR user scenario by switching to 'idir-account-detail` new IDIM-proxy endpoint.

ianliuwk1019:
WIP...
- This has dependency on pr: https://github.com/bcgov/nr-fam-idim-lookup-proxy/pull/133


craigyu (approved):
Thanks Ian, i removed the warning banner here, hope you don't mind :P

Issue #1638, opened 2024-10-25
Add client details to GET user_role_assignment endpoint


craigyu:
**Description:** Add ClientDetails for each FamApplicationUserRoleAssignmentGetSchema in the paginated result of GET user_role_assignment endpoint

**Acceptance Criteria**
- [x] Integrate the /api/clients/search endpoint from Forest Client API 
- [x] Add Client name property to response body
- [x] modify frontend column to display client name and client number
- [x] sort by client number
- [x] Change column name from `Client number` -> `Organization`
- [x]  Add tests!

**Additional context**

ianliuwk1019:
FC API is reacting interestingly (TEST environment, `/api/clients/search`):
![Image](https://github.com/user-attachments/assets/ff1eea72-86c9-4566-8882-d8aa02d1fe52)

* Same (2) ids searching, it responds with ranging from 950ms to ~3s; the range seems a bit inconsistent, not sure if it is due to my network or on server end.
* It can accept the parameters with `&id=id1&id=id2` or `&id=id1,id2`. not sure if it is a flexible feature or is it something we need to watch out to choose one to use.
@craigyu @basilv

ianliuwk1019:
Hi @OlgaLiber2 , how do we like to add "client name" to frontend (table below)? I can't find the mockup. @craigyu 
![Image](https://github.com/user-attachments/assets/348bb537-ce76-4857-b9fd-94255fb16cb4)

Example:
Akieca Explorers Ltd. (**00001011**)
**00001011** (Akieca Explorers Ltd.)
**00001011** - Akieca Explorers Ltd.

OlgaLiber2:
@ianliuwk1019 my figma crashed and I need to reinstall. Can you reach out to Marina please? She will can point you to the mockup.

Issue #1639, opened 2024-10-25
Add client details to GET access_control_privileges endpoint


craigyu:
**Description:**  Add ClientDetails for each FamApplicationUserRoleAssignmentGetSchema in the paginated result of GET access_control_privileges endpoint

**Acceptance Criteria**
- [x] Integrate the /api/clients/search endpoint from Forest Client API
- [x] Add Client name property to response body
- [ ] ~~modify frontend column to display client name and client number~~
- [x] sort by client number
- [ ] ~~Change column name from `Client number` -> `Organization`~~
- [x] Add tests!
- [x] Adjust all router endpoints to follow convention using "-" for naming instead of "_" and do client-code-gen.
- [x] For FAM forest client "search" (GET /forest_clients/search), internally replace using "find_by_client_number" with "search" function just built in this ticket from "ForestClientIntegrationService" (to prevent confusion). Verify if that will cause frontend issue before replacing it (perhaps some error logic handling to the frontend).
- [x] Adjust endpoint `/admin-user-accesses` at response schema to return forest client name as well using this ticket's search ability.

**Additional context**

OlgaLiber2:
@craigyu we should sort by client name (not number), same for ticket 1638.

OlgaLiber2:
@ianliuwk1019 @craigyu if possible to do this ticket before other tickets in sprint backlog column that would be great.. this way we can have it deployed to prod sooner so Andrew & Rob can start working on the user manuals.

PR #1678, opened 2024-12-02
feat: #1638 Forest Client Search integration and show FC client name

ianliuwk1019:
Ref: #1638 
- Add `search` function to forest_client_integration.py to integrate with FC API Search endpoint.
- Create a `post_sync_forest_clients_dec` decorator to update get_application_role_assignments function return results 
  client_name (if available) through FC API Search endpoint.
- Adjust better naming for model's "client_number" relationship to "forest_client_relation" and adjust schema and other
  code impacted by the change.
- Client Code gen for the change of "client_number" object name to "forest_client".
- Adjust frontend ManagePermissionTable to display client name using `formatForestClientDisplayName`.
- Adjust exisiting `test_get_application_role_assignments_on_pagination` so tests can be tested without decorator (in isolation).
- Add tests for FC API Search integration testing.
- Add tests for new `post_sync_forest_clients_dec` decorator.


frontend/src/views/ManagePermissionsView/utils.ts
```diff
@@ -44,11 +37,7 @@ export const generateAppPermissionSuccessNotifications = (
             .map((filtered) => filtered.detail.role);
 
     const successClientList = successRoleList
-        .map((role) =>
-            isFamRoleWithClientSchema(role)
-                ? role.forest_client
-                : role.client_number
-        )
```
ianliuwk1019, line 51: Backend has changed to use `role.forest_client` for consistency, so no need to use `role.client_number` anymore (did not make sense for the previous name)

server/admin_management/api/app/models/model.py
```diff
@@ -245,7 +245,7 @@ class FamRole(Base):
 
     application: Mapped[FamApplication] = relationship("FamApplication", back_populates="fam_role")
 
-    client_number: Mapped['FamForestClient'] = relationship(
+    forest_client_relation: Mapped['FamForestClient'] = relationship(
```
ianliuwk1019, line 248: Minor variable refactoring on naming that make sense.

server/backend/api/app/models/model.py
```diff
@@ -617,7 +617,7 @@ class FamRole(Base):
     application: Mapped[FamApplication] = relationship(
         "FamApplication", back_populates="fam_role"
     )
-    client_number: Mapped[FamForestClient] = relationship(
+    forest_client_relation: Mapped[FamForestClient] = relationship(
```
ianliuwk1019, line 620: Minor variable refactoring on naming that make sense.

frontend/src/components/ManagePermissionsTable/index.vue
```diff
@@ -448,6 +449,33 @@ const highlightNewUserAccessRow = (
...
+                  forestClientData.forest_client_number,
+                  forestClientData.client_name
+              )
+            : "";
+    } else {
+        // For delegated admin data.
+        // TODO: No client name available from backend yet, implement soon. Only display client number. # noqa NOSONAR
+        return forestClientData?.forest_client_number;
+    }
+};
```
craigyu, line 478: Could you move this to the utils.ts under the same folder? so the file doesn't get too long

and perhaps rename it to getOrganizationName?

craigyu (changes requested):
Looks great Ian! Added one request.

Also let's change the `_` to `-` in the endpoint while we're add it :) 
image

frontend/src/components/ManagePermissionsTable/index.vue
ianliuwk1019: Thank for the suggestion Craig! This is moved.

ianliuwk1019:
> Looks great Ian! Added one request.
> 
> Also let's change the `_` to `-` in the endpoint while we're add it :) image

Hi Craig, for the swagger's 'search' endpoint, I have add this change of "_" to "-" in the ticket #1639 to do all the endpoints once and for all. Also that ticket will verify and swap internal calling to replace `find_by_client_number` if that's reasonable.

craigyu (approved):

PR #1694, opened 2024-12-11
feat: #1639 Forest Client Search integration with delegated admin endpoint forest client name update

ianliuwk1019:
* Refactoring both backend endpoints on router to replace path naming "_" with "-" and fixed affected frontend and backend.
* Remove no longer needed "find_by_client_number" method from both backend from Forest Client Integration. Use "search" instead.
* Fix tests affected by endpoints refactoring.
* Add Forest Client API Search integration to admin_management backend.
* Add forest client decorator to update forest client names for admin privilege grants endpoint and delegated admins endpoint return.
* Add tests for forest client integration and decorator.


craigyu (approved):
tested and working nicely, great work!

Issue #1647, opened 2024-11-05
Standardize datetime configuration on the backend


ianliuwk1019:
Several existing behaviors related to backend table models timestamp columns were observed:
- Tables' "**update_date**" (model.py) is empty during programmatically INSERT. We would like to have initial values as "**create_date**" during insert.
- Previously we didn't notice the **value** inserted into **timestamp** column using `datetime.datetime.now()` (in model.py) is a fixed value. This means the timestamp value is determined during model is loaded, and all the records inserted with this `datetime.datetime.now()` (in model.py) after backend model is loaded will have the same values and not the moment the records are inserted.  We will need to fix this.
- Most of the timestamp columns are timestamp **WITH TIME ZON**, but some tables have timestamp **_WITHOUT_ TIME ZON**. We need to make them consistent.

**Screenshots**
![image](https://github.com/user-attachments/assets/6a391da1-7240-4199-b66c-bb9c63262a66)
![image](https://github.com/user-attachments/assets/eb93dc3a-aba4-4caa-98e1-0f551970e72c)

**Acceptance Criteria**
- [x] Discuss if needs to have or fix "**update_date**" in model.py to have value during initial insert.
- [x] Fix timestamp columns for all tables to use Sqlalchemy `func.now()` than `datetime.datetime.now()`.
- [x] Use timestamp **WITH TIME ZON** for all timestamp columns.
- [x] Make sure all existing tests still pass.
- [ ] Document and close previous sqlacodegen tickets (on generating the model.py)

**Additional context**
- May need to discuss why we need to have value for the "**update_date**" during initial insert (so it is not empty when record is created), or if this is a good practice; and how to deal with current production empty "update_date" if we decide to have value in "update_date" initial during insert.

Related issue #1695 created to split off some refactoring of model.py

PR #1722, opened 2025-01-17
fix: #1647 Backend timestamp columns consistency and update_dates adjustment

ianliuwk1019:
ref: #1647 

- added V60 flyway for making timestamp columns in all model tables consistent:
	data type: "timestamp with time zone" (UTC),
	update_date: having initial value if empty as create_date; and set DEFAULT, NOT NULL if missing.
- used prefered SqlAlchemy dialects function 'func.now()' (for timestamp) than python 'datetime.datetime.utcnow' for setting 
  timestamp value and prevent it from having initial value using the time when model is loaded (should be dynamic, at the time when insert happens).
- added 'server_default', 'onupdate' and nullable=False (if missing) for "update_date".


server/flyway/sql/V59__timestamp_columns_consistency.sql
```diff
@@ -0,0 +1,185 @@
+-- This migration targets on making currently inconsistent 'create_date/update_date' to be consistent for their:
+-- * data type: TIMESTAMP WITH TIME ZONE (at UTC)
+-- * 'update_date' initial value: has the same intial value as 'create_date' and DEFAULT CURRENT_TIMESTAMP, with NOT NULL.
+
+
+-- >> -- Table: fam_access_control_privilege -- << --
+-- Change column type and convert existing values (with UTC timezone)
+ALTER TABLE app_fam.fam_access_control_privilege
+    ALTER COLUMN create_date TYPE timestamp with time zone USING create_date AT TIME ZONE 'UTC',
```
basilv, line 9: So my only significant question is what happens to the existing date values when alterting to UTC timestamp? I suspect that the result may cause some date shifting. E.g. 2025-01-17 converted to 2025-01-17 00:00:00 UTC is 2025-01-16 17:00:00 in BC timezone which is a date of 2025-01-16, one day earlier.

ianliuwk1019: Thanks Basil!! 
Either way of conversion from not timezone column to timezone column or vice versa change values.
That I was also having this concerns at the beginning, but had to change it anyway (because this is what the ticket needs). So had to use postgreSQL itself to convert for it. This depends on database itslef's server timezone (not the programming backend timezone). 

Only few tables having this needs of conversion. In production. When checking production db instance, it is actually in UTC zone:
image
So it means in this case, the existing value (without timezone) was in UTC. So the conversion will be the same but with timezone column it will have offset (+- 00:00). We should be good for the conversion.

ianliuwk1019: Sorry Basil, I probably said it wrong (not 100%), the db instance for the server on AWS probably does not affect db instance's timezone, it is the postgre db itself comes with default UTC timezone. So I am guessing even AWS has different region's availability, that does not affect where the db instance is and the timestamp column value should be the same (if db timezone does not change).

basilv (approved):

Issue #1754, opened 2025-02-15
Delegated admin data loading issue


OlgaLiber2:
**Describe the task**
A clear and concise description of what the task is.

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
I was onboarding Jill as a delegated admin for FOM on Friday. She only saw 14 people associated with her client number (00002176) in the user table while I see 28 users. 

see screenshots in our standup chat.

when she added a person that I could see but she couldn't, she got an error the user already exists and then she could see the user in her user table. 

No issues deleting users or performing other actions in FAM.

OlgaLiber2:
@ianliuwk1019 bug discovered on Friday. 
Craig will join our scrum on Tuesday to discuss.

craigyu:
Upon inspection, users who do not appear in the Delegated Admin’s users table are missing due to the absence of a `business_guid` in the fam_user table

basilv:
Scenario 1 - business BCeID user tries to log in to a FAM-protected app without any access and no user record. Auth-lambda creates the fam_user entry with null business GUID and null first/last name and null email. 

Scenario 2 - FAM admin grants access to a user from scenario #1 which results in an update of the fam_user entry. The first name, last name and email are _always_ populated, but the business GUID is only sometimes populated, otherwise remaining null.

Scenario 3 - FAM admin grants access to a new business BCeID user not in fam_user table yet. Similar to scenario #2, the first name, last name, and email are always populated, but the business GUID is only sometimes populated, otherwise remaining null. In cases where the business GUID remains null, subsequent IDIM lookup does find a business GUID and populates.

Most recent occurance of scenario 2/3 is from about 1 week ago, so this isn't old data.
We have ~278 user records missing business GUID.

basilv:
To check what's happening in the api lambda CloudWatch logs use the following search term which shows the log entries for updating the user entry: %Add first_name|Add business_guid|properties were updated%
There should be 3 log entries for updating name/email, business GUID, and generic update.

basilv:
Root cause (mostly) determined - see https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Missing+Business_GUID+for+Some+Business+BCeID+Users for analysis.

[...]

PR #1755, opened 2025-02-18
fix: #1754 Missing business_guid

basilv:
Update auth function to stop updating business_guid for existing users (since the attribute might be missing).
Also have the auth function update the update_user/update_date audit columns.

### DIFF
--- a/server/auth_function/lambda_function.py
+++ b/server/auth_function/lambda_function.py
@@ -188,7 +188,7 @@ def populate_user_if_necessary(db_connection, event) -> None:
         VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid},
         CURRENT_USER, CURRENT_DATE, CURRENT_USER, CURRENT_DATE)
         ON CONFLICT (user_type_code, user_guid) DO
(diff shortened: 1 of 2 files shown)


server/auth_function/lambda_function.py
```diff
@@ -188,7 +188,7 @@ def populate_user_if_necessary(db_connection, event) -> None:
         VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid},
         CURRENT_USER, CURRENT_DATE, CURRENT_USER, CURRENT_DATE)
         ON CONFLICT (user_type_code, user_guid) DO
-        UPDATE SET user_name = {user_name},  cognito_user_id = {cognito_user_id}, business_guid = {business_guid};"""
```
ianliuwk1019, line 191: Hi Basil, do you think we can still update `business_guid` using sql `CASE... WHEN`, so that if the Cognito event does provide `business_guid` for "some users" we still update the business_guid column, but if the business_guid is not present, then no update to the column?

basilv: I just went with a simple change to minimize the need for testing, so the fix can be deployed as quickly as possible.

ianliuwk1019 (approved):

Issue #1740, opened 2025-01-27
AWS RDS postgresql engine major version upgrade


ianliuwk1019:
**Describe the task**
AWS RDS postgres engine is currently at version 13.18 (https://github.com/bcgov/nr-forests-access-management/issues/1737).
We may like to plan for major version upgrade for the postgres engine.

As of 2025-01-27, the available versions are:
```
~ $ aws rds describe-db-engine-versions --engine postgres --query "DBEngineVersions[].EngineVersion" --output text
11.22   11.22-rds.20240418      11.22-rds.20240509      11.22-rds.20240808      11.22-rds.20241121      12.15   12.16   12.17   12.18   12.19   12.20   12.21   12.22   13.15   13.16   13.17   13.18   14.12   14.13   14.14   14.15          15.7    15.8    15.9    15.10   16.3    16.4    16.5    16.6    17.1    17.2

```
The highest being v17.2.

**Acceptance Criteria**
- [x] Discuss to pick a major version to upgrade.
- [x] Adjust `aurora-v2.tf` terraform file at block `"aws_rds_engine_version" "postgresql"` for the picked version.
- [ ] Deploy to DEV first (do manual backup before deployment?) and verify everything is working.
- [ ] Deploy to TEST -> PROD (do manual backup before deployment?) and verify everything is working.

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
Attempted to run deployment for 17.4 on DEV but failed for not finding the version. Strange, as AWS console does print that version out as available....
```
│ Error: no RDS engine versions found: &{DBParameterGroupFamily:<nil> DefaultOnly:<nil> Engine:0xc0028f9260 EngineVersion:0xc0028f9330 Filters:[] IncludeAll:<nil> ListSupportedCharacterSets:0xc00009e990 ListSupportedTimezones:0xc00009e991 Marker:<nil> MaxRecords:<nil> noSmithyDocumentSerde:{}}
│ 
│   with data.aws_rds_engine_version.postgresql,
│   on aurora-v2.tf line 42, in data "aws_rds_engine_version" "postgresql":
│   42: data "aws_rds_engine_version" "postgresql" {
```

basilv:
You might need to confirm that RDS allows upgrading from 13.18 straight to 17.4 in one step. However, the error message is unclear to me as to whether that's the issue or whether its some other issue...

Ahh, the problem is probably here:
[...]

PR #1760, opened 2025-02-26
fix: #1740 AWS RDS postgresql engine major version upgrade

ianliuwk1019:
#1740 
- adjusted docker db image to `17.4` (although not the same with db version at AWS aurora-postgresql enginer version).
- adjusted `aurora-v2.tf` for:
  - engine_version upgrade: 16.6 (from 13.8)
  - renamed parameter_group resource name to have just general name (previously was `"famdb_postgresql13"`) and it is hard to change in future.
  - used `name_prefix` for parameter_group naming instead of `name` to prevent Terraform error when changing engine family (resource group already exists error).
  - adjusted parameter_group engine family to `family = "aurora-postgresql16"` to match engine_version.

### DIFF
--- a/docker-base-services.yml
+++ b/docker-base-services.yml
@@ -6,50 +6,50 @@
 #
 ---
 services:
-  fam-database:
-    image: postgres:14.1-alpine
-    mem_limit: 128m
-    environment:
-      - POSTGRES_USER=postgres
-      - POSTGRES_PASSWORD=postgres
-      - POSTGRES_DB=fam
-      - POSTGRES_HOST_AUTH_METHOD=password
-    healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres"]
-      interval: 5s
-      timeout: 5s
-      retries: 6
-    networks:
-      - fam
+    fam-database:
+        image: postgres:17.4-alpine
+        mem_limit: 128m
+        environment:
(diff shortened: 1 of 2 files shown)


docker-base-services.yml
```diff
@@ -6,50 +6,50 @@
...
-      - POSTGRES_HOST_AUTH_METHOD=password
-    healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U postgres"]
-      interval: 5s
-      timeout: 5s
-      retries: 6
-    networks:
-      - fam
+    fam-database:
+        image: postgres:17.4-alpine
```
basilv, line 10: Why is this 17 when AWS is set to 16 below?

ianliuwk1019: I tried 17.4 on docker first and it worked locally but until I found out 17.4 is not available for AWS aurora-postgresql yet when deployment failed (version not found error). I leave it there in docker thought it didn't hurt to have local a bit higher version. Maybe you might prefer they are the same version?

basilv: I'd keep it the same version.

basilv (approved):

docker-base-services.yml
ianliuwk1019: It is the same now.

Issue #1784, opened 2025-03-19
Results Exam CHEFS request


craigyu:
Confirmation ID: 47FA4F50

- Separate results exam from Silva in all environment
- Complete the results exam request

PR #1787, opened 2025-03-24
feat: #1784 results exam chefs request

craigyu:
### Description
- Added Results exam app with no roles
- Separated results exam from silva


server/flyway/sql/V62__add_results_exam.sql
```diff
@@ -0,0 +1,95 @@
...
+        create_user,
+        create_date
+    )
+SELECT
+    c.client_id,
+    a.application_id,
+    CURRENT_USER,
+    CURRENT_DATE
+FROM
+    temp_combined_apps c
```
basilv, line 94: I just noticed this doesn't drop the temp tables at the end of the script, and I think it should.

basilv (changes requested):

server/flyway/sql/V62__add_results_exam.sql
craigyu: I think they are dropped at the start of the script with `DROP TABLE IF EXISTS temp_TABLE_NAME;`.
Since temp tables in Postgre are automatically dropped at the end of the session, they won’t persist after the migration completes. So we don’t need to manually drop them at the end of each migration file.

basilv (approved):

PR #1788, opened 2025-03-26
fix: apps not showing for fam admins

craigyu:
Allows FAM admins to grant app admin access for any application.


basilv (commented):
Approved

Issue #1763, opened 2025-03-04
Fix Cognito APP Clients Attributes Permission


ianliuwk1019:
**Describe the bug**
During the investigation of the issue ticket #1754 (PROD Cognito user pool business_guid missing from some users) we found out users may need to sign in through different Cognito APP Clients for different FAM's downstream applications. The APP Clients configuration for attributes' permission mapping are different (from the past for some purpose not known why) and this difference causes the attributes for some users in Cognito User Pool in some user cases showing as missing the attributes (e.g., `business_guid`, `display_name`). 
This also causes previous AUTH Lambda updates (overrides) the `business_guid` to be null in FAM's user table.

We have concluded, the first step is to adjust this configuration for different Cognito APP Clients for "necessary" attributes permission (especially from FAM point of view) to be consistent to prevent user attributes being overridden into FAM user table.  The Terraform code for each `oidc_clients_[app's name]` needs to be adjusted for the `read_attributes` and `write_attributes`.

**Acceptance Criteria**
- [x] Verify and list necessary attributes to be adjusted for Terraform `oidc_client` files for the  `read_attributes` and `write_attributes`. Solution may be to adjust the `minimum_oidc_attribute_list` for making consistency easier for all app clients (FAM included). 
- [ ] Deploy and verify for users previous missing the attributes (e.g., business_guid) in Cognito User Pool showing up the attributes again.

**Additional context**
- ticket #1754 
- Analysis: https://apps.nrs.gov.bc.ca/int/confluence/display/FSAST1/Missing+Business_GUID+for+Some+Business+BCeID+Users .

ianliuwk1019:
Current listed available attributes read/write permission for Cognito App Clients:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/po-yi.liu/AppData/Local/Temp/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/po-yi.liu/AppData/Local/Temp/msohtmlclip1/01/clip_filelist.xml">
</head>

<body link="#467886" vlink="#96607D">


[...]

PR #1792, opened 2025-04-03
fix: #1763 Adjust Cognito app clients read/write attributes

ianliuwk1019:
closes: #1763 

Adjust Cognito App Clients read/write attributes to be consistent to avoid User Pools users necessary attributes missing when log on to different clients.

- updated `minimum_oidc_attribute_list` to include: `custom:idp_business_id`, `email`, `custom:idp_display_name`, `given_name`, `family_name`.
- adjusted all App Clients on Terraform file to reflect the change above.

Note: the `forest_client` App Client was having maximum attributes initially (unclear for the reason why it is very different than others) so no change for this app client.

### DIFF
--- a/infrastructure/server/oidc_clients_fam.tf
+++ b/infrastructure/server/oidc_clients_fam.tf
@@ -1,3 +1,8 @@
+/*
+User OIDC attributes, FAM includes:
+  - minimum_oidc_attribute_list: required for all OIDC clients
+  - custom:idp_business_name: FAM uses this attribute to display user's organization name.
+*/
 resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
   access_token_validity                = "5"
   allowed_oauth_flows                  = ["code"]
--- a/infrastructure/server/oidc_clients_fam.tf
+++ b/infrastructure/server/oidc_clients_fam.tf
@@ -20,7 +20,7 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
   id_token_validity                             = "60"
   name                                          = "fam_console"
   prevent_user_existence_errors                 = "ENABLED"
-  read_attributes                               = concat(var.minimum_oidc_attribute_list, ["custom:idp_display_name", "email", "custom:idp_business_id", "custom:idp_business_name"])
+  read_attributes                               = concat(var.minimum_oidc_attribute_list, ["custom:idp_business_name"])
   refresh_token_validity                        = "24"
   supported_identity_providers                  = [var.fam_console_idp_name, var.fam_console_idp_name_bceid]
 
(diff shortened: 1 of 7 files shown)


infrastructure/server/oidc_clients_fam.tf
```diff
@@ -31,7 +31,7 @@ resource "aws_cognito_user_pool_client" "fam_console_oidc_client" {
   }
 
   user_pool_id     = aws_cognito_user_pool.fam_user_pool.id
-  write_attributes = concat(var.minimum_oidc_attribute_list, ["custom:idp_display_name", "email", "custom:idp_business_id", "custom:idp_business_name"])
+  write_attributes = concat(var.minimum_oidc_attribute_list, ["custom:idp_business_name"])
```
basilv, line 34: Why are we keeping FAM different? Should have a comment explaining why.

basilv:
We should have a comment explaining why client is different? My guess is it would be linked to their use of BC Services Card, but can we confirm that? What extra attributes does Client get?

ianliuwk1019:
> We should have a comment explaining why client is different? My guess is it would be linked to their use of BC Services Card, but can we confirm that? What extra attributes does Client get?

Yes. I think I will create a ticket separately to investigate further with Client team for what attributes they really need/use and adjust for it later. Interestingly to me when I worked on this ticket and see Client was configured with maximum attributes available but I have no previous knowledge for this; the Git history shows it was dated back to Conrad who configured this for Client back on year 2023 and no one touches it since then.

infrastructure/server/oidc_clients_fam.tf
ianliuwk1019: There is no additional attribute added; they are still the same attributes as before for FAM.
I added a comment at the beginning of the [this TF file](https://github.com/bcgov/nr-forests-access-management/pull/1792/files#diff-0234a9835dc96b38911131d11fc726158f9e043243821d6766060a0fd8c9ec72).

basilv: My question is why does FAM require idp_business_name and will this work given what we know know about other apps not having this attribute listed means the user isn't guaranteed to have this stored in the FAM_USER table...

In other words, if our goal in the PR is to handle attributes consistently, are we clear on why FAM is different and confident that it will work.

basilv: Sorry, just saw the comment on why FAM needs idp_business_name, so the question remaining is whether this will work...

basilv: Okay, I've assessed. FAM uses idp_business_name only in the front-end, extracted from the identity token, and the FAM_USER table doesn't even have a column to store idp_business_name. Therefore, it is safe to have only FAM add this extra attribute, with the understanding it isn't stored in the FAM_USER table. All attributes of the user stored in FAM_USER table must be part of the minimum attribute set used by all applications.

basilv:
> > We should have a comment explaining why client is different? My guess is it would be linked to their use of BC Services Card, but can we confirm that? What extra attributes does Client get?
> 
> Yes. I think I will create a [ticket #1796](https://github.com/bcgov/nr-forests-access-management/issues/1796) separately to investigate further with Client team for what attributes they really need/use and adjust for it later. Interestingly to me when I worked on this ticket and see Client was configured with maximum attributes available but I have no previous knowledge for this; the Git history shows it was dated back to Conrad who configured this for Client back on year 2023 and no one touches it since then.

I agree with this approach.

basilv (approved):

Issue #1764, opened 2025-03-04
Enable AUTH Lambda Trigger to Update User Record again


ianliuwk1019:
**Describe the task**
This ticket is the continuation from the solution ticket #1763 to resolve issue discovered in ticket #1754 (business_guid missing from Cognito User Pool for some users) so #1763 needs to be done first.

After #1763 is deployed and verified users in Cognito User Pool now having correct attributes showing up and consistent when login through different APP Clients then we can enable again for AUTH Lambda (pre-token generation Cognito lambda trigger) to update FAM user table's `business_guid`, `user_name`, `email`. This was previous disabled due to emergency fix for issue described in ticket #1754.

**Acceptance Criteria**
- [x] Adjust AUTH lambda function at backend to enable the attributes update for the above list.
- [x] Make sure the log is useful for troubleshooting purpose.
- [x] Adjust tests if necessary.
- [x] Deploy and verify attributes are being updated into FAM user records through user login and verify the log on Cloud Watch showing correctly and useful. 

**Additional context**
- Fix Cognito APP Clients Attributes Permission #1763

ianliuwk1019:
Note: for this ticket, although "given_name" and "famil_name" are also available, but I didn't include them as they are still questionable for that I am currently not seeing both attributes having values for all users (except user with BCSC log in).

PR #1806, opened 2025-04-15
fix: #1764 enable auth lambda trigger to update user record again

ianliuwk1019:
ref: #1764 
- lambda_function (auth):
	* adjusted logic for user (fam_user table) insert/update for additional update on 'business_guid' and 'email'
	* added more debug log.
	* added capability to enable debug logging with environment switch.
	* migrated audit logging to its own place with decorator and not mixing with main code.

- fixed/refactored auth tests.
- add 30s wait time before backend smoke test to avoid the check failure after deployment.


server/auth_function/lambda_function.py
```diff
@@ -183,19 +200,20 @@ def populate_user_if_necessary(db_connection, event) -> None:
...
     raw_query = """INSERT INTO app_fam.fam_user
-        (user_type_code, user_guid, cognito_user_id, user_name, business_guid,
+        (user_type_code, user_guid, cognito_user_id, user_name, business_guid, email,
         create_user, create_date, update_user, update_date)
-        VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid},
+        VALUES( {user_type_code}, {user_guid}, {cognito_user_id}, {user_name}, {business_guid}, {email},
         CURRENT_USER, CURRENT_DATE, CURRENT_USER, CURRENT_DATE)
         ON CONFLICT (user_type_code, user_guid) DO
-        UPDATE SET user_name = {user_name},  cognito_user_id = {cognito_user_id}, update_user = CURRENT_USER, update_date = CURRENT_DATE;"""
+        UPDATE SET user_name = {user_name},  cognito_user_id = {cognito_user_id}, business_guid = {business_guid}, email = {email}, update_user = CURRENT_USER, update_date = CURRENT_DATE;"""
```
ianliuwk1019, line 208: "business_guid" is brought back for update query.
"email" is added for insert/update as this attribute is available to user log in with all Cognito app client (but could be empty).

**Note**: for this ticket, although "given_name" and "famil_name" are also available, but I didn't include them as they are still questionable for that I am currently not seeing both attributes having values for all users (except user with BCSC log in).

basilv (approved):

Issue #1801, opened 2025-04-10
Permissions History page for granting application admin history


ianliuwk1019:
FAM currently has permission history for granting regular user and granting delegated admin but it does not have history for granting application admin.

Might need to find out some details for below questions before refining ticket:
- Is this feature needed?
- Does current data model support for implementing permission history?
- Can current frontend history page/table structure and display be reused?
- Who should allowed to view this page? 

![Image](https://github.com/user-attachments/assets/ad41a0c0-8c0f-4b3d-9657-530ba8eb55b3)

**Additional Context**

PR #1816, opened 2025-04-25
feat: #1801 Permissions history page for granting application admin

ianliuwk1019:
ref: #1801 
- Adjust frontend ManagePermissionTable.vue to include navigating to
- Adjust router_application_admin.py application_admin_repository.py for implementing application admin history.
- Adjust application_admin_service to save audit record for granting or revoking application admin.
- Add methods to permission_audit_service.py to save application admin granting/revoking audit history.
- Add tests


server/admin_management/api/app/routers/router_application_admin.py
```diff
@@ -103,7 +103,7 @@ def create_application_admin(
         )
 
         response = application_admin_service.create_application_admin(
-            application_admin_request, target_user, requester.cognito_user_id
```
basilv, line 106: This looks wrong - create_application_admin above was changed to accept cognito_user_id as an argument?

basilv: Or this is fine because the service was changed as well.

server/admin_management/api/app/services/permission_audit_service.py
```diff
@@ -60,6 +61,31 @@ def store_delegated_admin_permissions_granted_audit_history(
...
+        requester: Requester,
+        change_target_user: FamUser,
+        fam_application_admin_user: FamApplicationAdmin
+    ):
+        change_type = PrivilegeChangeTypeEnum.GRANT
+        privilege_details = PrivilegeDetailsSchema(
+            permission_type=PrivilegeDetailsPermissionTypeEnum.APPLICATION_ADMIN,
+        )
+        audit_record = PermissionAuditHistoryCreateSchema(
+            application_id=fam_application_admin_user.application.application_id,
```
basilv, line 75: This line to obtain the application id looks weird.

basilv: Why does the admin user's application matter?

server/admin_management/api/app/routers/router_application_admin.py
ianliuwk1019: Yes, this line was only a refactored code to accept the 'requester' object change from the service method.

server/admin_management/api/app/services/permission_audit_service.py
ianliuwk1019: When application admin is granted we need to store the audit record for this.
To save the audit record, one attribute to be saved is the application_id, and this can be obtained from the currently granted `fam_application_admin_user` (this is mapped to the `fam_application_admin` table entity) from it's `application` relationship to grab the id.

basilv: Ah, I think I was confused between the admin user performing the action and the admin access being granted, which is really a permission being granted and not a user, so I was confused by it being called fam_application_admin_user.

ianliuwk1019: I see, the variable name confuses your thought.
I like to refactor the variable name (not only on this file context, some other files too related to this name), what name would you give or prefer so that you don't get confused?
The table entity class is called `FamApplicationAdmin` and the variable is associated with that type, perhaps removing `_user` at the end (`fam_application_admin`), will it be less confusing to you (or maybe, `fam_application_admin_grant`?

basilv: Yes, removing _user suffix would likely be clearer.

ianliuwk1019: Variable is renamed to `fam_application_admin` without _user suffix. Thanks Basil!

basilv (approved):

Issue #1821, opened 2025-05-02
FOM Callback URL Change


ianliuwk1019:
FOM Test environment no longer use vanity URL and corresponding paid certificate.
This is done at pr: https://github.com/bcgov/nr-fom/pull/780.

The callback URL need to be adjusted at Cognito for FOM TEST.
New setting: "`https://fom-test.apps.silver.devops.gov.bc.ca/admin/search`"

**Acceptance Criteria**
- [x] Adjust Terraform Cognito callback url for FOM TEST
- [ ] Deploy and FOM TEST still works.

**Additional context**
- Add any other context about the task here.
- Or here

PR #1822, opened 2025-05-02
fix: 1821 fom callback url change

ianliuwk1019:
ref #1821 
- Adjust FOM TEST callback url due to change of using default OpenShift route.

### DIFF
--- a/frontend/src/aws-exports.js
+++ b/frontend/src/aws-exports.js
@@ -5,23 +5,23 @@
  *
  * A note for purpose of using optional ("env?.") is that this is not to skip values when "env"
  * is not loaded. When there is no "env.json" loaded yet, the error message is strange; so give it
- * optional "?" simply let the down stram Amplify to throw more meaningful error.
+ * optional "?" simply let the Amplify to throw more meaningful error.
  */
-const env = JSON.parse(window.localStorage.getItem('env_data'));
+const env = JSON.parse(window.localStorage.getItem("env_data"));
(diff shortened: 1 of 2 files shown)


frontend/src/aws-exports.js
```diff
@@ -5,23 +5,23 @@
  *
  * A note for purpose of using optional ("env?.") is that this is not to skip values when "env"
  * is not loaded. When there is no "env.json" loaded yet, the error message is strange; so give it
- * optional "?" simply let the down stram Amplify to throw more meaningful error.
+ * optional "?" simply let the Amplify to throw more meaningful error.
```
ianliuwk1019, line 8: Nothing change in this file, only remove a typo.

infrastructure/server/oidc_clients_fom.tf
```diff
@@ -45,8 +45,7 @@ resource "aws_cognito_user_pool_client" "test_fom_oidc_client" {
   allowed_oauth_scopes                          = ["openid", "profile", "email"]
   callback_urls                                 = [
     var.oidc_sso_playground_url,
-    "https://fom-test.nrs.gov.bc.ca/admin/search",
-    "https://fom-demo.apps.silver.devops.gov.bc.ca/admin/search",
```
ianliuwk1019, line 49: Also remove the 'demo' since no longer needed from business.

basilv (approved):

PR #1825, opened 2025-05-05
fix: Add service unavailable message.

ianliuwk1019:
- Add service unavailable message at FAM landing page. Temporary message for SSO login instability.


ianliuwk1019:
The backend error is due to Forest Client API timeout. I think we can ignore it as FC API might experience similar service interruption from gov.

basilv (approved):

PR #1836, opened 2025-06-05
chore: update nr-spar TEST and PR config

DerekRoberts:
Updates for nr-spar's shortened frontend URLs.


ianliuwk1019:
Hi @DerekRoberts , I am not sure but I think in the past the 'chore' will not add to the release-please action flow to have a version change. I maybe wrong.

ianliuwk1019:
> @ianliuwk1019 Chore works with Conventional Commits. Release-Please is a history muddying abomination!

I understand "chore" is from conventional commit, but just wondering in the past, somehow "chore" will not trigger the version increase, might be in some setting (I didn't look into that). Will see if it will trigger it.

Issue #1839, opened 2025-06-12
FAM security vulnerabilities


OlgaLiber2:
**Describe the task**
 Address security vulnerabilities

**Acceptance Criteria**
- [x] Dependabot: https://github.com/bcgov/nr-forests-access-management/security/dependabot
- [ ] Codescanning: https://github.com/bcgov/nr-forests-access-management/security/code-scanning, please refer to #1838

MCatherine1994:
There are three Dependabot warnings left:
<img width="1304" alt="Image" src="https://github.com/user-attachments/assets/f55f55f9-5cce-4465-9aa2-3bbb88ebbe24" />

The first warning https://github.com/bcgov/nr-forests-access-management/security/dependabot/214 is caused by the following reason, full comment please see #1843:
<img width="817" alt="Image" src="https://github.com/user-attachments/assets/ca5fa594-ef05-4836-972d-1f4f3f3fbf9d" />

The last two warnings related to the package `cryptography` which is used by the BCSC login. We have a pr opened for them #1845. But since we don't have a reliable local testing approach yet, we'll postpone that until after the initial production release and handle it in a separate release. Same as #1853.

PR #1843, opened 2025-06-18
fix(deps): #1839 Update the openapitool package in code gen

MCatherine1994:
for fixing: https://github.com/bcgov/nr-forests-access-management/security/dependabot and https://github.com/bcgov/nr-forests-access-management/security/dependabot/205 and https://github.com/bcgov/nr-forests-access-management/security/dependabot/214


ianliuwk1019 (approved):
Thanks for the update Catherine!

PR #1844, opened 2025-06-18
fix(deps): #1839 Update vite related packages for frontend

MCatherine1994:
fix dependabot warnings: https://github.com/bcgov/nr-forests-access-management/security/dependabot/206, https://github.com/bcgov/nr-forests-access-management/security/dependabot/204, https://github.com/bcgov/nr-forests-access-management/security/dependabot/203, https://github.com/bcgov/nr-forests-access-management/security/dependabot/200, https://github.com/bcgov/nr-forests-access-management/security/dependabot/194, https://github.com/bcgov/nr-forests-access-management/security/dependabot/185, https://github.com/bcgov/nr-forests-access-management/security/dependabot/184

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -13,7 +13,7 @@
                 "@carbon/themes": "^11.43.0",
                 "@carbon/type": "^11.31.0",
                 "@tanstack/vue-query": "^5.56.2",
-                "aws-amplify": "^5.3.27",
+                "aws-amplify": "^6.15.0",
                 "axios": "1.8.2",
                 "bootstrap": "^5.3.3",
                 "fam-admin-mgmt-api": "file:../client-code-gen/gen/admin-management-api",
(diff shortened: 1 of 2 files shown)


frontend/package-lock.json
```diff
@@ -13,7 +13,7 @@
                 "@carbon/themes": "^11.43.0",
                 "@carbon/type": "^11.31.0",
                 "@tanstack/vue-query": "^5.56.2",
-                "aws-amplify": "^5.3.27",
+                "aws-amplify": "^6.15.0",
```
MCatherine1994, line 16: I'm not sure why there is still aws-amplify changes in the package-lock.json, I thought it should be captured in the last pr. But I tried, everything works fine.

ianliuwk1019 (approved):
I am not sure if we still need 'vitest' and can be removed. We can leave it next time when we visit again.

MCatherine1994:
> I am not sure if we still need 'vitest' and can be removed. We can leave it next time when we visit again.

The sonar pipeline is trying to run the test,  and it's using "vitest". And because we don't have any test file, it always fails.
Screenshot 2025-06-18 at 3 00 09 PM

PR #1845, opened 2025-06-18
fix(deps): #1839 Upgrade cryptography version for api

MCatherine1994:
fix dependabot https://github.com/bcgov/nr-forests-access-management/security/dependabot/187, https://github.com/bcgov/nr-forests-access-management/security/dependabot/186

### DIFF
--- a/server/backend/requirements.txt
+++ b/server/backend/requirements.txt
@@ -11,5 +11,5 @@ requests==2.33.0
 psycopg2-binary==2.9.11
 mangum==0.19.0
 cryptography==46.0.6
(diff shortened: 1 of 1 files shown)


ianliuwk1019 (approved):
Hi Catherine, I approve this but the upgrade as mentioned is within bcsc logic.
I feel we can hold this one in pr and not merging to main with others.
After all other FAM changes are released, this can go separately on its own to have its release. If indeed it introduces some issue in production after bcsc testing, we can quickly roll it back to previous release.

Issue #1846, opened 2025-06-19
FAM deployment issue caused by terraform version change


MCatherine1994:
**Describe the task**
FAM deployment failed because the latest terraform version doesn't support stage_name argument in the aws_api_gateway_deployment. 

We don't define the terraform version in our pipeline, so every time it will use the latest version. The deployment works fine on [`hashicorp/aws v5.100.0`](https://github.com/bcgov/nr-forests-access-management/actions/runs/15720169607/job/44299300242), but failed on [`hashicorp/aws v6.0.0`].

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/0ba876d8-3089-4691-a60e-6f35dc9e38b5" />
<img width="1175" alt="Image" src="https://github.com/user-attachments/assets/cd8eab9b-212b-4a67-ac6b-ca0354eea4ce" />

**Acceptance Criteria**
- [ ] Check which version we want to stay for aws terraform 
- [ ] Update in the terraform config file
- [ ] (Optional) Update the code to support the terraform version if needs

**Additional context**
- Add any other context about the task here.
- Or here

PR #1852, opened 2025-06-19
fix: #1846 added terraform required provider version

MCatherine1994:
I picked v5.99.1 for now as this is the version that we used for the [last successful deployment in PROD](https://github.com/bcgov/nr-forests-access-management/actions/runs/15477253127/job/43575553582).

Tested the deployment in Tools namespace, works fine https://github.com/bcgov/nr-forests-access-management/actions/runs/15766641334

### DIFF
--- a/infrastructure/frontend/version.tf
+++ b/infrastructure/frontend/version.tf
@@ -1,0 +1,8 @@
+terraform {
+  required_providers {
+    aws = {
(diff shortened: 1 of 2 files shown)


ianliuwk1019 (approved):
Awesome, thanks Catherine!

Issue #1933, opened 2025-08-21
Setup Github action for LZA dev and deploy services


MCatherine1994:
**Describe the task**
Setup Github Action for LZA dev space and deploy all the services

**Acceptance Criteria**
- [ ] Config the github action token
- [ ] Setup the deployment pipeline for dev environment
- [ ] Run the pipeline to deploy all the services

**Additional context**
- Add any other context about the task here.
- Or here

PR #1947, opened 2025-08-29
feat: #1933 LZA migration dev deployment add default redirect path

ianliuwk1019:
* Only to add temporary default Cloudfront domain for Cognito rediret/logout url.


MCatherine1994 (approved):
Thanks Ian! This pr is still in draft though.

Issue #2024, opened 2025-11-14
[FAM UserRole Expiry] API and Service adjustment for New UserRole Expiry Date


ianliuwk1019:
**Describe the task**
For the new feature of allowing admin to assign user-role with expiry date, the current FAM backend need to be adjusted to support the use case.

Basic behaviors are outlined below:

1. DB - New Column Addition
   * Add **expiry_date** column to fam_user_role_xref table.
   * Data type: TIMESTAMP WITH TIME ZONE.
   * Column should allow NULL (meaning no expiry).

2. Existing Records
   * For MVP: Existing user-role assignments will have **expiry_date = NULL** (no expiry).
   * No default expiry date will be applied for existing or new records (unless date is picked by admins).

3. Deletion Behavior
   * Current behavior (**physical delete**) remains unchanged.
   * Do not set expiry_date = today on deletion, just delete the record, same as current behavior.
   * No soft-delete mechanism introduced for MVP.

4. Audit Records
   * No new audit record for automatic expiry.
   * Newly grant audit record should contain expiry information.
   * If expiry date is changed manually by an admin, then a new audit record is required.

5. UI Behavior
   * When a role expires (based on expiry_date), UI should display “expired” status automatically.
   * No admin action required for expired roles.

6. Authorization roles on JWT token
   * Auth function should omit roles that are expired.

**Acceptance Criteria**
- [ ] Adjust `/user-role-assignment` endpoint to receive new request body param `expiry_date` date with updated pydantic schema.
- [ ] Make sure the `expiry_date` is valid for its format and date range for business validation.
- [ ] Adjust `crud_user_role::def create_user_role_assignment_many` method to set the expiry date when new user-role assignment is created.
- [ ] Adjust `PrivilegeDetailsRoleSchema` to store expiry date for the new user-role assigned to the audit record.
- [ ] Generate new frontend client swagger code
- [ ] Adjust `/get_fam_application_user_role_assignment` endpoint response to return exipry-date in each user-role assignment in result.

**Additional context**
- Dependency tickets: #2023 , #2022 
- Or here

ianliuwk1019:
For **Permission Audit History** saved in db when user role permission is granted or revoked, I am tentatively adding a field "**role_assignment_expiry_date**" under the "**scope**" in the `privilege_details` JSONB column. Although from the mockup the expiry date is applied to both forest_client_number(s) for the same role.
[...]

Issue #2022, opened 2025-11-13
[FAM UserRole Expiry] UI Mockup for User-Role Expiry feature


ianliuwk1019:
**Description:**

Design mockups for the new user-role expiry functionality, including:
- User role assignment form with expiry date input.
- A status column in user-role list 
- Admin edit/extend expiry date screen (to be considered?).

**Acceptance Criteria:**
- [x] Only 'date' as select input, but indicate time or expiry in BC timezone.
- [x] Select expiry date or ~~default expiry date behavior (e.g., 1 year, max date)~~, AND can be empty (never expired)?
- [x] Clearly indicates what "expiry date" means: the date chosen is the date the role assignment is expired or is the last date the role assignment still available. (research the business meaning or common usage)
- [x] Status column clearly shows expiry state (showing Active/Expired?).
- [x] Figma design linked in the issue.

**Additional context:**
- Mural note - https://app.mural.co/t/climatechange4363/m/climatechange4363/1762361971685/dfd5a53d927a7e94c1ba1a44709f817e9aa19506

ianliuwk1019:
Hello @dianaarvayo , do I have view access to the above Figma flow? Could you grant me (ianliuwk1019) in? Thank you.

dianaarvayo:
@ianliuwk1019 invitation sent, let me know if you can see the Figmas 

<img width="516" height="636" alt="Image" src="https://github.com/user-attachments/assets/5a4a644b-efe1-40c5-9d30-f59893accd0e" />

PR #2033, opened 2025-12-01
feat: #2024 User Role Expiry Backend

ianliuwk1019:
ref: #2024 
The PR implement the User Role Expiry feature at the backend.

Expiry-Date input meaning: (until midnight 23:59:59, decision commented here: https://github.com/bcgov/nr-forests-access-management/issues/2022#issuecomment-3603821614)
Database new column "app_fam.fam_user_role_xref.**expiry_date**" - 
`COMMENT ON COLUMN app_fam.fam_user_role_xref.expiry_date IS`: **The date (in BC timezone) until which the user role assignment is valid. The assignment expires at midnight (00:00:00) following this date. NULL means no expiry.**

- Add optional input expiry_date field and validation to FamUserRoleAssignmentCreateSchema for creation endpoint.
- Add optional expiry_date field to FamApplicationUserRoleAssignmentGetSchema for get user role assignment endpoint.
- Pass expiry_date input to save user role assignment in db.
- Adjust event log to log expiry date for router when creating user role assignment.
- Adjust saving Permission Audit entry to record expiry date.
- Add test to verify expiry date is saved for Permission Audit History.
- Add unit tests for crud/server user role creation with expiry date.
- Add endpoint tests for user role creation with expiry date.
- Frontend client-code-gen produced.

Endpoint checks:
* "expiry_date_date": "" (Empty, no expiry)
   image

* "expiry_date_date": "12-03-2025" (Wrong format)
   image

* "expiry_date_date": "2026-10-32" (No such date/non-existing)
   image

* "expiry_date_date": "2025-10-31" (In the past)
   image

* "expiry_date_date": "2025-12-03" (today)
   Valid until today (2025-12-03, 23:59:59), db saved as timezone-aware UTC
   image
   Audit history:
   image
   Logging:
   image

* "expiry_date_date": "2026-10-31" (future date)
   image
   image
   image

### DIFF
--- a/server/backend/api/app/schemas/fam_user_role_assignment_create.py
+++ b/server/backend/api/app/schemas/fam_user_role_assignment_create.py
@@ -6,8 +6,8 @@ from zoneinfo import ZoneInfo
 
 from api.app.constants import UserType
 from api.app.datetime_format import BC_TIMEZONE, DATE_FORMAT_YYYY_MM_DD
-from pydantic import (BaseModel, ConfigDict, PrivateAttr, StringConstraints,
-                      field_validator, model_validator)
+from pydantic import (BaseModel, ConfigDict, Field, PrivateAttr,
+                      StringConstraints, field_validator, model_validator)
 from typing_extensions import Annotated
 
 LOGGER = logging.getLogger(__name__)
@@ -29,7 +29,10 @@ class FamUserRoleAssignmentCreateSchema(BaseModel):
         List[Annotated[str, StringConstraints(min_length=1, max_length=8)]], None
     ] = None
     requires_send_user_email: bool = False
-    expiry_date_date: Optional[str] = None  # e.g. '2025-12-31', BC timezone.
+    expiry_date_date: Optional[str] = Field( # e.g. '2025-12-31', BC timezone.
+        default=None,
+        description="The expiry date as a string (YYYY-MM-DD), BC timezone. If provided, the role is valid until the end of this day."
+    )
     # Internal use only, not exposed in OpenAPI or request/response bodies
     _expiry_date: Optional[datetime] = PrivateAttr(default=None)
 
--- a/client-code-gen/app-access-control-openapi.json
+++ b/client-code-gen/app-access-control-openapi.json
@@ -1638,7 +1638,8 @@
                 "type": "null"
               }
             ],
-            "title": "Expiry Date Date"
+            "title": "Expiry Date Date",
+            "description": "The expiry date as a string (YYYY-MM-DD), BC timezone. If provided, the role is valid until the end of this day."
           }
         },
         "type": "object",
--- a/client-code-gen/gen/app-access-control-api/.openapi-generator/FILES
+++ b/client-code-gen/gen/app-access-control-api/.openapi-generator/FILES
@@ -1,6 +1,5 @@
 .gitignore
 .npmignore
-.openapi-generator-ignore
 README.md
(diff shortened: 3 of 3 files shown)


server/backend/api/app/crud/crud_user_role.py
```diff
@@ -143,7 +144,7 @@ def create_user_role_assignment_many(
             )
             # Create user/role assignment
             new_user_role_assginment_res = create_user_role_assignment(
-                db, fam_user, child_role, requester.cognito_user_id,
+                db, fam_user, child_role, requester.cognito_user_id, expiry_date=request._expiry_date
```
basilv: Where do we validate what was supplied for expiry date via the API argument?

server/backend/api/app/schemas/fam_user_role_assignment_create.py
```diff
@@ -16,5 +29,39 @@ class FamUserRoleAssignmentCreateSchema(BaseModel):
...
+        if v is None:
+            return v
+        if isinstance(v, str) and v.strip() == "":
+            return None
+        bc_tz = ZoneInfo(BC_TIMEZONE)
+        try:
+            dv = datetime.strptime(v, DATE_FORMAT_YYYY_MM_DD).date()
+        except Exception:
+            raise ValueError('expiry_date_date must be a valid YYYY-MM-DD string')
+        now_bc = datetime.now(bc_tz).date()
```
basilv: I'm not sure I understand why this logic is necessary. If the user enters the expiry date, shouldn't that just be what we save. The conversion to BC timezone only makes sense if we are going to store the date & time, but I'm not sure we are doing that. And if we are storing the date & time, shouldn't we be setting the time component to midnight somewhere in here?

server/backend/api/app/crud/crud_user_role.py
ianliuwk1019: Hi Basil, the validation is on a Pydantic Schema class: server/backend/api/app/schemas/fam_user_role_assignment_create.py. This is used for endpoint body argument intake.

server/backend/api/app/schemas/fam_user_role_assignment_create.py
```diff
@@ -16,5 +29,39 @@ class FamUserRoleAssignmentCreateSchema(BaseModel):
...
+        return v
+
+
+    @model_validator(mode="after")
+    def set_expiry_date(self):
+        if self.expiry_date_date:
+            bc_tz = ZoneInfo(BC_TIMEZONE)
+            d = datetime.strptime(self.expiry_date_date, DATE_FORMAT_YYYY_MM_DD).date()
+            # Set expiry to end of day (23:59:59) in BC timezone
+            self._expiry_date = datetime.combine(d, time(23, 59, 59)).replace(tzinfo=bc_tz)
```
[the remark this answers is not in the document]
ianliuwk1019: Hi Basil,
The `expiry_date_date` (meaning "expiry_date" only the date part) is the actual input (string type) param (like '2025-12-15'). After this input param is validated to be valid, here is just converting to Python's datetime type (with timezone aware in BC) so can save into database correctly. 
I purposely make it another private variable (`_expiry_date` this contains date/time, but just called it `_expiry_date`) so it does not show this field on the Swagger (as not for frontend or endpoint), but internally this is the date/time being saved into database column (TIMESTAMPE WITH TIMEZONE).

Not sure if the variable naming is confusing, if it is, I can rename them.

PR #2109, opened 2026-03-18
feat: updates for apt2 to work in dev, test and prod

gpascucci:


basilv (approved):

lza-infrastructure/terraform/server/oidc_clients_apt2.tf
```diff
@@ -98,15 +81,13 @@ resource "aws_cognito_user_pool_client" "prod_apt_oidc_client" {
   allowed_oauth_scopes                          = ["openid", "profile", "email"]
   callback_urls                                 = [
     var.oidc_sso_playground_url,
-    "https://apps.nrs.gov.bc.ca/int/apt2/dashboard",
-    "https://apps.nrs.gov.bc.ca/int/apt2/callback",
```
jazzgrewal: @gpascucci removing the /callback urls will break the legacy APT which is currently deployed in PROD, are you fine with that? 
Or you would like to keep the /callback, so that was the legacy APT stays up and running till you deploy your version to the PROD. And once that it done, the /callback urls can be removed, as the legacy app is no longer needed.

Issue #2063, opened 2026-01-20
[FAM] Handle FAM When Forest Client API Service Unavailable


ianliuwk1019:
**Describe the task**

FAM interfaces with few external services. One of the external service is **Forest Client API** and is the service that most commonly goes down at its TEST environment. Users who login into FAM (all environment, including prod, since prod also hosts non-prod downstream apps and roles) will get timeout on the dashboard and can't navigate to anywhere. This is the most common timeout FAM encounters in the past.

Won't do -
~~It would be a good idea when this timeout happens on Forest Client API,  FAM's AWS Dashboard can send have alarm and can send notification to the team mailbox.~~

After team discussion -
* When **Forest Client API** service is down, upon user login, when FAM receives Forest Client API service is unavailable, then ignore hitting the service for getting 'forest client name' and other information for displaying on dashboard or in table.

* Showing users warning message, wording such as "**External service is unavailable, you will not be able to grant user access. Please come back again later.**".

**Acceptance Criteria**
- [x] Code change to ignore hitting **"Forest Client API"**, the information previously needed from **"Forest Client API"** will be just empty.
- [x] The endpoint `/admin-user-accesses` call during login will have empty `client_name` for the abstract roles.
       <img width="916" height="482" alt="Image" src="https://github.com/user-attachments/assets/eeffe5d2-6c55-4bc7-ae96-08a3aa1abd5a" />
- [x] The endpoint `/fam-applications/2/user-role-assignment` call when selecting an application to show role-assignment results will have empty `client_name` for the abstract roles.
       <img width="468" height="181" alt="Image" src="https://github.com/user-attachments/assets/ad124089-adab-4bfd-9423-c2113aeb58d8" />
- [x] The endpoint `/access-control-privileges?application_id=2` call when selecting an application to show delegated admin results will have empty `client_name` for the abstract roles.
- ~[ ] "Add Permission" will be disabled due to **"Forest Client API"** is unavailable.~
- ~[ ] Add a user friendly warning message based on UI design.~

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
@basilv I don't see **Forest Client API** itself has a smoke test endpoint to ping for health check.
FAM does have Forest Client search endpoint itself `/forest-clients/search` (internally, integrated to Forest Client API). I guess FAM frontend will need to periodically retry this endpoint to determine if the service is down so the frontend can display warning (service unavailable) message to the users, is that a common approach?

[...]

PR #2161, opened 2026-04-27
fix: #2063 Handle Forest Client API Service Unavailable

ianliuwk1019:
ref: #2063

### Summary

Skips Client Information update and adds error handling for when the Forest Client (FC) external API is unavailable so frontend can render propoer message for users.
> - User-initiated FC add action → error propagates as HTTP 504 → frontend renders
>   `ServiceUnavailableNtfnTemplate` with a clear message
> - Background FC name lookup → soft-fail after retry; returns empty results, existing data is unaffected

---

### Backend — Both `server/backend` and `server/admin_management`

- Added `retry_on_timeout` param and logic to `search()` method and refactored `ForestClientIntegrationService`
- Added `requests_gateway_timeout_error_handler` for `requests.exceptions.Timeout` and `ConnectionError` to returns
  proper response: **HTTP 504 Gateway Timeout** with structured body.
- At forest_client_dec decorator added `__search_forest_clients_with_retry()` wrapping FC search with `retry_on_timeout=True`
- **Soft-fail**: forest_client_dec decorators, catches the error and returns `[]`
  — client names are simply omitted rather than throwing error.
- Added unit tests.

---

### Frontend

**`ForestClientAddTable.vue`**
- Added `isForestClientServiceDown` reactive state
- On mutation error: uses `getAxiosErrorStatus()` to detect HTTP 504; shows service-down notification
  instead of generic error message
- Resets `isForestClientServiceDown` when the input field is cleared

**New `ServiceUnavailableNtfnTemplate.vue`**
- Reusable notification component for upstream service unavailability
- Styled with red error border + Carbon `WarnIcon`; accepts optional `message` prop
- Default: `"External service is unavailable. Please try again later."`


server/admin_management/api/app/main.py
```diff
@@ -66,6 +68,8 @@ def custom_generate_unique_id(route: APIRouter):
     allow_headers=["*"],
     allow_credentials=True,
 )
+app.add_exception_handler(Timeout, requests_gateway_timeout_error_handler)
+app.add_exception_handler(ConnectionError, requests_gateway_timeout_error_handler)
```
ianliuwk1019, line 72: Man.py registers connection/timeout errors handler.

server/admin_management/api/app/decorators/forest_client_dec.py
```diff
@@ -134,4 +145,23 @@ def __update_get_delegated_admin_assignment_fc_return(result_list: List[FamAcces
...
-    return result_list
\ No newline at end of file
+    return result_list
+
+
+def __search_forest_clients_with_retry(
+    forest_client_integration_service: ForestClientIntegrationService,
+    fc_search_params: ForestClientIntegrationSearchParmsSchema,
+):
+    """Search Forest Client API with retry and soft-fail handling."""
```
basilv: Maybe I'm missing something, but I don't see the retry logic. I guess its the integration service that is performing the retry...

server/admin_management/api/app/integration/forest_client_integration.py
```diff
@@ -28,7 +29,8 @@ class ForestClientIntegrationService():
     """
     # https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts
     # https://docs.python-requests.org/en/latest/user/advanced/#timeouts
-    TIMEOUT = (5, 10) # Timeout (connect, read) in seconds.
+    TIMEOUT = (5, 10)  # Timeout (connect, read) in seconds.
+    RETRY_DELAY_SECONDS = 2
```
basilv, line 38: I kind of feel like the max attempts constant should be defined in the same place as this constant. We want to ensure the maximum time delay after all retries doesn't exceed our lambda/API gateway max call duration. Probably worth having a comment to that effect next to these constants. 

I think 2 sec retry delay and 2 max attempts is reasonable.

server/admin_management/api/app/integration/forest_client_integration.py
```diff
@@ -66,36 +76,89 @@ def search(self, search_params: ForestClientIntegrationSearchParmsSchema):
...
-            if ((status_code == HTTPStatus.NOT_FOUND) or (status_code == HTTPStatus.BAD_REQUEST)):
-                return [] # return empty for FAM forest client search
-
-            # Else raise error, including 500
-            # There is a general error handler, see: requests_http_error_handler
-            raise he
+        return self.__do_request(url=url, retry_on_timeout=retry_on_timeout)
+
+    def __do_request(self, url, params=None, retry_on_timeout: bool = False):
+        max_attempts = 2 if retry_on_timeout else 1
```
basilv: So as per earlier comment, have a constant for "2".

basilv (changes requested):

server/admin_management/api/app/decorators/forest_client_dec.py
ianliuwk1019: Yes, it is in the integration service. I updated the comment so it is less confusing now.

server/admin_management/api/app/integration/forest_client_integration.py
ianliuwk1019: Good catch. I define a constant here with some comment added as you suggested. Please see if it looks clear. Thanks Basil.

server/admin_management/api/app/integration/forest_client_integration.py
ianliuwk1019: Used constant now.

basilv (approved):

PR #2197, opened 2026-05-27
feat: onboard lexis to fam

alexmcdermid:


server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
```diff
@@ -0,0 +1,78 @@
+-- Replace placeholder LEXIS industry roles with submitter roles.
+-- Keep ministry roles as concrete (C).
+-- Provincial submitter is abstract (A) for forest-client scoped assignments.
+-- Federal submitter is concrete (C) for a single named user.
```
ianliuwk1019: Just curious, so there are two types of submitters with BCeID for the users? (Provincial submitter and Federal submitter)
And one is 'Concrete' and the other is 'Abstract'? That looks strange...

server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
```diff
@@ -0,0 +1,78 @@
...
+AND role_name IN (
+    'ADMIN',
+    'READ_ONLY',
+    'APPLICATION_APPROVER',
+    'EXEMPTION_APPROVER',
+    'PROVINCIAL_SUBMITTER',
+    'FEDERAL_SUBMITTER'
+)
+;
+
```
ianliuwk1019, line 70: Hi Alex, please have all the role names prefix with "LEXIS_" to be consistent with other apps. Thank you!

server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
```diff
@@ -0,0 +1,78 @@
...
+-- Enable API access for all active LEXIS roles used by the current security model.
+UPDATE app_fam.fam_role
+SET call_api_flag = true
+WHERE application_id in (select application_id from app_fam.fam_application where application_name IN ('LEXIS_DEV', 'LEXIS_TEST', 'LEXIS_PROD'))
+  AND role_name IN ('ADMIN', 'READ_ONLY', 'APPLICATION_APPROVER', 'EXEMPTION_APPROVER', 'PROVINCIAL_SUBMITTER', 'FEDERAL_SUBMITTER')
+;
```
ianliuwk1019, line 76: Are you sure all roles are having `call_api_flag = true`? Normally, it might be only for 'admin' roles (but depending on what you find out the **need from app**).

ianliuwk1019 (changes requested):
Hi Alex,
A bit hard to read the sql using CASE, but as long as it works :)
Leaving some comments if you could review again. Thank you!

server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
alexmcdermid: We spoke with Judith earlier and were informed that there is a single federal user and it's just to transmit xml data from the federal system into lexis. For the time being it can be concrete, it may be removed moving forward depending

server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
alexmcdermid: Yes, apologies, good catch!

server/flyway/sql/V87__update_lexis_industry_roles_to_abstract.sql
alexmcdermid: Good call I'll just go with admin for now, thank you

ianliuwk1019 (approved):
Looks good :)

Issue #2052, opened 2025-12-16
APT3 Application and Cognito client cleanup.


ianliuwk1019:
**Describe the task**
Previously APT3 was created for onboarding future APT modernized work to FAM.
Later the team that works on the APT modernization decided to reuse existing APT2, so this newly created APT3 is no longer used.
We will need to delete the application and app client to prevent confusion.

**Acceptance Criteria**
- [ ] Verify FAM dev/test/prod for APT3 if there is any data associated with this APT3 and delete them (no need since team never uses this app client and the data should only be for testing purpose).
- [ ] Flyway script to remove APT3 application, roles and application client associated with this app. 
- [ ] Terraform change to remove this application client from Cognito.
- [ ] Test on Tools space to see everything is working.

**Additional context**
- Add any other context about the task here.
- Or here

ianliuwk1019:
Is it ok in this case to delete APT3 application, roles, app-client and their associated user-role assignment (exists in prod), application admin assignment (exists in prod) and fam_privilege_change_audit records **directly from PROD** (not through flyway). Then remove APT3 from Terraform?
@basilv

ianliuwk1019:
As of June 04, if we were to execute the deletion; verified on all environments:
Tools Space:
> NOTICE:  === STARTING CLEANUP FOR PATTERN: APT3_% ===
> NOTICE:  Deleted 2 rows from fam_user_role_xref
> NOTICE:  Deleted 2 rows from fam_access_control_privilege
> NOTICE:  Deleted 5 rows from fam_application_admin
> NOTICE:  Deleted 9 rows from fam_privilege_change_audit
> NOTICE:  Deleted 3 rows from fam_application_client
> NOTICE:  Deleted 6 rows from fam_role
> NOTICE:  Deleted 3 rows from fam_application
> NOTICE:  === CLEANUP COMPLETE ===

Dev Space: 
> NOTICE:  === STARTING CLEANUP FOR PATTERN: APT3_% ===
> NOTICE:  Deleted 0 rows from fam_user_role_xref
> NOTICE:  Deleted 0 rows from fam_access_control_privilege
> NOTICE:  Deleted 0 rows from fam_application_admin
> NOTICE:  Deleted 0 rows from fam_privilege_change_audit
> NOTICE:  Deleted 3 rows from fam_application_client
> NOTICE:  Deleted 6 rows from fam_role
> NOTICE:  Deleted 3 rows from fam_application
[...]

PR #2210, opened 2026-06-05
fix: #2052 Flyway Remove APT3 App

ianliuwk1019:
ref: #2052 

To Reviewer: please do let me know if any concern from flyway as there are  delete statements.

**Note**: This PR only contains a flyway script, to **DELETE** "**APT3**" application (**ONLY**) and related roles and permission assignments from database from ALL environments. APT3 application setup was abandoned previously, so any existing assignments were not important and was just seeds.
- Additional Cognito client for "APT3" and some adjustment will be in separate PR.
- Script was experimented and executed successfully in Tools space: https://github.com/bcgov/nr-forests-access-management/issues/2052#issuecomment-4636231745
- Some safety checks are in script.


arcshiftsolutions (approved):

ianliuwk1019:
Thanks @arcshiftsolutions for your feedback, I probably was too into testing but the above statements should not break others!

PR #2240, opened 2026-07-07
feat: add nr lexis nexcol service client

alexmcdermid:


alexmcdermid:
switching to keycloak for service account

